PR #2, opened 2021-07-24
[WIP] feat: split archive functionality

jpower432:
Features added:
- Bundles are archived and split per user defined size
- Initial functionality for validating bundle checksum and contents
- Bundle full extraction included
- Tests included (67% code coverage)


pkg/archive/archive.go
```diff
@@ -0,0 +1,184 @@
+package archive
```
afflom, line 1: With regard to portability, I think that this might best added as a subpackage to bundle. Thoughts?

jpower432: @afflom I considered it a functional component outside of the bundle that interacts with bundles, but I would be willing to move it under bundle as sub package.

afflom: I vote making it a sub-package. I asked Eric for his opinion.

estroz: Since we don't yet know how these packages/command will look when integrating with `oc`, this is fine for now. Once we know those requirements, we can reorganize packages if needed.

jpower432:
@afflom, as discussed, a couple notes/request for change:

- Allow the user to specify an output directory
- Rely on gzip to validate the tar integrity with CRC
- Ensure that bundles cannot be written to themselves

PR #13, opened 2021-07-27
User-defined images mirroring feature with simple blocking

jpower432:
Features Added:
- User defined image mirroring
- Simple image blocking
- Tests added

### DIFF
--- a/pkg/bundle/additional.go
+++ b/pkg/bundle/additional.go
@@ -1,6 +1,64 @@
 package bundle
 
-// getAdditional downloads specified images in the imageset-config.yaml under mirror.additonalImages
-//func getAdditional(i BundleSpec, rootDir string) {
-//
-//}
+import (
+	"fmt"
+	"os"
+
+	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	"github.com/openshift/oc/pkg/cli/image/mirror"
+	"github.com/sirupsen/logrus"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// GetAdditional downloads specified images in the imageset-config.yaml under mirror.additonalImages
+func GetAdditional(cfg *BundleSpec, rootDir string) error {
+
+	var mappings []mirror.Mapping
+
+	stream := genericclioptions.IOStreams{
+		In:     os.Stdin,
+		Out:    os.Stdout,
+		ErrOut: os.Stderr,
+	}
+
+	opts := mirror.NewMirrorImageOptions(stream)
+	opts.FileDir = rootDir + "/src/"
+
+	logrus.Infof("Downloading %d image(s) to %s", len(cfg.Mirror.AdditionalImages), opts.FileDir)
+
+	for _, img := range cfg.Mirror.AdditionalImages {
+
+		// Get source image information
+		srcRef, err := imagesource.ParseReference(img)
+
+		if err != nil {
+			return fmt.Errorf("error parsing source image %s: %v", img, err)
+		}
+
+		// Set destination image information
+		path := "file://" + img
+
+		dstRef, err := imagesource.ParseReference(path)
+
+		if err != nil {
+			return fmt.Errorf("error parsing destination reference %s: %v", path, err)
+		}
+
+		// Check if image is specified as a blocked image
+		if IsBlocked(cfg, srcRef.Ref) {
+			return fmt.Errorf("additional image %s also specified as blocked, remove the image one config field or the other", img)
+		}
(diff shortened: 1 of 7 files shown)


pkg/bundle/additional.go
```diff
@@ -1,6 +1,67 @@
...
+			logrus.Warnf("Image %s specified as blocked, skipping...", img)
+			continue
+		} else {
+			// Create mapping from source and destination images
+			mappings = append(mappings, mirror.Mapping{
+				Source:      srcRef,
+				Destination: dstRef,
+				Name:        srcRef.Ref.Name,
+			})
+		}
```
estroz, line 58: Nit: the user "knows" the ref is blocked since they configured both blocked and additional images, so throw an error here.
```suggestion
		if IsBlocked(cfg, srcRef.Ref) {
			return fmt.Errorf("additional image %s also specified as blocked, remove the image one config field or the other", img)
		}
		// Create mapping from source and destination images
		mappings = append(mappings, mirror.Mapping{
			Source:      srcRef,
			Destination: dstRef,
			Name:        srcRef.Ref.Name,
		})
	}
```

pkg/bundle/create/create.go
```diff
@@ -28,19 +28,19 @@ func CreateFull(rootDir string) error {
 		return err
 	}
 	logrus.Info(config)
-	if &config.Mirror.Ocp != nil {
+	if config.Mirror.Ocp.Channels != nil {
```
estroz, line 31: This slice may still be empty if non-nil.
```suggestion
	if len(config.Mirror.Ocp.Channels) != 0 {
```

pkg/bundle/image.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/openshift/library-go/pkg/image/reference"
+)
+
+// IsBlocked will return a boolean value on whether an image
+// is specified as blocked in the BundleSpec
+func IsBlocked(cfg *BundleSpec, imgRef reference.DockerImageReference) bool {
+
+	for _, block := range cfg.BlockedImages {
+		if strings.Contains(imgRef.Name, block) {
```
estroz, line 14: Probably better to directly compare names rather than fuzzy matching, since one character set can be a strict subset of another and have this evaluate to true when, which I don't think is desirable behavior right? Ex `strings.Contains("specific-image-name", "reg/ns/my-specific-image-name:latest") == true` is bad because the images are different.
```suggestion
		if imgRef.Name.Exact() == block {
```

pkg/bundle/additional_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"os"
+	"testing"
+)
+
+func Test_GetAdditional(t *testing.T) {
+
+	cfg := &BundleSpec{
+		Mirror: Mirror{
+			BlockedImages:    []string{"alpine"},
+			AdditionalImages: []string{"docker.io/library/alpine:latest", "docker.io/library/busybox:latest"},
```
estroz, line 13: Is there a way to create a fake mirrorer with `oc` libs? Our CI may hit docker limits fairly quickly like this.

estroz (changes requested):
Some suggestions, overall LGTM.

pkg/bundle/create/create.go
```diff
@@ -28,19 +28,19 @@ func CreateFull(rootDir string) error {
...
 	//}
 	//if &config.Mirror.Samples != nil {
 	//GetSamples(*config, rootDir)
-	//}
-	//if &config.Mirror.AdditionalImages != nil {
-	//	getAdditional(*config, rootDir)
-	//}
-	*/
+	//}*/
+	if config.Mirror.AdditionalImages != nil {
```
estroz, line 40:
```suggestion
	if len(config.Mirror.AdditionalImages) != 0 {
```

pkg/bundle/image.go
jpower432: @estroz I agree here. I was going back and forth on this one because if the blocked name is something generic or small if might get a false positive. I will add this one.

pkg/bundle/additional_test.go
jpower432: @estroz Good point. I will into this. Checking some `oc` tests will probably be helpful here.

estroz: In the meantime we can probably merge these tests as-is, just to get this stuff in.

estroz (approved):
Nit: squash commits. Not a big deal if you don't though, so LGTM.

afflom (approved):

PR #19, opened 2021-07-28
feat(config): refactor spec into formal configs

estroz:
This new spec organization follows the component config format (in a sense), which is a super nice pattern for versioned configs. In accordance with a conversation had earlier about naming, a "bundle" is really an "imageset", so some types have been renamed as such. I'll leave further renaming of stuff in this repo to a follow-up PR.


jpower432 (approved):
@estroz This LGTM and I really like the changes you made with making the config/spec conform the Kubernetes API standard. It just needs that merge conflict fixed. Other than that, I think if could use a few more comments especially on the exported types and functions, but not a deal breaker. @afflom I will leave the final approval to you.

afflom (approved):

PR #23, opened 2021-07-29
Add http client interface

david-rh:
Adding some structure to mock the http client in order to be able to create tests for release.go


jpower432 (approved):
LGTM. Just needs the commits squashed before merging.

PR #28, opened 2021-08-06
feat: archiving functionality

jpower432:
Features Added:

- Segmented archiving
- Archive combining
- Test include w/ package 61% test coverage


pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+	err = filepath.Walk(sourceDir, func(fpath string, info os.FileInfo, err error) error {
+
+		if err != nil {
+			return fmt.Errorf("traversing %s: %v", fpath, err)
+		}
+		if info == nil {
+			return fmt.Errorf("no file info")
+		}
+
+		if _, found := foundFiles[fpath]; found {
```
estroz, line 82:
```suggestion
		if _, found := foundFiles[fpath]; !found {
```

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+				}
+				defer file.Close()
+			}
+
+			f := archiver.File{
+				FileInfo: archiver.FileInfo{
+					FileInfo:   info,
+					CustomName: nameInArchive,
+				},
+				ReadCloser: file,
```
estroz, line 104: If this file doesn't have regular bits set, then `file` is nil. Will this cause the archiver to panic? More generally, did you mean to skip directories?

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+
+// ExtractArchive will unpack the archive at the specified directory
+func ExtractArchive(a Archiver, src, dest string) error {
+	return a.Unarchive(src, dest)
+}
+
+// CombineArchives take a list of archives and combines them into one file
+func CombineArchives(in Archiver, out Archiver, destDir, name string, paths ...string) error {
+
+	outPath := fmt.Sprintf("%s/%s", destDir, name)
```
estroz, line 164:
```suggestion
	outPath := path.Join(destDir, name)
```

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+		} else {
+			logrus.Debugf("File %s found, skipping...", fpath)
+		}
+
+		return nil
+	})
+
+	// Close final archive
+	a.Close()
+	splitFile.Close()
```
estroz, line 151: Check these errors

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+					return err
+				}
+			}
+		}()
+
+		if readErr != nil {
+			return readErr
+		}
+
+		in.Close()
```
estroz, line 218: Check this error

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	out.Close()
```
estroz, line 221: Check this error

pkg/bundle/create/create.go
```diff
@@ -154,17 +159,53 @@ func CreateDiff(configPath, rootDir string) error {
...
+
+	// Set get absolute path to output dir
+	output, err := filepath.Abs(outputDir)
+
+	if err != nil {
+		return err
+	}
+
+	// Change dir before archving to avoid issues with symlink paths
+	os.Chdir(rootDir)
```
estroz, line 194: Check this error.

pkg/archive/archive.go
```diff
@@ -0,0 +1,224 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/mholt/archiver/v3"
+	"github.com/sirupsen/logrus"
+)
+
+type Archiver interface {
```
estroz, line 13: WDYT about using [counterfeiter](https://github.com/maxbrunsfeld/counterfeiter/) to test this?

estroz (changes requested):
A few fixes needed, otherwise LGTM

pkg/archive/archive.go
jpower432: @estroz This is set to only process found files, so found should equal true here. Kind of opposite logic from the ReconcileFiles logic; `bundle.ReconcileFiles` now returns a slice of string with the files names that should be packed.

pkg/archive/archive.go
jpower432: @estroz Here is the logic in the library for Write https://github.com/mholt/archiver/blob/81fac4ba96e457b3f32c56b848b725351f0470b2/tar.go#L362. It does alternative processing for directories and symlinks, but here I did mean to skip directories.

pkg/archive/archive.go
estroz: SGTM. The log message below threw me off, thanks for clarifying.

pkg/archive/archive.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+
+			// Write file to current archive file
+			if err = a.Write(f); err != nil {
+				return fmt.Errorf("%s: writing: %s", fpath, err)
+			}
+
+			splitSize += info.Size()
+
+			foundFiles[fpath] = struct{}{}
```
estroz, line 141: Since the path is already in `foundFiles`
```suggestion
```

pkg/archive/archive.go
```diff
@@ -0,0 +1,233 @@
...
+		} else {
+			logrus.Debugf("File %s found, skipping...", fpath)
```
estroz, line 143:
```suggestion
			logrus.Debugf("File %s not found, skipping...", fpath)
```

pkg/archive/archive.go
estroz: I think this will cause an error: https://github.com/mholt/archiver/blob/81fac4ba96e457b3f32c56b848b725351f0470b2/tar.go#L397-L399

pkg/bundle/create/create.go
```diff
@@ -3,28 +3,31 @@ package create
...
 	"github.com/RedHatGov/bundle/pkg/operator"
 )
 
 // CreateFull performs all tasks in creating full imagesets
-func CreateFull(configPath, rootDir string, dryRun, insecure bool) error {
+func CreateFull(configPath, rootDir, outputDir string, dryRun, insecure bool, segSize int64) error {
 
 	ctx := context.Background()
 
+	rootAbs, err := filepath.Abs(rootDir)
```
estroz, line 25: Check this error.

pkg/bundle/create/create.go
```diff
@@ -154,17 +159,55 @@ func CreateDiff(configPath, rootDir string) error {
...
+
+	// Set get absolute path to output dir
+	output, err := filepath.Abs(outputDir)
+
+	if err != nil {
+		return err
+	}
+
+	// Change dir before archving to avoid issues with symlink paths
+	if err := os.Chdir(rootDir); err != nil {
```
estroz, line 194: Is it possible to archive without chdir, or at least `defer os.Chdir(oldWorkingDir)` so this function doesn't cause a potentially unexpected side-effect for downstream execution?

pkg/archive/archive.go
jpower432: @estroz I was just checking out this tool and it looks really cool. I would like to do a follow on ticket to use this.

pkg/bundle/create/create.go
jpower432: @estroz All good with adding a `defer` statement to change back directories. Really good idea. The reason I am doing a `Chdir` is because of how the symlinks are being processed during the write. I believe there is an open PR on the archiver library to get that fixed.

pkg/bundle/create/create.go
jpower432: Done

cmd/oc-bundle/create.go
```diff
@@ -7,7 +7,15 @@ import (
 	"github.com/RedHatGov/bundle/pkg/bundle/create"
 )
 
+type createOpts struct {
+	segSize   int64
```
estroz, line 11: Thoughts on hiding this for now?

afflom: How about putting it in the imageset-config?

estroz (approved):
/lgtm

estroz:
@jpower432 are each of the 4 commits self-contained? If not I'd squash them, up to you though.

PR #29, opened 2021-08-06
feat(operator): catalog metadata types for prior index data

estroz:
This PR adds the ability to store and pull image set metadata from a backend, and fixes some `create` metadata handling issues. Currently only a local directory storage backend is supported.


estroz:
Blocking on #28 so I can rebase.

pkg/bundle/create/create.go
```diff
@@ -14,40 +15,53 @@ import (
...
-	if len(metadata.PastMirrors) != 0 {
-		return config.NewFullMetadataError(metadataPath)
+	// Read in current metadata
+	meta := v1alpha1.Metadata{}
+	switch err := backend.ReadMetadata(ctx, &meta); {
+	case err != nil && !errors.Is(err, storage.ErrMetadataNotExist):
+		return err
+	case err == nil && len(meta.PastMirrors) != 0:
+		// No past run(s) are allowed when generating a full imageset.
+		// QUESTION: should the command just log a warning and ignore this metadata instead?
```
jpower432, line 55: @estroz Responding to your comment in the code, would it better to prompt the user or would we send a warning when creating a full and destroy the current metadata. @afflom and I talked about this and the user should be able to create a full even if there is existing metadata, but the only area of concern I have about this is the user accidentally running `create full`. Maybe a prompt or a metadata backup? WDYT?

pkg/bundle/create/create.go
```diff
@@ -117,30 +130,36 @@ func CreateFull(configPath, rootDir, outputDir string, dryRun, insecure bool) er
...
 	}
 
-	lastRun := metadata.PastMirrors[len(metadata.PastMirrors)-1]
-
-	// Set timestamp
-	lastRun.Timestamp = int(time.Now().Unix())
+	lastRun := meta.PastMirrors[len(meta.PastMirrors)-1]
+	run := v1alpha1.PastMirror{
+		Sequence:  lastRun.Sequence + 1,
+		Uid:       uuid.New(),
```
jpower432, line 160:
```suggestion
		Uid:       lastRun.Uid,
```

jpower432 (changes requested):
LGTM, but needs a change for the diff Uid to maintain the same Uid between runs. We may want to also discuss how we are going to handle new `create full` runs on currently managed data.

pkg/bundle/create/create.go
estroz: Prompt sounds fine. I'll create a follow-up.

pkg/bundle/create/create.go
afflom: That seems logical

jpower432 (approved):
/lgtm

Issue #8, opened 2021-07-26
Handle Blocked images


afflom:
Prevent blocked images from being downloaded when declared in additionalImages and operators

PR #1, opened 2021-07-23
feat(mirror): mirror operator catalogs

estroz:
Closes #4

Signed-off-by: Eric Stroczynski <ericstroczynski@gmail.com>

PR #32, opened 2021-08-11
feat: block operator images

jpower432:
- Added a IsBlocked check for the operator catalogs
- Main caveat is that I need to edit a function that we had temporarily so I could have access to each image.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
+	"github.com/RedHatGov/bundle/pkg/bundle"
 	"github.com/RedHatGov/bundle/pkg/config"
 	"github.com/RedHatGov/bundle/pkg/config/v1alpha1"
 )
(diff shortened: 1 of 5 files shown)


jpower432:
Closes issue #8. Blocked by PR #1.

data/imageset-config.yaml
```diff
@@ -10,17 +10,7 @@ mirror:
...
-      packages:
-        - name: couchbase-operator
-        - name: mongodb-operator
-          versions:
-            - '1.4.0'
-        - name: crunchy-postgresql-operator
-          channels:
-            - name: 'stable'
-    - catalog: community-operators:v4.7
+    - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.8
```
estroz, line 13: Let's keep this config as-is unless the config spec changes, since it's an example config for users.

pkg/operator/operator.go
```diff
@@ -274,15 +275,21 @@ func refToFileScheme(ref imgreference.DockerImageReference) string {
...
 		return nil, err
 	}
 
 	return func(mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference) error {
+
 		mappings := []imgmirror.Mapping{}
 		for from, to := range mapping {
+
+			if bundle.IsBlocked(cfg, from.Ref) {
+				logrus.Warnf("image %s was specified as blocked per config, skipping...", from.Ref.Name)
```
estroz, line 290: Since this is a user-configured action, I'd just debug log this.
```suggestion
				logrus.Debugf("image %s was specified as blocked per config, skipping...", from.Ref.Name)
```

estroz (changes requested):
One fix, otherwise LGTM

estroz (approved):

PR #33, opened 2021-08-12
fix(mirror): use quay.io images in unit tests to avoid docker limits

estroz:
This is just a holdover until a better solution is found.


jpower432 (approved):
/lgtm

PR #34, opened 2021-08-12
feat(mirror): handling pull secrets through imageset config

jpower432:
Features:
- Processes pullSecret for OCP, each catalog, and each user defined image
- If the pullSecret is not specified `oc mirror` will default for the docker config default place

### DIFF
--- a/pkg/config/credentials.go
+++ b/pkg/config/credentials.go
@@ -1,0 +1,130 @@
+package config
+
+import (
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/docker/distribution/registry/client/auth"
+	"github.com/docker/distribution/registry/client/transport"
+	"github.com/openshift/library-go/pkg/image/registryclient"
+	"github.com/openshift/oc/pkg/helpers/image/credentialprovider"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+)
+
+// New creates a new credential store for the provided secret content
+func New(secret []byte) (auth.CredentialStore, error) {
+
+	cfg, err := readDockerConfigJSONFileFromBytes(secret)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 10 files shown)


pkg/config/credentials.go
```diff
@@ -0,0 +1,127 @@
...
+	}, nil
+}
+
+// CreateContext a new context for the registryClient of `oc mirror`
+func CreateContext(secret []byte, skipVerification bool) (*registryclient.Context, error) {
+	rt, err := rest.TransportFor(&rest.Config{})
+	if err != nil {
+		return nil, err
+	}
+	insecureRT, err := rest.TransportFor(&rest.Config{TLSClientConfig: rest.TLSClientConfig{Insecure: true}})
```
estroz, line 39: TLS verification should only be skipped when `--skip-tls` is set right?

pkg/config/credentials.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, err
+	}
+	creds, err := New(secret)
+
+	if err != nil {
+		return nil, err
+	}
+	context := registryclient.NewContext(rt, insecureRT).WithCredentials(creds).
+		WithRequestModifiers(transport.NewHeaderRequestModifier(http.Header{http.CanonicalHeaderKey("User-Agent"): []string{rest.DefaultKubernetesUserAgent()}}))
```
estroz, line 49: Nit: this is more readable:
```suggestion
	context := registryclient.NewContext(rt, insecureRT)
	  .WithCredentials(creds)
	  .WithRequestModifiers(transport.NewHeaderRequestModifier(
		http.Header{
			http.CanonicalHeaderKey("User-Agent"): []string{rest.DefaultKubernetesUserAgent()},
		},
	  ))
```

pkg/config/credentials.go
jpower432: Good catch. Yes, you are right. Updating now.

estroz (approved):

PR #41, opened 2021-08-19
feat(publish): worker pool implementation

jpower432:
Features added:
- Concurrent searching for files in Metadata using a worker pool library

### DIFF
--- a/pkg/bundle/files.go
+++ b/pkg/bundle/files.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/RedHatGov/bundle/pkg/config"
 	"github.com/RedHatGov/bundle/pkg/config/v1alpha1"
-	"github.com/gammazero/workerpool"
+	"github.com/openshift/oc/pkg/cli/image/workqueue"
 	"github.com/sirupsen/logrus"
 )
 
@@ -16,9 +16,20 @@ import (
 // and checks against the current list
 func ReconcileFiles(meta v1alpha1.Metadata, rootDir string) (newFiles []v1alpha1.File, err error) {
 
-	wp := workerpool.New(10)
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	wp := workqueue.New(10, stopCh)
 
 	mu := new(sync.RWMutex)
+
+	if err != nil {
+		return nil, err
+	}
+
 	results := make(chan v1alpha1.File)
 
 	foundFiles := make(map[string]struct{}, len(meta.PastFiles))
@@ -28,6 +39,21 @@ func ReconcileFiles(meta v1alpha1.Metadata, rootDir string) (newFiles []v1alpha1
 
 	// Ignore the current dir.
 	foundFiles["."] = struct{}{}
+
+	go func() {
+
+		defer wg.Done()
+		for {
+			rst, ok := <-results
+			if !ok {
+				break
+			}
+			if base := filepath.Base(rst.Name); base != config.MetadataFile && base != config.AssociationsFile {
+				newFiles = append(newFiles, rst)
+				logrus.Debugf("File %s added", rst.Name)
+			}
+		}
+	}()
 
 	//Create a work pool
 	logrus.Debug("Creating worker pool")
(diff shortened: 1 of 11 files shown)


estroz (commented):
Have you considered using an existing parallelized queue implementation like https://pkg.go.dev/k8s.io/client-go/util/workqueue#ParallelizeUntil? Ultimately if you find you have to write such code yourself that’s fine, but I’d rather the project not have to test a hand-rolled implementation if avoidable.

jpower432:
@estroz Understandable. I will look through this library and see if it meets our needs.

estroz:
There are other generic Go libs out there too, that’s just one suggestion from k8s. Feel free to shop around. I’m open to whatever!

jpower432:
Sounds good. I did a bit of looking around before I wrote my own and couldn't really find a library that was actively maintained and allowed for error handling. I never came across this one though so I need to alter my search criteria. Totally agree that it would be better to not have to maintain it ourselves.

Makefile
```diff
@@ -1,12 +1,12 @@
-# Add build flags to GOFLAGS=<flags> here.
 GO := go
+GO_BUILD_FLAGS := -tags=json1
```
estroz, line 2: Some of these changes are from a prior commit in main. Can you rebase so they go away?

pkg/bundle/files.go
```diff
@@ -4,42 +4,80 @@ import (
...
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/RedHatGov/bundle/pkg/config/v1alpha1"
+	"github.com/gammazero/workerpool"
+	"github.com/sirupsen/logrus"
 )
 
+type Reconciler struct {
```
estroz, line 14: Typically I either define an interface for a type that requires instantiation, or permit the struct to be used directly with an `init()` method called in exported methods; in this case I'd go with the latter option. Right now `Reconciler` can be used directly but will panic since its unexported fields are unset. For example:
```go
type Reconciler struct {
	newFiles   map[string]v1alpha1.File
	foundFiles map[string]struct{}
	mu         *sync.Mutex
    initOnce sync.Once
}

func (r *Reconciler) init() {
	r.initOnce.Do(func() {
		r.newFiles = make(map[string]v1alpha1.File)
		r.foundFiles = make(map[string]struct{})
		r.mu = new(sync.Mutext)
	})
}

func (r *Reconciler) Files(v1alpha1.Metadata, string) (map[string]v1alpha1.File, error) {
	r.init()
	...
}
```

pkg/bundle/files.go
```diff
@@ -4,42 +4,80 @@ import (
...
+	newFiles   map[string]v1alpha1.File
+	foundFiles map[string]struct{}
+	mu         *sync.Mutex
```
estroz, line 17: Since `Reconciler`'s methods have a pointer receiver, I'd use a `sync.Mutex` value instead. Also this can be a `sync.RWMutex` with the below suggested changes.
```suggestion
	mu         sync.RWMutex
```

pkg/bundle/files.go
```diff
@@ -4,42 +4,80 @@ import (
...
+		logrus.Debugf("Creating task for %s", fpath)
+
+		wp.Submit(func() {
+			file := v1alpha1.File{
+				Name: fpath,
+			}
+
+			r.mu.Lock()
+			_, found := r.foundFiles[fpath]
+			r.mu.Unlock()
```
estroz, line 64:
```suggestion
			r.mu.RLock()
			_, found := r.foundFiles[fpath]
			r.mu.RUnlock()
```

estroz (changes requested):
Glad you found a suitable library. Have you benchmarked this change, and if so can you share the results?

pkg/bundle/files.go
jpower432: Resolving. End up reworking this over the weekend and found that I didn't really need the struct.

pkg/bundle/files.go
jpower432: Changed added. Resolving.

pkg/bundle/files.go
jpower432: Changed added. Resolving.

jpower432:
@estroz I did run a benchmark on concurrent vs serial file reconciling and I don't really see a major difference in performance between the two (in fact running this concurrency seems to be a tad slower). Since the original intent of this PR was to find a worker pool solution and not necessarily to make file reconciling concurrent, we might be able to just note the worker pool library we want to use and close this PR. The worker pool would be much more valuable on the publisher side. WDYT?

estroz:
Up to you. I'll approve it but feel free to close it and just use the worker pool in `publish`.

estroz (approved):

jpower432:
Closing, but the chosen worker pool solution is library github.com/openshift/oc/pkg/cli/image/workqueue

PR #44, opened 2021-08-20
chore(build): add -tags=json1 so sqlite CGo compiles correctly

estroz:
`-tags=json1` is needed for the sqlite CGo libs that operator-registry uses for handling sqlite-backed catalog indices.


afflom:
@estroz, Is this good to go? I'd like to test with it soonest.

afflom (approved):

PR #47, opened 2021-08-23
fix: changed path to remove extra src and added targetfn to release dest

jpower432:
This PR adds a fix required to successfully add image associations for releases.
- Remove extra "src" from mapping destination path
- Used `oc` TargetFn to get the manifest location for the release image


afflom (approved):
Appears to resolve release bug

PR #49, opened 2021-08-24
fix: create a custom error for image mapping

jpower432:
- Creates a custom error for "image missing mapping" because operator do not always have mapping information. This logs a warning instead.


afflom (approved):
This is just a workaround for now, but it works as advertised.

PR #55, opened 2021-08-30
fix: added a check for err being nil when casting an error

jpower432:
Fixed a bug when type casting an error in operator.go

### DIFF
--- a/pkg/operator/mirror.go
+++ b/pkg/operator/mirror.go
@@ -423,11 +423,11 @@ func (o *MirrorOptions) associateDeclarativeConfigImageLayers(mappingDir string,
 		srcDir := filepath.Join(o.RootDestDir, config.SourceDir)
 		assocs, err := image.AssociateImageLayers(srcDir, imgMappings, images)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/operator/mirror.go
```diff
@@ -422,10 +422,12 @@ func (o *MirrorOptions) associateDeclarativeConfigImageLayers(mappingDir string,
...
-		if merr, ok := err.(*image.MirrorError); ok {
-			logrus.Warn(merr)
-		} else {
-			return err
+		if err != nil {
+			if merr, ok := err.(*image.MirrorError); ok {
+				logrus.Warn(merr)
+			} else {
+				return err
+			}
```
estroz, line 430: Casting this error may work sometimes, but not if the error has a parent. Use [`errors.As`](https://pkg.go.dev/errors?utm_source=godoc#example-As) here instead:
```suggestion
			merr := &image.MirrorError{}
			if !errors.As(err, &merr) {
				return err
			}
			logrus.Warn(err)
```

estroz (approved):

PR #60, opened 2021-09-01
 feat(publish): mirror all locally reconstituted images to mirror registry

estroz:
This PR implements a bunch of mirroring logic from archived image mounts, and adds a test for heads-only catalog mirror mode.


pkg/config/v1alpha1/metadata_types.go
```diff
@@ -53,7 +53,10 @@ func (pms PastMirrors) Swap(i, j int)      { pms[i], pms[j] = pms[j], pms[i] }
 func (pms PastMirrors) Less(i, j int) bool { return pms[i].Sequence < pms[j].Sequence }
 
 type Blob struct {
-	Name string `json:"name"`
+	ID string `json:"id"`
+	// NamespaceName of image that owns this blob.
+	// Required for blob lookups during the publish step.
+	NamespaceName string `json:"namespaceName"`
```
jpower432, line 59: @estroz Can you explain this addition a bit more? What about blobs that belong to multiple images/namespaces? Could we make this a slice or remove it? Right now we are doing lookup by layer digest, but I want to make sure I am not missing something.

jpower432 (commented):
Just one question, but other than that LGTM. Will review again after merge conflicts are resolved.

pkg/config/v1alpha1/metadata_types.go
estroz: In order to pull a blob, you need to know what image stores that blob in the registry. With NamespacedName the request url would look like `<registry host>/v2/<NamespacedName>/blobs/<ID>`. The blob is the same regardless of whether multiple image manifests specify that blob since it has a globally unique ID, so only one path is needed to pull it. For redundancy this could be a list of all images in case one image was removed from the registry, but that can be a follow-up.

pkg/bundle/additional.go
```diff
@@ -88,5 +88,13 @@ func (o *AdditionalOptions) GetAdditional(_ v1alpha1.PastMirror, cfg v1alpha1.Im
 		return nil, err
 	}
 
-	return image.AssociateImageLayers(opts.FileDir, assocMappings, images)
+	assocs, err := image.AssociateImageLayers(opts.FileDir, assocMappings, images)
+	if err != nil {
+		return nil, err
+	}
+	for _, assoc := range assocs {
+		assoc.Type = image.TypeGeneric
```
jpower432, line 96: @estroz This type value is being set on a copy of the Association and not being set in the return Associations map.

estroz: Fixed by updating the map after setting type.

jpower432 (approved):
LGTM

PR #69, opened 2021-09-02
fix: updated logic to only process top level release association

jpower432:
Fixed release mapping on publisher to only the top level association is processed

### DIFF
--- a/pkg/bundle/publish/publish.go
+++ b/pkg/bundle/publish/publish.go
@@ -230,21 +230,21 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
 		m.Destination = toMirrorRef
 		m.Destination.Ref.Namespace = m.Source.Ref.Namespace
 		m.Destination.Ref.Name = m.Source.Ref.Name
+		m.Destination.Ref.Tag = m.Source.Ref.Tag
+		m.Destination.Ref.ID = m.Source.Ref.ID
 
 		switch assoc.Type {
 		case image.TypeGeneric:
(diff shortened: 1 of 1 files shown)


pkg/bundle/publish/publish.go
```diff
@@ -229,15 +230,21 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
 		case image.TypeGeneric:
+			m.Destination.Ref.Tag = m.Source.Ref.Tag
+			m.Destination.Ref.ID = m.Source.Ref.ID
 			genericMappings = append(genericMappings, m)
 		case image.TypeOCPRelease:
-			releaseMappings = append(releaseMappings, m)
+			// Only add top level release images to
+			// release mapping
+			if strings.Contains(assoc.Name, "ocp-release") {
+				releaseMappings = append(releaseMappings, m)
```
estroz, line 243: Destination tag/id is not set here. Is that ok for release images?

jpower432: Yes. That was actually causing an error with mirror release. It is not expecting a tag or digest in the destination.

estroz: Ok. Small not: so bugs don’t crop up for other image types we may add later, it’s better to set tag and id at the top level (like the code is doing prior to these changes) then clear them in the `ocp-release` conditional branch, rather than set them in all other branches.

jpower432: Pushed the change.

afflom: testing

estroz (approved):

PR #71, opened 2021-09-03
fix(publish): bug/initial sequence check

jpower432:
Changes added:
- Initial publish unit tests
- path validation for user flags
- initial metadata sequence check
- fixes error handling on operator associations


afflom (approved):
Identified app hang at end of publishing, but all tasks complete successfully prior to runtime timeout.

PR #75, opened 2021-09-09
feat(publish): feat/top level associations

jpower432:
Problem: 
- For catalogs and release, we need to identify the top level association to pass to the `oc mirror` Run function.

Solution: 
- Created a new field in Association to include a TopLevel bool. This value is set based on whether the manifest is an index or by 
the operators and release packages.

Notes:
- Also adds an unarchiving function with the ability to exclude files and directories (not a follow on PR/issue for unit tests)

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -219,6 +219,8 @@ func includeFile(fpath string) bool {
 
 // Copied from mholt archiver repo. Temporary and can
 // add back to repo
+// Changes include the additional of the excludePath variable and is
+// passed to the untarNext function
 func Unarchive(a Archiver, source, destination string, excludePaths []string) error {
 
 	if !fileExists(destination) {
@@ -287,6 +289,9 @@ func untarNext(a Archiver, destination string, exclude []string) error {
 		return fmt.Errorf("checking path traversal attempt: %v", errPath)
 	}
 
+	// Added change here to check if
+	// current path is in the exclusion
+	// list
 	for _, path := range exclude {
 		if within(path, header.Name) {
 			return nil
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -1,10 +1,13 @@
 package archive
 
 import (
+	"archive/tar"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/RedHatGov/bundle/pkg/config"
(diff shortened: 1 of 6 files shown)


pkg/bundle/publish/publish.go
```diff
@@ -374,13 +376,14 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
 		catOpts.DryRun = o.DryRun
 		catOpts.MaxPathComponents = 2
 		catOpts.SecurityOptions.Insecure = o.SkipTLS
-		//catOpts.FilterOptions = imagemanifest.FilterOptions{FilterByOS: ".*"}
+		catOpts.FilterOptions.FilterByOS = "linux/amd64"
```
estroz, line 379: This won't actually work since it gets [hardcoded](https://github.com/openshift/oc/blob/183090787d00d5150440dd75d71b00a72739a86c/pkg/cli/admin/catalog/mirror.go#L310-L314) in the library function.

pkg/archive/archive.go
```diff
@@ -212,3 +216,177 @@ func includeFile(fpath string) bool {
 	split := strings.Split(filepath.Clean(fpath), string(filepath.Separator))
 	return split[0] == config.InternalDir || split[0] == config.PublishDir || split[0] == "catalogs"
 }
+
+// Copied from mholt archiver repo. Temporary and can
+// add back to repo
+func Unarchive(a Archiver, source, destination string, excludePaths []string) error {
```
estroz, line 222: Can you add comments above what you changed in the copied functions? They'll be easier to reference when rebasing later.

pkg/bundle/publish/publish.go
```diff
@@ -374,13 +376,14 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
 		catOpts.DryRun = o.DryRun
 		catOpts.MaxPathComponents = 2
 		catOpts.SecurityOptions.Insecure = o.SkipTLS
-		//catOpts.FilterOptions = imagemanifest.FilterOptions{FilterByOS: ".*"}
+		catOpts.FilterOptions = imagemanifest.FilterOptions{FilterByOS: o.FilterByOS}
```
estroz, line 379: Another (non-blocking) thought: this could be done in the `create` step right?

estroz (approved):

pkg/bundle/publish/publish.go
jpower432: I believe so. That’s why I put it in root options. Can be done in a follow on PR :).

PR #80, opened 2021-09-11
feat: multimap associations

jpower432:
- Changes to associations allow for process images and all of the associations values one at a time.
- Releases are kept together by updating the key to the release image after gathering the associations
- Added cleanup actions to archiver

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"fmt"
 	"io"
+	"io/fs"
 	"log"
 	"os"
 	"path/filepath"
@@ -70,7 +71,7 @@ func NewPackager(manifests []v1alpha1.Manifest, blobs []v1alpha1.Blob) *packager
 }
 
 // CreateSplitAchrive will create multiple tar archives from source directory
-func (p *packager) CreateSplitArchive(maxSplitSize int64, destDir, sourceDir, prefix string) error {
+func (p *packager) CreateSplitArchive(maxSplitSize int64, destDir, sourceDir, prefix string, skipCleanup bool) error {
 
 	// Declare split variables
 	splitNum := 0
@@ -141,8 +142,12 @@ func (p *packager) CreateSplitArchive(maxSplitSize int64, destDir, sourceDir, pr
 		if info.Size()+splitSize > maxSplitSize {
 
 			// Close current tar archive
-			p.Close()
-			splitFile.Close()
+			if err := p.Close(); err != nil {
+				return err
+			}
+			if err := splitFile.Close(); err != nil {
+				return err
+			}
 
 			// Increment split number and reset splitSize
 			splitNum += 1
@@ -160,6 +165,14 @@ func (p *packager) CreateSplitArchive(maxSplitSize int64, destDir, sourceDir, pr
 		// Write file to current archive file
 		if err = p.Write(f); err != nil {
 			return fmt.Errorf("%s: writing: %s", fpath, err)
+		}
+
+		// Delete file after written to archive
+		if shouldRemove(fpath, info) && !skipCleanup {
+			if err := os.Remove(fpath); err != nil {
+				return err
+			}
+
 		}
 
 		logrus.Debugf("File %s added to archive", fpath)
@@ -215,6 +228,10 @@ func blobInArchive(file string) string {
 func includeFile(fpath string) bool {
 	split := strings.Split(filepath.Clean(fpath), string(filepath.Separator))
 	return split[0] == config.InternalDir || split[0] == config.PublishDir || split[0] == "catalogs"
+}
+
+func shouldRemove(fpath string, info fs.FileInfo) bool {
+	return !includeFile(fpath) && !info.IsDir()
 }
 
 // Copied from mholt archiver repo. Temporary and can
--- a/pkg/bundle/publish/manifests.go
+++ b/pkg/bundle/publish/manifests.go
@@ -1,0 +1,268 @@
+package publish
+
+import (
+	"fmt"
+	"hash/fnv"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	"gopkg.in/yaml.v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Copied from https://github.com/openshift/oc/blob/5d8dfa1c2e8e7469d69d76f21e0a166a0de8663b/pkg/cli/admin/catalog/mirror.go#L549
+// Changes made are breaking ICSP and Catalog Source generation into difference functions
+func WriteICSP(out io.Writer, dir string, icsps [][]byte) error {
+
+	if err := ioutil.WriteFile(filepath.Join(dir, "imageContentSourcePolicy.yaml"), aggregateICSPs(icsps), os.ModePerm); err != nil {
+		return fmt.Errorf("error writing ImageContentSourcePolicy")
+	}
+
+	fmt.Fprintf(out, "wrote mirroring manifests to %s\n", dir)
+
+	return nil
+}
+
+func WriteCatalogSource(out io.Writer, source imagesource.TypedImageReference, dir string, mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference) error {
+
+	catalogSource, err := generateCatalogSource(source, mapping)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(filepath.Join(dir, fmt.Sprintf("catalogSource-%s.yaml", source.Ref.Name)), catalogSource, os.ModePerm); err != nil {
+		return fmt.Errorf("error writing CatalogSource")
+	}
+
+	fmt.Fprintf(out, "wrote mirroring manifests to %s\n", dir)
+
+	return nil
+}
+
+func GenerateICSP(out io.Writer, name string, byteLimit int, icspScope string, mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference) ([]byte, error) {
+	registryMapping := getRegistryMapping(out, icspScope, mapping)
+	icsp := operatorv1alpha1.ImageContentSourcePolicy{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: operatorv1alpha1.GroupVersion.String(),
+			Kind:       "ImageContentSourcePolicy"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: strings.Join(strings.Split(name, "/"), "-"),
+			Labels: map[string]string{
+				"operators.openshift.org/catalog": "true",
+			},
+		},
+		Spec: operatorv1alpha1.ImageContentSourcePolicySpec{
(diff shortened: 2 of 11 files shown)


afflom:
closes #84

pkg/archive/archive.go
```diff
@@ -217,6 +230,13 @@ func includeFile(fpath string) bool {
 	return split[0] == config.InternalDir || split[0] == config.PublishDir || split[0] == "catalogs"
 }
 
+func delete(fpath string, info fs.FileInfo) bool {
```
estroz, line 233: Can you call this something else so there's no conflict with the builtin `delete()` function?
```suggestion
func shouldRemove(fpath string, info fs.FileInfo) bool {
```

pkg/archive/archive.go
```diff
@@ -217,6 +230,13 @@ func includeFile(fpath string) bool {
...
+	if !includeFile(fpath) && !info.IsDir() {
+		return true
+	}
+	return false
```
estroz, line 237:
```suggestion
	return !includeFile(fpath) && !info.IsDir()
```

pkg/bundle/publish/manifests.go
```diff
@@ -0,0 +1,309 @@
...
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/oc/pkg/cli/image/imagesource"
+	"gopkg.in/yaml.v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Copied from https://github.com/openshift/oc/blob/5d8dfa1c2e8e7469d69d76f21e0a166a0de8663b/pkg/cli/admin/catalog/mirror.go#L549
+// Changes made are breaking ICSP and Catalog Source generation into difference functions
```
estroz, line 22: Any reason why you're doing this? If the image digests are provided, why not use the default ICSPs and CatalogSources generated by `oc` libs?

pkg/bundle/publish/publish.go
```diff
@@ -185,223 +173,166 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
-		// All manifest layers will be pulled below if associated,
-		// so just sanity-check that the layers are referenced in some association.
-		if len(assoc.ManifestDigests) != 0 {
-			for _, manifestDigest := range assoc.ManifestDigests {
-				if _, hasManifest := assocs[manifestDigest]; !hasManifest {
-					errs = append(errs, fmt.Errorf("image %q: expected associations to have manifest %s but was not found", imageName, manifestDigest))
+					switch _, err := os.Stat(manifestPath); {
+					case err == nil:
+						logrus.Debugf("Manifest found %s found in %s", manifestDigest, assoc.Path)
+					case errors.As(err, &os.ErrNotExist):
```
estroz, line 211:
```suggestion
					case errors.Is(err, os.ErrNotExist):
```

pkg/bundle/publish/publish.go
```diff
@@ -185,223 +173,166 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
+			for _, layerDigest := range assoc.LayerDigests {
+				logrus.Debugf("Found layer %v for image %s", layerDigest, imageName)
+				// Construct blob path, which is adjacent to the manifests path.
+				blobPath := filepath.Join("blobs", layerDigest)
+				imagePath := filepath.Join(o.tmp, "v2", assoc.Path)
+				imageBlobPath := filepath.Join(imagePath, blobPath)
+				switch err := unpack(blobPath, imagePath, filesInArchive); {
+				case err == nil:
+					logrus.Debugf("Blob %s found in %s", layerDigest, assoc.Path)
+				case errors.As(err, &os.ErrNotExist) || errors.As(err, &aerr):
```
estroz, line 235:
```suggestion
				case errors.Is(err, os.ErrNotExist) || errors.As(err, &aerr):
```

pkg/bundle/publish/manifests.go
jpower432: @estroz we are creating ICSPs for user defined images and releases so I broke it up so we will only generate catalog sources for catalog index images.

pkg/bundle/publish/publish.go
```diff
@@ -185,223 +173,166 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
-				logrus.Debugf("Blob %s found in %s", layerDigest, assoc.Path)
-			case errors.As(err, &os.ErrNotExist) || errors.As(err, &aerr):
-				// Image layer must exist in the mirror registry since it wasn't archived,
-				// so fetch the layer and place it in the blob dir so it can be mirrored by `oc`.
-				missingLayers[layerDigest] = append(missingLayers[layerDigest], imageBlobPath)
-			default:
-				errs = append(errs, fmt.Errorf("accessing image %q blob %q at %s: %v", imageName, layerDigest, blobPath, err))
+			// Unpack association main manifest
+			if err := unpack(filepath.Join(manifestPath, assoc.ID), o.tmp, filesInArchive); err != nil {
+				errs = append(errs, err)
```
estroz, line 223: Need to `continue` on errors.

pkg/archive/archive.go
jpower432: Good point. I will change the name of this.

pkg/image/association.go
```diff
@@ -35,8 +35,18 @@ func (e *ErrInvalidComponent) Error() string {
 	return fmt.Sprintf("image %q has invalid component %q", e.image, e.tag)
 }
 
-// Associations is a set of image Associations.
-type Associations map[string]Association
+// AssociationSet is a mapped for Association
+// searching
+type AssociationSet map[*Association]struct{}
```
estroz, line 40: Using a pointer as a map key is dangerous. I would use the association's ID value instead.

jpower432: The intent here was trying to map the associations struct to the images. Do you think it would be better to just use a a map with a slice as a values so we can have multiple values per key or would it be better to move the Association to the values in the AssociationSet and use the ID as the key?

pkg/image/association.go
```diff
@@ -35,8 +35,18 @@ func (e *ErrInvalidComponent) Error() string {
 	return fmt.Sprintf("image %q has invalid component %q", e.image, e.tag)
 }
 
-// Associations is a set of image Associations.
-type Associations map[string]Association
+// AssociationSet is a mapped for Association
+// searching
+type AssociationSet map[string]Association
```
estroz, line 40:
```suggestion
type Associations map[string]Association
```

pkg/image/association.go
```diff
@@ -35,8 +35,18 @@ func (e *ErrInvalidComponent) Error() string {
...
+
+// Associations is a set of image Associations
+// mapped to their images
+type Associations map[string]AssociationSet
```
estroz, line 44: Wdyt about
```suggestion
type AssociationsSet map[string]AssociationSet
```
with the above change

pkg/image/association.go
```diff
@@ -89,10 +96,97 @@ func (it ImageType) String() string {
...
+	values = make([]Association, len(asSet))
+	count := 0
+	for _, value := range asSet {
+		values[count] = value
+		count++
+	}
+	return
+}
+
+// UpdateKey will get values for the provided key and update the to the new key.
```
estroz, line 111:
```suggestion
// UpdateKey will move values under oldKey to newKey in Assocations.
```

pkg/image/association.go
```diff
@@ -116,36 +210,43 @@ func (as *Associations) Decode(r io.Reader) error {
...
-		}
-		switch a.Type {
-		case TypeInvalid:
-			// TypeInvalid is the default value for the concrete type, which means the field was not set.
-			errs = append(errs, fmt.Errorf("image %q: must set image type", a.Name))
-		default:
-			errs = append(errs, fmt.Errorf("image %q: unknown image type %v", a.Name, a.Type))
-		}
+	for _, imageName := range as.Keys() {
+		assocs, _ := as.Search(imageName)
```
estroz, line 226: This error should be checked right?

pkg/image/association.go
```diff
@@ -35,8 +35,18 @@ func (e *ErrInvalidComponent) Error() string {
...
+// AssociationSet is a mapped for Association
+// searching
+type AssociationSet map[string]Association
+
+// Associations is a set of image Associations
+// mapped to their images
+type Associations map[string]AssociationSet
+
+// NewAssociations instantiates a new multimap.
+func NewAssociations() Associations {
```
estroz, line 47: I don't think this constructor is adding much over the bare `a := AssociationSet{}`, unless you intend on adding parameters at some point?

pkg/image/association.go
jpower432: That is actually a boolean, but I could just do the check and generate a new error if the key isn't found in the map.

pkg/image/association.go
jpower432: That's a fair point. I don't expect that parameter will come into play at any point here.

pkg/image/association.go
```diff
@@ -174,11 +275,12 @@ func ReadImageMapping(mappingsPath string) (map[string]string, error) {
 	return mappings, scanner.Err()
 }
 
-func AssociateImageLayers(rootDir string, imgMappings map[string]string, images []string) (Associations, utilerrors.Aggregate) {
+func AssociateImageLayers(rootDir string, imgMappings map[string]string, images []string, typ ImageType) (Associations, utilerrors.Aggregate) {
```
estroz, line 278: :+1:

pkg/bundle/publish/publish.go
```diff
@@ -185,223 +174,187 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
+			}
+			// The mirrorer is not a fan of accepting an image ID when a tag symlink is available
+			// for some reason.
+			// TODO(estroz): investigate the cause of this behavior.
+			if assoc.TagSymlink == "" {
+				m.Source.Ref.ID = assoc.ID
 			} else {
-				logrus.Infof("skipping %v", assoc)
+				if err := unpack(filepath.Join(manifestPath, assoc.TagSymlink), o.tmp, filesInArchive); err != nil {
+					errs = append(errs, err)
```
estroz, line 267:
```suggestion
					errs = append(errs, err)
					continue
```

pkg/bundle/publish/publish.go
```diff
@@ -185,223 +174,187 @@ func (o *Options) Run(ctx context.Context, cmd *cobra.Command, f kcmdutil.Factor
...
+
+		genericMappings := []imgmirror.Mapping{}
+		releaseMapping := imgmirror.Mapping{}
+		missingLayers := map[string][]string{}
+		icspMapping := make(map[imagesource.TypedImageReference]imagesource.TypedImageReference)
+
+		values, _ := assocs.Search(imageName)
+
+		// Create temp workspace for image processing
+		cleanup, err := o.mktempDir()
```
estroz, line 198: I would just `if !o.SkipCleanup { defer cleanup() }` here since a cond branch that runs cleanup might never be entered.

estroz: Also this overwrites `o.tmp`, which is an unexpected mutation that could lead to bugs later. I'd rewrite this function to return the dir name, or use the preexisting `o.tmp`.

pkg/image/association.go
jpower432: Resolving. Pointer as map key removed.

pkg/image/association.go
jpower432: Resolving. Removed the NewAssociations function.

estroz (approved):

PR #88, opened 2021-09-17
feat(operators): rebuild, push catalog images during publish

### LINKED ISSUE #85: operator catalog diffs are not built into images nor combined with disconnected catalogs
When I run `create full` with a config that has `headsOnly: true` for some catalog, or `create diff`, I expect that diff to be mirrored to disk as an image. I also expect `publish` to combine that diff with an existing catalog image (pulled from data in a CatalogSource) and pushed to the catalog's registry under a desired tag (defaulting to the tag from the existing catalog image).

Currently this functionality is broken/does not exist. Since `opm diff` libraries only generate an file-based catalog, a new image must be built containing this file then mirrored. The same needs to happen during `publish`, but with the output of `opm render <existing catalog image> <mirrored diff catalog image>`.

estroz:
See #85 for details. An important implementation detail: catalog images themselves are no longer mirrored in `create`. Instead the images specified in the file-based catalog (DC) packaged into archive(s) are mirrored and treated as generic images, and `publish` will build and push catalog image directly using `docker buildx build --push`.

Closes #85

### DIFF
--- a/pkg/bundle/publish/manifests.go
+++ b/pkg/bundle/publish/manifests.go
@@ -18,21 +18,31 @@ import (
 )
 
 // Copied from https://github.com/openshift/oc/blob/5d8dfa1c2e8e7469d69d76f21e0a166a0de8663b/pkg/cli/admin/catalog/mirror.go#L549
-// Changes made are breaking ICSP and Catalog Source generation into difference functions
+// Changes made are breaking ICSP and Catalog Source generation into different functions
 func WriteICSP(out io.Writer, dir string, icsps [][]byte) error {
 
 	if err := ioutil.WriteFile(filepath.Join(dir, "imageContentSourcePolicy.yaml"), aggregateICSPs(icsps), os.ModePerm); err != nil {
 		return fmt.Errorf("error writing ImageContentSourcePolicy")
 	}
 
-	fmt.Fprintf(out, "wrote mirroring manifests to %s\n", dir)
+	fmt.Fprintf(out, "wrote ICSP manifests to %s\n", dir)
 
 	return nil
 }
 
 func WriteCatalogSource(out io.Writer, source imagesource.TypedImageReference, dir string, mapping map[imagesource.TypedImageReference]imagesource.TypedImageReference) error {
 
-	catalogSource, err := generateCatalogSource(source, mapping)
+	dest, ok := mapping[source]
+	if !ok {
+		return fmt.Errorf("no mapping found for index image")
+	}
+
+	return writeCatalogSource(out, source, dest, dir)
+}
+
+func writeCatalogSource(out io.Writer, source, dest imagesource.TypedImageReference, dir string) error {
+
+	catalogSource, err := generateCatalogSource(source, dest)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


pkg/operator/mirror.go
```diff
@@ -430,103 +445,67 @@ func (o *MirrorOptions) associateDeclarativeConfigImageLayers(ctlgRef imagesourc
...
-		}
-
-		for _, err := range []error{
-			// NB(estroz): assuming that any related image will not point to
-			// an operator catalog or bundle may not be a valid assumption.
-			// Even if not valid, mirroring bundle and catalog images as normal images
-			// should work as though they were mirrored via the `catalog mirror` API.
-			associateWithType(imgMappings, relatedImages, image.TypeOperatorRelatedImage),
-			associateWithType(imgMappings, bundleImages, image.TypeOperatorBundle),
-			associateWithType(imgMappings, []string{ctlgRef.Ref.Exact()}, image.TypeOperatorCatalog),
```
jpower432, line 456: @estroz We maybe have to keep the `TypeOperatorCatalog` differentiation in here because that is how we are creating the catalog sources on the publishing side. That is why they (catalog sources) are not being created on this branch right now.

jpower432 (changes requested):
Overall LGTM, just seeing the one issue with Catalog Sources being generated.

pkg/operator/mirror.go
estroz: I can fix this by generating CS's for each catalog image rebuilt.

pkg/bundle/publish/manifests.go
```diff
@@ -57,14 +67,11 @@ func GenerateICSP(out io.Writer, name string, byteLimit int, icspScope string, m
...
 		},
-		Spec: operatorv1alpha1.ImageContentSourcePolicySpec{
-			RepositoryDigestMirrors: []operatorv1alpha1.RepositoryDigestMirrors{},
-		},
 	}
 
 	for key := range registryMapping {
 		icsp.Spec.RepositoryDigestMirrors = append(icsp.Spec.RepositoryDigestMirrors, operatorv1alpha1.RepositoryDigestMirrors{
-			Source:  name,
+			Source:  key,
```
jpower432, line 74: @estroz Question here. I changed this from `key` to `name` so the `source` on the ICSP would reflect original image name. The key in the mapping is the location on disk so the source is coming out like this 
```  repositoryDigestMirrors:
  - mirrors:
    - localhost:5001/baz-operator/baz
    source: baz-operator/baz
```
Is this correct for catalogs?

estroz: It's almost correct; using `name` won't work sometimes because that name may contain a tag, and ICSP sources should not contain a tag or digest afaik. ICSP's should look like:

```yaml
  - mirrors:
    - localhost:5001/baz-operator/baz
    source: quay.io/baz-operator/baz
```

I think we need to add one more component to the association type for the pre-mirror fq image name, which I'll do in a follow-up. Thoughts?

jpower432: We are actually capturing that with the key in the AssociationSet. Checkout this line, there may be better ways to do this, but I am only adding an association to the `icspMapping` is the tag field is empty. https://github.com/RedHatGov/bundle/blob/74677b6fa3af6d899eec816ff95306494047cfc6/pkg/bundle/publish/publish.go#L318

estroz: Actually we could just use `name` but would need to parse it as to remove tag/digest. I'll whip something up.

jpower432 (approved):

PR #98, opened 2021-09-22
fix(publish): do not return an error when no catalogs are present

### LINKED ISSUE #95: bug: publishing fails if no catalogs are being mirrored
There is an issue within `publish.go` where if the imageset archive does not contain a catalog the publish operation will fail. We need to add a condition to check for the existence of catalogs before we try to rebuild the image.

estroz:
Closes #95 


jpower432 (approved):
LGTM. Tested without catalog and the issue is resolved.

Issue #72, opened 2021-09-05
Output of create diff does not update imageset file name


afflom:
When running create full, the output imageset name is bundle_000000.tar

When running create diff in the same workspace, the output imageset name is the same (bundle_000000.tar).

As a user, I would expect the name of the imageset to reflect the sequence number of the creation.

jpower432:
@afflom The new prefix on imageset is `bundle_seq<sequence-number>_00000.tar`. Do you think that change is sufficient enough for this ticket to be closed?

PR #99, opened 2021-09-22
feat(create): handle multiple architectures

jpower432:
### Features Added
Added support for multiple architecture for user-defined images and releases
Fixed the prefix for the tar archive to resolve ticket #72 

### Removed
Removed release_test.go test and moved `GetChannelLatest` into `cincinnati.go`. Used the test from `oc` and created a similar test for `GetChannelLatest`


pkg/bundle/release.go
```diff
@@ -6,64 +6,61 @@ import (
...
 // NewReleaseOptions defaults ReleaseOptions.
-func NewReleaseOptions(ro cli.RootOptions) *ReleaseOptions {
-	return &ReleaseOptions{RootOptions: ro}
-}
+func NewReleaseOptions(ro cli.RootOptions, flags *pflag.FlagSet) *ReleaseOptions {
+	var arch string
+	opts := ro.FilterOptions
+	opts.Complete(flags)
+	if opts.IsWildcardFilter() {
+		arch = runtime.GOARCH
```
estroz, line 49: This behavior is a little weird. If the user wants to filter no platforms, then the local platform is used, which is not what was asked for. Is there a better way to use a wildcard to determine arch, like query `api.openshift.com` for available platforms or hardcode a known list of platforms?

pkg/bundle/release.go
```diff
@@ -6,64 +6,61 @@ import (
...
+	} else {
+		arch = strings.Split(opts.FilterByOS, "/")[1]
```
estroz, line 51: I think users can pass something like `linux/arm/*` which would break this parser. In this case I would also either query for available platforms or match against a hard-coded list.

pkg/bundle/release.go
jpower432: @estroz Very good point here. I can totally see where you are coming from. I was kind of trying to recreate the options use on `oc release`, but I can see how that I was toying with the idea of of using the `oc image info` or `oc release info` library to gather that info from the API response.

afflom: If arch is omitted from the api query, the multi-arch images are returned

jpower432: Then for now I can make that query parameter to the API server optional, but that doesn't really solve the issue for @estroz's other comment with the example `linux/arm/*`. We probably should have knowledge of what is support so we can reported back invalid inputs as well.

estroz (approved):

PR #103, opened 2021-09-25
[wip] Research buildah for building new catalog images

### LINKED ISSUE #96: spike: research buildah for building new catalog images
As as admin, I was to use a daemonless options to rebuild the catalog image. Currently we are building it with `docker buildx` being an environmental dependency.

afflom:
Adds buildah bud and push to catalog_images.go with unit tests.

A little rough because of using bud instead of buildah native builder, but I figured that it would be nice to preserve the dockerfile for later ref if needed. 

I don't have a good way of testing/verifying the intended behavior beyond the included unit test. 

Let's discuss in draft! Any direction here would be helpful. 

closes #96


afflom:
buildah import requires `dnf install btrfs-progs-devel gpgme-devel device-mapper-devel` to compile. How do I add this to the tests?

jpower432:
> buildah import requires `dnf install btrfs-progs-devel gpgme-devel device-mapper-devel` to compile. How do I add this to the tests?

Maybe just add the installation of those packages as a step in the Makefile and make sure it runs before any compiling steps. It will have to detect the package manager though, using Ubuntu with the CI.

afflom:
Buildah reexecs the main process which results in imageset sequence errors. We will continue to shell to buildx until we find another suitable solution for adding arbitrary artifacts and layers to the index images.

PR #107, opened 2021-10-05
fix(create): Bug/issue 106

jpower432:
- Handles errors when pinning images in catalog mirror when bundles have images with no tags

### DIFF
--- a/pkg/operator/mirror.go
+++ b/pkg/operator/mirror.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containerd/containerd/reference"
 	"github.com/joelanford/ignore"
 	imgreference "github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/oc/pkg/cli/admin/catalog"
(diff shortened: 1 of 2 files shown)


pkg/operator/mirror.go
```diff
@@ -325,13 +326,21 @@ func pinImages(ctx context.Context, dc *declcfg.DeclarativeConfig, resolverConfi
 	for i, b := range dc.Bundles {
 		if !isImagePinned(b.Image) {
 			if dc.Bundles[i].Image, err = image.ResolveToPin(ctx, resolver, b.Image); err != nil {
+				if errors.As(err, &reference.ErrObjectRequired) {
```
estroz, line 329: It would be more efficient to just check if `b.Image == ""`, ex
```go
if b.Image == "" {
    logrus.Warnf("bundle %s: bundle image not set", b.Name)
    continue
}
if !isImagePinned(b.Image) {
   ...
```

estroz (changes requested):
One suggestion

estroz (approved):

Issue #104, opened 2021-09-28
Add buildah


afflom:
Import buildah in place of relying on external buildx.

PR #108, opened 2021-10-06
spike/podman

afflom:
The results of oras, buildx, podman, buildah research. We have arrived at podman as an acceptable shim that we have high confidence will be on the majority of systems running this for testing. We will address arbitrary file placement in container images in alpha3.


afflom:
@jpower432  This is ready for your review

afflom:
closes #104

Issue #102, opened 2021-09-23
Add OKD as ocp channel


afflom:
As a user, I would like to be able to mirror okd release content. OKD release payload info can be discovered at: `https://origin-release.ci.openshift.org/graph`. An easy way to add this functionality would be to use the following logic:

pseudo code:
```
if spec:mirror:ocp:channel = okd
then use alternate api uri
parse cincinnati as usual
fi
```

PR #117, opened 2021-10-11
Okd graph support

afflom:
This seems to work, but I dont like the regex in the associations. How can we solve this in the mappings so that we can keep associations generic to all images?

### DIFF
--- a/pkg/image/association.go
+++ b/pkg/image/association.go
@@ -301,6 +301,7 @@ func AssociateImageLayers(rootDir string, imgMappings map[string]string, images 
 		}
 		tagOrID := dirRef[idx+1:]
 		dirRef = dirRef[:idx]
+		//logrus.Infof("tagOrID: %s \n dirRef: %s", tagOrID, dirRef)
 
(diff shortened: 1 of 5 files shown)


afflom:
closes #102

pkg/image/association.go
```diff
@@ -300,7 +304,34 @@ func AssociateImageLayers(rootDir string, imgMappings map[string]string, images
 			idx = idIdx
 		}
 		tagOrID := dirRef[idx+1:]
+
+		// afflom - Origin mappings cutoff the arch part of the filename.
+		// this adds it back in as needed.
```
jpower432, line 309: @afflom rather than including okd specifc login in `associations.go` I think this should be done when building the mapping passed to `AssociateImageLayer`. WDYT?

afflom: @jpower432 , I agree. I never wanted this in association.go, but I don't understand how to implement this in bundle.getMapping()

afflom:
@jpower432 Works as intended. Thank you for your help.

PR #137, opened 2021-10-21
chore: temp remove unit test

afflom:


jpower432 (approved):
Approving with the comment that we have an open PR to phase this test out completely. Removing to allow unit tests to run without calling `podman`

PR #235, opened 2021-12-11
fix(create): add metadata rollback for errors after metadata write in create operation

### LINKED ISSUE #200: [BUG] metadata is out of sync if archiving fails
# Version

```console
$ oc-mirror version
v0.1.0-alpha.3
```

# What happened?

If the archiving operation fails, the metadata has already been written to its final destination and causes out-of-sequence errors.

# What you expected to happen?

The metadata is rolled back if there is a failure

```console
$ oc-mirror --config imagetset-config.yaml file://archives
```
Cancel before the archive finishes. The metadata will still be in place. 

# References

jpower432:
# Description

This PR will add functionality that will roll back the metadata in the event an error occurs during archiving or the user cancels during archiving. It also moves `testdata` for each package under the package root.

Fixes #200 

## Type of change

# How Has This Been Tested?

- [X] Unit test for Create with Cancel

Instructions to manually test:
1. Use stateless mode
2. Run a an oc-mirror mirror to disk operation
3. Cancel the operation during archiving
4. Ensure that when you run again there are not sequence errors

# Checklist:


afflom:
Rollback works, but archiving operations continue after interrupt. This solution needs to cleanup the interrupted PR and exit when an interrupt is received.

pkg/cli/mirror/create.go
```diff
@@ -117,16 +125,25 @@ func (o MirrorOptions) Create(ctx context.Context, flags *pflag.FlagSet) error {
...
+	go func() {
+		<-sigs
+		if err := rollbackMeta(); err != nil {
+			logrus.Error(err)
+		}
+		done <- true
+	}()
 
-	// Handle Committer backends.
-	if committer, isCommitter := backend.(storage.Committer); isCommitter {
```
estroz, line 126: Does this no longer need to be called?

jpower432: I think this was added when we were intending to use the Git backend, but right now we have no concrete types that implement that interface to I just remove this line.

estroz: Might be worth putting a comment in, but not a blocker.

pkg/cli/mirror/create.go
```diff
@@ -117,17 +125,33 @@ func (o MirrorOptions) Create(ctx context.Context, flags *pflag.FlagSet) error {
...
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	done := make(chan bool, 1)
+	go func() {
+		<-sigs
+		if err := rollbackMeta(); err != nil {
+			logrus.Error(err)
+		}
+		os.Exit(1)
+		done <- true
```
estroz, line 138: If the parent goroutine does not wait on `done` then this goroutine may not complete if triggered.

pkg/cli/mirror/create.go
```diff
@@ -117,17 +125,33 @@ func (o MirrorOptions) Create(ctx context.Context, flags *pflag.FlagSet) error {
...
+	// Catch any cancel signals
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	done := make(chan bool, 1)
+	go func() {
+		<-sigs
+		if err := rollbackMeta(); err != nil {
+			logrus.Error(err)
+		}
+		os.Exit(1)
```
estroz, line 137: Does the signal notifier channel catch this? Probably not, but I’m not sure. I would simply let this goroutine run to completion, block on `done` or the function’s completion using `select`, and call `os.Exit(1)` if `done` is not empty.

jpower432: @estroz good advice. Going to back to the drawing board on this. I was really using the goroutine as a way to test the metadata rollback functionality without giving a lot of thought to the implementation of that itself. Since I still think that we should handle cancel signals I will go back and incorporate your suggestions and add some unit tests for this.

pkg/cli/mirror/create.go
```diff
@@ -124,17 +130,30 @@ func (o MirrorOptions) Create(ctx context.Context, flags *pflag.FlagSet) error {
...
-	// Run archiver
+	// Allow for for user interrupts
+	ctx, done := o.CancelContext(ctx)
+	defer done()
+
+	// If any errors occur after the metadata is written
+	// initiate metadata rollback
 	if err := o.prepareArchive(ctx, cfg, backend, thisRun.Sequence, manifests, blobs); err != nil {
+		if err := rollbackMeta(); err != nil {
+			return fmt.Errorf("error occurred during metadata rollback")
```
estroz, line 141:
```suggestion
			return fmt.Errorf("error occurred during metadata rollback: %v", err)
```

pkg/cli/mirror/options.go
```diff
@@ -43,3 +52,41 @@ func (o *MirrorOptions) BindFlags(fs *pflag.FlagSet) {
...
+	o.cancelCh = makeCancelCh(syscall.SIGINT, syscall.SIGTERM)
+}
+
+// CancelContext will return a cancellable context and listen for
+// cancellation signals
+func (o *MirrorOptions) CancelContext(parent context.Context) (context.Context, context.CancelFunc) {
+	if o.cancelCh == nil {
+		// return unchanged parent when cancel channel is
+		// not initialized
+		return parent, func() {}
```
estroz, line 66: This seems like a source of future bugs. I would either panic here, or make `init()` atomic with `sync.Once` and call it within `CancelContext()` so the channel is always set up.

jpower432: This is a really good poiint. I can see how not returning a cancellable context from this would cause issues.

estroz (commented):
/lgtm

PR #277, opened 2022-01-20
general docs cleanup

afflom:
- Bumps ocp release version refs to 4.9
- Expanded example descriptions
- Add troubleshooting
- Misc housekeeping

### DIFF
--- a/docs/design/overview.md
+++ b/docs/design/overview.md
@@ -3,18 +3,18 @@
 # oc-mirror Design Overview
 
 ## Summary 
-oc-mirror assists with lifecycle management of internet-disconnected OpenShift environments by performing differential update operations and synchronization between internet-connected and internet-disconnected environments.   
+oc-mirror assists with lifecycle management of mirror registries that support OpenShift environments by performing differential update operations and synchronizations.   
   
 ## Features
 1. Differential Synchronization of OpenShift Release Images
 2. Differential Synchronization of Operator images
-3. Synchronization of OpenShift cincinnati graph data image
+3. Synchronization of OpenShift cincinnati graph data image (planned)
 4. Synchronization of OpenShift Release images by version or channel latest
-5. Synchronization of Operator Images by full catalog and full catalog at latest operator versions. Also, allows for synchronization of individual operators by version or channel head.
+5. Synchronization of Operator Images by full catalog and full catalog at latest operator versions. Also, allows for synchronization of individual operators by version to channel(s') HEAD.
 6. Synchronization of helm charts
 7. Synchronization of additional images
 8. Remote state backends
-9. Base image filtering
+9. Base image filtering (planned)
 10. (multi) Cluster resource management
 
 ## Design Considerations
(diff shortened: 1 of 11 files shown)


docs/design/overview.md
```diff
@@ -3,26 +3,26 @@
...
 4. User Workloads -  oc-mirror synchronizes helm charts and arbitrary images to streamline the importation of user workloads into their OpenShift deployments. 
-5. Internet Connected Experience - Operator Lifecycle Manager, OpenShift Update Service, Advanced Cluster Management, and oc-mirror; in concert, can facilitate user experience parity with that of internet-connected environments. This ux is achieved by updating cluster resources during the publishing phase of oc-mirror. 
+5. Internet Connected Experience - Operator Lifecycle Manager, OpenShift Update Service, Advanced Cluster Management, and oc-mirror; in concert, can facilitate user experience parity with that of internet-connected environments. This ux is achieved by updating cluster resources during the publishing phase of oc-mirror(planned).
```
dinhxuanvu, line 25: oc-mirror (planned) (missing a whitespace)

docs/examples/imageset-config-release.yaml
```diff
@@ -1,10 +1,12 @@
...
   ocp:
     channels:
-      - name: stable-4.8
-      - name: stable-4.7
+      - name: stable-4.9
         versions:
-          - 4.7.36
-    graph: true
\ No newline at end of file
+        - 4.9.13
```
dinhxuanvu, line 12: I think this indentation is not correct, is it?

dinhxuanvu: Also, the `graph: true` is removed here. I assume the field is deprecated/removed?

docs/design/overview.md
```diff
@@ -3,26 +3,26 @@
...
+oc-mirror assists with lifecycle management of mirror registries that support OpenShift environments by performing differential update operations and synchronizations.   
   
 ## Features
 1. Differential Synchronization of OpenShift Release Images
 2. Differential Synchronization of Operator images
-3. Synchronization of OpenShift cincinnati graph data image
+3. Synchronization of OpenShift cincinnati graph data image (planned)
 4. Synchronization of OpenShift Release images by version or channel latest
-5. Synchronization of Operator Images by full catalog and full catalog at latest operator versions. Also, allows for synchronization of individual operators by version or channel head.
+5. Synchronization of Operator Images by full catalog and full catalog at latest operator versions. Also, allows for synchronization of individual operators by version to channel(s') HEAD.
```
dinhxuanvu, line 13: Not sure why there are parentheses in `channel(s')`?

dinhxuanvu (commented):
Just a few nits.

docs/design/overview.md
jpower432: Yea I think with the `'` it looks odd. I know typically that notation is used to convey one or more.

docs/examples/imageset-config-helm.yaml
```diff
@@ -8,7 +12,4 @@ mirror:
         charts:
           - name: podinfo
             version: 5.0.0
-    local:
```
jpower432, line 11: Why remove the local example?

docs/examples/imageset-config-release.yaml
afflom: Weird, the indentation works either way in testing, but I think that you are right about the syntax being technically incorrect as written here.

afflom: We do not use the graph key yet.

docs/design/overview.md
afflom: I dont know how to say this in a coherent way that is close to grammatically correct. I'll try again

dinhxuanvu: I think what you wrote here is fine. I just wonder what parentheses are for? You can just move them I think.

docs/examples/imageset-config-helm.yaml
afflom: I validated that each of these examples were functional. I'll add local back as commented so that it can still be functional while referencing the local example.

docs/design/overview.md
afflom: Good catch. Fixed

docs/examples/imageset-config-filter-catalog.yaml
```diff
@@ -1,10 +1,23 @@
...
 mirror:
   ocp:
   operators:
-    - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.8
+    - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.9
       headsOnly: false
       packages:
         - name: elasticsearch-operator
-          startingVersion: '5.2.3-31'
+          startingVersion: 5.3.2-20
```
dinhxuanvu, line 21: I'm curious if this non-quoted string will cause a problem during unmarshal. Sometimes, this string is interpreted as decimal numbers.

afflom: I tested with literals and without. The outcome was the same, but I included the literals in the example as you had suggested.

dinhxuanvu (commented):
/lgtm

PR #287, opened 2022-01-27
Stopping the process when no changes are found between runs. [190]

### LINKED ISSUE #190: Do not generate imageset on no updates
# Version
Alpha3

# What happened?

No updates were mirrored during imageset creation, but an empty imageset was generated.

# What you expected to happen?

If there was nothing mirrored during creation, I would not expect an imageset to be generated.

run oc-mirror with a single operator selected that would normally return 404 (missing) and add --skip-missing as a command line flag.

This is problematic because it updates the imageset sequence number and forces the user to publish an empty imageset.

ianwatsonrh:
# Description

Checks if there are any blob differences during the process. If there are no changes stop the process to avoid unnecessary sequence update / tar files with only the manifests in. 

Fixes #190

## Type of change

# How Has This Been Tested?

- [x] Unit tests updated
- [x] Run yourself via two runs of oc-mirror --config imageset-config.yaml file://archives and oc-mirror --config imageset-config.yaml file://archives2. Note that in archives2 there are no tar files. 

# Checklist:

### DIFF
--- a/pkg/cli/mirror/create.go
+++ b/pkg/cli/mirror/create.go
@@ -20,6 +20,11 @@ import (
 	"github.com/openshift/oc-mirror/pkg/metadata/storage"
 )
 
+var (
+	// NoUpdatesExist should be returned by Create() when no updates are found
+	NoUpdatesExist = errors.New("no updates detected, process stopping")
+)
+
 func (o *MirrorOptions) Create(ctx context.Context) error {
 
 	// Read the imageset-config.yaml
@@ -126,6 +131,13 @@ func (o *MirrorOptions) Create(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	// Stop the process if no new blobs
+	if len(blobs) == 0 {
+		logrus.Infof("no updates detected, process stopping")
+		return NoUpdatesExist
+	}
+
 	// Add only the new manifests and blobs created to the current run.
 	thisRun.Manifests = append(thisRun.Manifests, manifests...)
 	thisRun.Blobs = append(thisRun.Blobs, blobs...)
--- a/pkg/cli/mirror/create_test.go
+++ b/pkg/cli/mirror/create_test.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"io/ioutil"
+	"strings"
 
 	"github.com/openshift/oc-mirror/pkg/cli"
 	"github.com/openshift/oc-mirror/pkg/config/v1alpha1"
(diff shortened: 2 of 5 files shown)


pkg/cli/mirror/create_test.go
```diff
@@ -132,3 +179,19 @@ func TestCreateWithCancel(t *testing.T) {
 
 	require.Equal(t, true, opts.interrupted)
 }
+
+func GenerateConfigPath(path string, baseFile string, targetFile string) {
```
jpower432, line 183: Nit: This function could be private if it will just be used in this test file

ianwatsonrh: Ack. Changed to private.

pkg/cli/mirror/create.go
```diff
@@ -126,6 +126,12 @@ func (o *MirrorOptions) Create(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	// Stop the process if no new blobs
+	if len(blobs) == 0 {
```
jpower432, line 131: @ianwatsonrh One thing I would suggest here is having a logging message to signal to the user that an image will not be created. Something like "no updates detected". Other than that it LGTM.

ianwatsonrh: Ack. Log line added.

afflom:
Very cool! It works as advertised. I only noticed one issue when I tested it mirror2mirror. The log output might be a bit confusing to users. Can you please configure it to exit without calling the publishing functions when using mirror2mirror?: 
```
INFO no updates detected, process stopping        
INFO Publishing image set from archive "./mirrortmp468199673" to registry "registry.test.com:5000" 
DEBUG Unarchiving metadata into oc-mirror-workspace/images.178718783 
Image Publish:
ERROR: publishing operation failed: no archives found in directory ./mirrortmp468199673
To retry this operation run "oc-mirror --from ./mirrortmp468199673 docker://registry.test.com:5000
```

jpower432:
> 

@afflom good catch on the mirror2mirror error @ianwatsonrh Maybe instead of returning nil, you could return a custom error that can be handled by the caller since we use the `Create` function in multiple places.

ianwatsonrh:
> @afflom good catch on the mirror2mirror error @ianwatsonrh Maybe instead of returning nil, you could return a custom error that can be handled by the caller since we use the `Create` function in multiple places.

Ack. 

Custom error thrown instead of nil. Caller then handles the error. 

New e2e-simple test case added to reflect.

afflom:
Spot on. Give me the weekend to play with it and I'll get back to you.

jpower432:
Tested and everything works as expected
/lgtm

PR #355, opened 2022-03-11
Ci/add ci targets

jharmison-redhat:
This adds the integration test target that will be consumable by OpenShift CI.

It also temporarily validates secrets and will need a manual test run once [the config and jobs](https://github.com/openshift/release/pull/26915) are merged. After that test run it  will need force pushed with some more tweaks to conform to the secrets.


jpower432 (approved):
/lgtm

PR #360, opened 2022-03-14
Bug 2063955: fix(mirror): adds back continue-on-error use for missing images

jpower432:
# Description

This PR adds fixes for errors that occur when missing images are provided in the config (via catalog or additional/helm). This allows `continue-on-error` to log a warning about the missing images instead of throwing an error if specified.

Related to #359 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Unit test added
- [X] Manual test

# Checklist:

### DIFF
--- a/pkg/cli/mirror/mirror.go
+++ b/pkg/cli/mirror/mirror.go
@@ -240,18 +240,21 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error) 
 		// Create assocations
 		assocDir := filepath.Join(o.Dir, config.SourceDir)
 		assocs, errs := image.AssociateLocalImageLayers(assocDir, mapping)
-		skipErr := func(_ error) bool {
+
+		skipErr := func(err error) bool {
 			ierr := &image.ErrInvalidImage{}
+			cerr := &image.ErrInvalidComponent{}
+			return errors.As(err, &ierr) || errors.As(err, &cerr)
+		}
+
+		if errs != nil {
 			for _, e := range errs.Errors() {
-				if !errors.As(e, &ierr) {
-					return false
+				if err := o.checkErr(e, skipErr); err != nil {
+					return err
 				}
 			}
-			return true
 		}
-		if err := o.checkErr(errs, skipErr); err != nil {
-			return err
-		}
+
 		// Pack the images set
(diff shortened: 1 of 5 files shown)


afflom (approved):
I was able to mirror the 4.9 red hat catalog at HEADs with `--continue-on-error` over to you for code review @deejross

jpower432:
@soltysh PTAL (priority 3). Thanks!

pkg/cli/mirror/mirror.go
```diff
@@ -504,3 +516,22 @@ func (o *MirrorOptions) generateAllICSPs(mapping image.TypedImageMapping, dir st
 
 	return WriteICSPs(dir, allICSPs)
 }
+
+func (o *MirrorOptions) checkErr(err error, acceptableErr func(error) bool) error {
+	if err != nil {
```
soltysh, line 521: Fail fast:
```golang
if err == nil {
    return 
}
```
is preferred over what you have here.

pkg/cli/mirror/mirror.go
```diff
@@ -233,8 +233,19 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error)
 		// Create assocations
 		assocDir := filepath.Join(o.Dir, config.SourceDir)
 		assocs, errs := image.AssociateImageLayers(assocDir, mapping)
-		if errs != nil {
-			return errs
+
+		// Instead of returning an error, just log it.
+		skipErr := func(_ error) bool {
+			ierr := &image.ErrInvalidImage{}
+			for _, e := range errs.Errors() {
```
soltysh, line 240: Are you sure you want to iterate over `errs` returned above in line 235 and not over error passed as input to this method? If you really do want that, then I'd suggest changing these invocations, the way it currently reads is super confusing.

soltysh (changes requested):

pkg/cli/mirror/mirror.go
jpower432: Great point. Updated.

pkg/cli/mirror/mirror.go
jpower432: I can see how looping in the anonymous function could be confusing here. Updated this to make it more straightforward. Thanks for the feedback!

jpower432:
@soltysh PTAL. Thanks!

PR #362, opened 2022-03-16
[CI] Update CI targets for use

jharmison-redhat:
# Description

- Ensures that the cleanup will occur even on test failure (while still reporting failure)
- Configures secrets for CI use

Fixes # (issue)

## Type of change

# How Has This Been Tested?

The test job configured on openshift/release.

# Checklist:

# TODO

- [x] Fix hanging `oc-mirror` runs from CI pipeline.
  - See https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oc-mirror/362/pull-ci-openshift-oc-mirror-main-integration/1504563114840952832#1:build-log.txt%3A599 for an example of the bad behavior.
    - Investigation of the integration environment during this run revealed that `oc-mirror` is successfully running, publishing the imageset, and exiting with a return code of 0 but Ansible does not appear to be picking it up and hangs waiting for it to return.
  - It appears that the issue may be related to the Ansible version, and the appropriate response would be a better test-runner image with a Dockerfile in the openshift/release pipeline.
  - To confirm this, I'd like to run some local tests with RHEL 8.4, Python 3.6, and Ansible 4.10 to confirm I can replicate this behavior before creating a new runner image.
- [x] Artifact recovery needs to be ironed out. ${ARTIFACT_DIR} is the appropriate place to store these artifacts. Secret obfuscation is already being done by ci-operator, so if we dump raw logs (without compression) it should be safe to recover the following:
  - [ ] Debug Ansible logs
  - [x] The entire `output` directory
    - the installer directory, including tfstate and logs for the cluster installation
    - registry and certificate information
    - infrastructure tfstate, including the registry, bastion, and proxy
  - [x] The `oc-mirror` logs from the connected host (These are not yet recovered by the Ansible, but should probably be handled there to improve local debugging of the normal `test-integration` target as well.
- [x] The cleanup process needs to tolerate cancellation in OpenShift CI.
  - As of right now, a failed test will clean up properly but a cancelled test (e.g. new pushes are made to the PR while a previous test is running) will not and resources will be orphaned.
  - This is mitigated by the artifact recovery above, as it will enable manual `terraform destroy` of `tfstate` from an interrupted run, but this is less than ideal.
  - I have [a question](https://coreos.slack.com/archives/CBN38N3MW/p1647625906561009) in to #form-testplatform on CoreOS Slack for this. I believe the best answer in the short term is to use a `literal_step` for `post` in the `test` config to do this cleanup. It may be that  this is infeasible without a proper step-registry implementation.


jharmison-redhat:
Rebased on main

/test integration

jharmison-redhat:
@jpower432 @afflom I'm not sure what the right order to merge these will be. As of right now, the integration test can't be run with a slash command until the [release PR](https://github.com/openshift/release/pull/27189) is merged, but I can't fully test out the rehearsal on the release PR until elements of this get merged.

If you're okay with the limited impact to the rest of the codebase on this PR, I think it makes the most sense to merge this in the state it's in and then retrigger the rehearsal on main before rebasing other PRs to validate that the slash command works as expected.

jharmison-redhat:
/hold

one more fix for teardown - oops

afflom (approved):

PR #376, opened 2022-03-18
cleanup: API types separation from methods and bool pointer removal from API

### LINKED ISSUE #367: 4.11 Req: Remove pointers for boolean types
Remove *bool in https://github.com/openshift/oc-mirror/blob/2edebb44f72e340e004067a98f9da6e3a7303837/pkg/config/v1alpha1/config_types.go#L66

The general API guidelines are against using pointers for boolean types.
If the only reason is to have nil be defaulted to true you should change
the field name to be opposite, AllPackages, for example?

### LINKED ISSUE #371: 4.11 Req: Separate types from methods/helpers
Have a harder split between API types and actual method, helpers, etc.
pkg/config, pkg/metadata, pkg/image should probably all be part of pkg/api/version with
just types_config.go, types_metadata.go and types_image.go containing just the
types and necessary conversions, schemas, etc.

jpower432:
# Description

This PR organizes API types and methods into separate packages and remove bool pointer `headsOnly` keys

Fixes #371 
Fixes #367 

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

# Checklist:

- [X] I have made corresponding changes to the documentation

/kind cleanup


jpower432:
cc @afflom @soltysh @deejross PTAL. Thanks!

Just a note - many of the line adds and deletions are comments added to public API types that were missing one. I separated that out on the last commit.

docs/examples/imageset-config-filter-catalog.yaml
```diff
@@ -15,7 +15,7 @@ mirror:
   ocp:
   operators:
     - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.9
-      headsOnly: false
+      allPackages: true
```
afflom, line 18: I think headsonly encompasses all packages at their channel heads, but allPackages doesnt encompass all channels within the packages. How about `full` or just `all`?

docs/examples/imageset-config-release-full-channel.yaml
```diff
@@ -7,4 +7,4 @@ mirror:
   ocp:
     channels:
       - name: stable-4.9
-        headsOnly: false
\ No newline at end of file
+        allVersions: true
```
afflom, line 10: allversions combined with `shortestPath` doesnt make much sense. I suggest `all` here as well. That also helps keep parity between operators and ocp releases.

jpower432: Perhaps we should create a validation that disallows the combination between `shortestPath` and `allVersions` or `all`. I doubt needing the `shortestPath` between the min and max of a channel would be a very common use case. WDYT?

docs/examples/imageset-config-filter-catalog.yaml
jpower432: `full` is fine with me.

dinhxuanvu:
This PR needs a rebase due to a recently-merged PR. Otherwise, it looks good.

dinhxuanvu (approved):
/lgtm

pkg/api/v1alpha2/types_config.go
```diff
@@ -0,0 +1,169 @@
...
+	// from the mirroring process if they exist in other content
+	// types in the configuration.
+	BlockedImages []BlockedImages `json:"blockedImages,omitempty"`
+	// Samples defines the configuration for Sample content types.
+	// This is currently not implemented.
+	Samples []SampleImages `json:"samples,omitempty"`
+}
+
+// OCP defines the configuration for OCP and OKD content types.
+type OCP struct {
```
soltysh, line 48: Nit: have you considered renaming this to something more readable, and less OCP related, for example `Product`? Also do you care about the distinction if that's OCP or OKD, if so maybe consider adding enum type allowing to discriminate that.

pkg/api/v1alpha2/types_config.go
```diff
@@ -0,0 +1,169 @@
...
+// and image mirroring
+type Helm struct {
+	// Repo is the helm repository containing the charts
+	Repos []Repo `json:"repos,omitempty"`
+	// Local is the configuration for locally stored helm charts
+	Local []Chart `json:"local,omitempty"`
+}
+
+// Repo defines the configuration for a Helm repo.
+type Repo struct {
```
soltysh, line 119: Nit: the general API guidelines (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions) I'd suggest renaming this to `Repositories`

pkg/api/v1alpha2/types_config.go
```diff
@@ -0,0 +1,169 @@
...
+// Image contains image pull information.
+type Image struct {
+	// Name of the image. This should be an exact image pin (registry/namespace/name@sha256:<hash>)
+	// but is not required to be.
+	Name string `json:"name"`
+}
+
+// AdditionalImages defines the configuration
+// individual image content types.
+type AdditionalImages struct {
```
soltysh, line 154: Each of the following 3 structures is identical, how about having a single structure, `Images` for example, and then in `Mirror` struct you'd just use:
```
	AdditionalImages []Images `json:"additionalImages,omitempty"`
	BlockedImages []Images `json:"blockedImages,omitempty"`
...
```

unless you can explain me the difference that you anticipate between these 3 structures.

soltysh: Less structures make the API cleaner and more readable :wink:

pkg/api/v1alpha2/types_config.go
jpower432: Ack. Will incorporate into a follow on PR.

pkg/api/v1alpha2/types_config.go
jpower432: Ack. Will incorporate it into a follow on PR.

pkg/api/v1alpha2/types_config.go
jpower432: I agree that the name of the type could be more general because we are supporting both OCP and OKD. We make the distinction between OCP and OKD from the channel name the OKD channel name is distinct.

PR #387, opened 2022-03-28
refactor: heads only workflow to remove the requirement to merge FBCs

### LINKED ISSUE #377: [RFE] headsOnly Rework
When headsOnly is annotated, write the HEAD version of each referenced  operator channel to the metadata as the referenced channel's startingVersion

Differential updates with headsOnly annotated, reference the startingVersion of each referenced operator channel.

jpower432:
# Description

When merging the catalog, unintentional changes to the upgrade and pruned bundles can be created in the target FBC generated by oc-mirror.
    
To remove the requirement to merge FBCs, package ranges are being used in metadata for heads-only workflows to generate FBCs that are scoped per the configuration instead of generating differential FBCs and merging with previous FBCs. The OCP releases heads only workflow has been updated work in the same way. To achieve the same optimization in regards to downloaded images, all images in the image mapping during each run are validated against the metadata to make sure they do not exist in PastAssociations.

### Workflows Conversion Behavior (Catalogs and Operators):

**Heads Only to Range**: Just the specified range will be downloaded the previous starting version between be erased.
**Range to Heads Only**: Previous range versions will not be recorded and the new Heads only starting version will be the current head.

Closes #377

Blocked by #364 

## Type of change

# How Has This Been Tested?

- [X] Unit tests for added methods
- [X] Manual testing of specified scenarios below

# Testing Scenarios
Pruned catalog test cases are being added to the E2E testing with PR #400.
1. Testing differential behavior when mirroring catalogs (new packages, new channels, pruned bundles)
2. Testing differential behavior when mirroring the latest releases ( a bit harder to validate due to the use of the Cincinnati API, can possibly replicate in a unit testing with a mock API)
3. Testing `IgnoreHistory` when using with image redownloads. Previously downloaded images in ranges are no longer redownloaded by default`. Implemented to replace previous heads only catalog behavior.

# Checklist:

# Notes to Reviewer
Changes to E2E bundles were made due to old refs no longer being passed during the differential operations. The older dependency bundles are no longer pruned out resulting in an additional bundle.

### DIFF
--- a/pkg/cli/mirror/catalog_images.go
+++ b/pkg/cli/mirror/catalog_images.go
@@ -128,7 +128,6 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage 
 	for ctlgRef, artifactDir := range catalogsByImage {
 		// Always build the catalog image with the new declarative config catalog
 		// using the original catalog as the base image
-		var layers []v1.Layer
 		var layoutPath layout.Path
 		refExact := ctlgRef.Ref.Exact()
 
@@ -153,12 +152,16 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage 
 		}
 
 		// Since we are defining the FBC as index.json, remove
-		// any .yaml files from the initial image to ensure they are not processed instead
-		deleted, err := deleteLayer("/configs/.wh.index.yaml")
+		// remove anything that may currently exist
+		deleted, err := deleteLayer("/.wh.configs")
 		if err != nil {
 			return fmt.Errorf("error creating deleted layer: %v", err)
 		}
-		layers = append(layers, add, deleted)
+
+		// Delete must be first in the slice
+		// so that the /configs directory is deleted
+		// and then add back with the new FBC.
+		layers := []v1.Layer{deleted, add}
 
 		layoutDir := filepath.Join(artifactDir, config.LayoutsDir)
 		layoutPath, err = imgBuilder.CreateLayout("", layoutDir)
--- a/pkg/cli/mirror/catalog_images.go
+++ b/pkg/cli/mirror/catalog_images.go
@@ -65,7 +65,7 @@ func (o *MirrorOptions) rebuildCatalogs(ctx context.Context, dstDir string) (ima
 		}
 
 		// From the index path determine the artifacts (index and layout) directory.
-		// Using that path determine the corresponding catalog image for processing.
+		// Using that path to determine the corresponding catalog image for processing.
 		slashPath := filepath.ToSlash(fpath)
 		if base := path.Base(slashPath); base == "index.json" {
 			slashPath = path.Dir(slashPath)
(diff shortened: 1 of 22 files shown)


pkg/cli/mirror/catalog_images.go
```diff
@@ -160,87 +140,30 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
-			renderedPath := filepath.Join(dcDirToBuild, "index.json")
-			f, err := os.Create(renderedPath)
-			if err != nil {
-				return err
-			}
-			defer f.Close()
-			if err := declcfg.WriteJSON(*dc, f); err != nil {
-				return err
-			}
+		add, err := builder.LayerFromPath("/configs", filepath.Join(artifactDir, config.IndexDir, "index.json"))
```
dinhxuanvu, line 150: It's not really something oc-mirror has control over but it always makes me slightly concerned that we have to rely on a filename to detect the FBC file and have to do the below cleanup to ensure we don't process wrong files. We need to ensure this `index.json` is something that the pipeline team will reliably use to produce their catalogsource images. Plus, technically OLM doesn't restrict to only one file per catalogsource. The catalogsource can have multiple FBC files which contains pieces of the whole FBC.

dinhxuanvu: I'm fine with what you are doing here. I just wanted to post my comment here as something to keep an eye on.

pkg/cli/mirror/mirror.go
```diff
@@ -223,6 +223,22 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error)
 			return err
 		}
 
+		prevAssociations, err := image.ConvertToAssociationSet(meta.PastAssociations)
```
dinhxuanvu, line 226: This block of code and the one below are identical. It would be nice if we can reduce the duplication here. Not a big deal. Just a nit.

pkg/metadata/store.go
```diff
@@ -28,22 +37,48 @@ func SyncMetadata(ctx context.Context, first storage.Backend, second storage.Bac
...
+	resolver, err := containerdregistry.NewResolver("", skipTLSVerify, plainHTTP, nil)
+	if err != nil {
+		return fmt.Errorf("error creating image resolver: %v", err)
+	}
+	cacheDir, err := os.MkdirTemp("", "imageset-catalog-registry-")
+	if err != nil {
+		return err
+	}
+
+	logger := logrus.New()
```
dinhxuanvu, line 66: Interesting. Look like this is just a placeholder log for the containderd stuff. Is there an option not to include the log option in the `NewRegistry()` method?

dinhxuanvu (commented):
Looking good. Just a few comments.

pkg/cli/mirror/catalog_images.go
jpower432: @dinhxuanvu `oc-mirror` is actually generating the `index.json` so we shouldn't have a problem there. Totally valid concern though when you consider that multiple files under `/configs` could be aggregated with what we are producing. We do remove any `index.yaml` that may have been under `/configs` originally, but it may be safer to ensure `/configs` is empty and the only FBC in the image is the one being produced by `oc-mirror`.

pkg/cli/mirror/mirror.go
jpower432: I will try to reduce the duplication here. I think there is some duplication between mirror to disk and mirror to mirror workflow options.

pkg/metadata/store.go
jpower432: I can add a comment here to clarify. We only use this to set the output to `ioutil.Discard` because the `containerd.Registry` was a bit verbose.

dinhxuanvu (approved):
/lgtm

PR #397, opened 2022-04-01
chore(api): modifies ImagesetConfigurationSpec to conform to Kubernet…

jpower432:
…es API Guidelines

Introduce Platform and PlatformType to replace OCP with
a more generic option. PlatformType is used to distinguish
between OKD and OCP. The original solution was to use the channel
name, but a specified type is most likely more survivable and allows
for later additions.

Replaces Repo type with Repositories in Helm Spec

Adds MarshalJSON and UnmarshalJSON methods to ImageType
to write the string representation to the metadata.json instead
of the int.

BREAKING CHANGE: Removes OCP and Repo types and invalidates past metadata

Signed-off-by: Jennifer Power <barnabei.jennifer@gmail.com>

# Description

Introduce Platform and PlatformType to replace OCP with
a more generic option. PlatformType is used to distinguish
between OKD and OCP. The original solution was to use the channel
name, but a specified type is most likely more survivable and allows
for later additions.

Replaces Repo type with Repositories in Helm Spec

Adds MarshalJSON and UnmarshalJSON methods to ImageType
to write the string representation to the metadata.json instead
of the int.

Removes redundant AdditionalImage and BlockedImage types

- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] Manual verification config below

```
kind: ImageSetConfiguration
storageConfig:
  local:
    path: metadata
mirror:
  platform:
    channels:
      - name: okd
        type: okd
      - name: stable-4.9
    graph: true
```

# Checklist:

- [X] I have made corresponding changes to the documentation


jpower432:
cc @jharmison-redhat

jpower432:
@soltysh This incorporates your feedback from #376. As far as removing the redundant structures (`AdditionalImages` and `BlockedImages`), I removed the former, but keep `SampleImages`. Then implementation of `Samples` is not planned for 4.11 so it may be wise to remove it from the API altogether. WDYT? cc @afflom

pkg/api/v1alpha2/types_config.go
```diff
@@ -26,26 +26,26 @@ type ImageSetConfigurationSpec struct {
...
+	Platform `json:"platform,omitempty"`
 	// Operators defines the configuration for Operator content types.
 	Operators []Operator `json:"operators,omitempty"`
 	// AdditionalImages defines the configuration for a list
 	// of individual image content types.
-	AdditionalImages []AdditionalImages `json:"additionalImages,omitempty"`
+	AdditionalImages []Image `json:"additionalImages,omitempty"`
 	// Helm define the configuration for Helm content types.
-	Helm Helm `json:"helm,omitempty"`
+	Helm `json:"helm,omitempty"`
```
soltysh, line 37: Any particular reason for this change?

soltysh (changes requested):
One question, rest lgtm

pkg/api/v1alpha2/types_config.go
jpower432: @soltysh Copy. I originally removed it because it was redundant, but I see that this diverged from the way the rest of the fields are specified. I put it back in place to keep the way the fields are specified through the API spec consistent.

PR #399, opened 2022-04-05
feat(mirror): image pruning from registry on mirroring workflows

### LINKED ISSUE #302: [RFE] Target registry image pruning
# Context
oc-mirror uses a metadata file to track the state of previous image downloads for differential operations and upgrade graph plotting. 
# Problem
## Metadata Size
Metadata created by oc-mirror will only increase in size. There is currently no way to reduce the size other than starting fresh. This leads to performance degradation of the application due to the amount of memory being used and the time needed to complete tasks related to metadata handling. Users have reported spikes in memory consumption and CPU usages during differential runs of oc-mirror.
## Registry Storage Size
A mechanism for oc-mirror is needed to prune old images out of a registry. It is preferably done through oc-mirror, but we must also tolerate organization cleaning requirements.

# Solution
The solution is that the `ImagesetConfiguration` `Mirror` spec will be a representation of what is the disconnected registry. When items are removed, oc-mirror will reconcile the current mirror and past mirror from the metadata to determine what can be pruned.

## Metadata
Only the most recent past run will need to be stored for reconciling with the new implementation. oc-mirror’s behavior will be based on the current and most recent metadata differentiation.
Use a flyweight pattern for describing the mirror type. We will need to benchmark to see how much we save on this vs. the mirror type in v1alpha1. We could, at the very least, do this for Blob types.

## Images
We can prune blobs from `PastBlobs` based on age to tolerate organization registry cleaning requirements. This should not be linked to PastMirror in any way. That is a read-only structure for a history only. A timestamp spec must be added to the Blob type.
- Bonus: We could allow users to turn off layer differentials by setting the temporal value to 0

A range for versions of Operator packages and OCP releases can be specified, and this will represent the contents of the upgrade graphs and what is targeted to get into the mirror.
Implementation - oc-mirror will differentiate between the current `Mirror` in the configuration and the `PastMirror` in the metadata. Any new version will be downloaded, and any previously downloaded version that is older than the minimum version will be pruned.

Cincinnati Implementation - We would build the graph for differential runs to get the upgrade path between the previous minimum and maximum versions and the currently requested versions.
Ensure the upgrade graph is maintained between channels (e.g., stable-4.8 to stable 4.10). On a best-effort basis, we will plot upgrades between the highest version in the lowest channel and the lowest version of the highest channel.
- Bonus: We could mirror all versions or just the shortest path with the graph boolean

### Example 
A user specifies 4.9.2-4.9.2 (this is the latest version at this time) and performs an initial mirror.
<time passes>

A user specifies 4.9.2-4.9.5 (this is the latest version at this time) and performs the first differential mirror. The Cincinnati graph is traversed, and an upgrade path is found 4.9.2->4.9.4->4.9.5, skipping 4.9.3. 4.9.4 and 4.9.5 are mirrored, with 4.9.2 known to already be mirrored.

<time passes>
A user specifies 4.9.2-4.9.10 (this is the latest version at this time) and performs a second differential mirror. The Cincinnati graph shows an upgrade path of 4.9.2->4.9.10 directly. Our last run metadata indicates that 4.9.2, 4.9.4, and 4.9.5 are present. We map an upgrade path from 4.9.5 to 4.9.10. 

- Caveat: There may be a shorter path between 4.9.2 - 4.9.10. Should we always plot the upgrade path between the minimum and maximum versions? If we don’t, use-cases like the one below would be a problem.

<time passes>
A user specifies 4.9.10-4.9.11 (this is the latest version at this time) and performs another differential mirror. The Cincinnati graph shows no path between 4.9.10 and 4.9.11 at this time, as the edge is blocked. We download 4.9.11 and log a warning. We now assume that they have no more clusters earlier than 4.9.10 in use and can effectively prune all metadata related to pre-4.9.10 mirroring and mark the images for deletion.

jpower432:
# Description

Adds image pruning based on metadata functionality to mirroring workflows 

Fixes #302
Blocked by #387 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Unit test added for association diff
- [X] Manually verified pruned images

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

Notes: 
Overall docs updates are coming on a separate PR

### DIFF
--- a/pkg/cli/mirror/list/releases.go
+++ b/pkg/cli/mirror/list/releases.go
@@ -78,6 +78,9 @@ func (o *ReleasesOptions) Run(ctx context.Context) error {
 	w := o.IOStreams.Out
 
 	client, err := cincinnati.NewOCPClient(uuid.New())
+	if err != nil {
+		return err
+	}
 
 	if o.Channels {
 		channels, err := cincinnati.GetChannels(ctx, client, o.Channel)
--- a/pkg/cli/mirror/mirror.go
+++ b/pkg/cli/mirror/mirror.go
@@ -223,7 +223,7 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error) 
 			return err
 		}
 
-		prevAssociations, err := o.removePreviouslyMirrored(mapping, meta)
+		prunedAssociations, err := o.removePreviouslyMirrored(mapping, meta)
 		if err != nil {
 			if errors.Is(err, ErrNoUpdatesExist) {
 				logrus.Infof("no new images detected, process stopping")
@@ -265,7 +265,7 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error) 
 		}
 
 		// Pack the images set
-		tmpBackend, err := o.Pack(cmd.Context(), prevAssociations, assocs, &meta, cfg.ArchiveSize)
+		tmpBackend, err := o.Pack(cmd.Context(), prunedAssociations, assocs, &meta, cfg.ArchiveSize)
 		if err != nil {
 			if errors.Is(err, ErrNoUpdatesExist) {
 				logrus.Infof("no updates detected, process stopping")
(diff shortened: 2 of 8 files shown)


afflom (approved):
I got a 405 from a registry:2 as expected. @dinhxuanvu over to you

dinhxuanvu:
> I got a 405 from a registry:2 as expected. @dinhxuanvu over to you

Is this expected or this is a bug? @jpower432

jpower432:
> > I got a 405 from a registry:2 as expected. @dinhxuanvu over to you
> 
> Is this expected or this is a bug? @jpower432

@dinhxuanvu Expected. Since some registries do not allow image deletion by default (e.g. registry:2) or it maybe be possible that some `oc-mirror` users do not have permissions to prune on their registry, `oc-mirror` will attempt to prune, but if a `transport` error is received, the pruning attempt will end and `oc-mirror` will log a warning with the error code and continue operations.

dinhxuanvu (approved):
/lgtm

PR #464, opened 2022-05-14
Bug 2088634: fix(operator): adds validation after rendering dc to check for invali…

### LINKED ISSUE #459: [BUG] the catalog is not validated after package filtering
# Version

```console
$ oc-mirror version
The main branch on commit 7bde4c7bde6b829a34592b2d12115fb38b15576f
```

# What happened?

In use cases when ranges are defined on operator packages, it is possible to specify the versions in a way that breaks the package validity by breaking the upgrade graph or removing the default channel.

# What did you expect to happen?

`oc-mirror` to report an error to signify that the catalog is not valid during the dry-run.

Mirror an operator package with a channel that is not the default channel.

N/A

# References

N/A

jpower432:
…d index

Signed-off-by: Jennifer Power <barnabei.jennifer@gmail.com>

# Description

Using ranges to specify operator bundle versions can result in pruned
default channels and invalid replace chains (semver ordering is not
enforced on operator upgrade graphs). This adds a step that converts
the declarative config to a model which performs validation
on the FBC data.

Fixes #459

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Unit tests added to check for errors for default channels and invalid graphs

# Checklist:

### DIFF
--- a/pkg/cli/mirror/operator.go
+++ b/pkg/cli/mirror/operator.go
@@ -29,12 +29,12 @@ import (
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/oc-mirror/pkg/api/v1alpha2"
 	"github.com/openshift/oc-mirror/pkg/config"
 	"github.com/openshift/oc-mirror/pkg/image"
 	"github.com/openshift/oc-mirror/pkg/operator"
-	"k8s.io/klog/v2"
 )
 
 // OperatorOptions configures either a Full or Diff mirror operation
@@ -120,7 +120,7 @@ func (o *OperatorOptions) run(ctx context.Context, cfg v1alpha2.ImageSetConfigur
 		// Render the catalog to mirror into a declarative config.
 		dc, ic, err := renderDC(ctx, reg, ctlg)
 		if err != nil {
-			return nil, err
+			return nil, checkErr(err)
 		}
 
 		mappings, err := o.plan(ctx, dc, ic, ctlgRef, targetCtlg)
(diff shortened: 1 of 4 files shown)


pkg/cli/mirror/operator.go
```diff
@@ -705,3 +717,25 @@ func validate(dc *declcfg.DeclarativeConfig) error {
...
 }
+
+// checkError will check validation errors
+// from operator registry and return a modified
+// error messages for mirror usage
+// FIXME(jpower432): Checking against the errors string could be an issue since
+// this is depending on the strings returned from `opm validate`. It would be better to propose that
+// the validation error are exposed here https://github.com/operator-framework/operator-registry/blob/master/alpha/model/error.go
+// and adds structure errors that return package and channels information.
+func checkErr(err error) error {
```
jpower432, line 728: @dinhxuanvu I am kind of going back and forth on whether adding this is worth it given it depends on string values from a library. An ideal situation would be if the `validationErr` type in operator registry was exposed and maybe defined as a structure error. Do you think that is possible in the future?

pkg/cli/mirror/operator.go
```diff
@@ -705,3 +717,29 @@ func validate(dc *declcfg.DeclarativeConfig) error {
...
+func (o *OperatorOptions) checkValidationErr(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	fmt.Fprintln(o.ErrOut, "\nThe rendered catalog is invalid.")
+	// handle known error causes
+	var validationMsg string
+	switch {
+	case strings.Contains(err.Error(), "channel must contain at least one bundle"):
```
dinhxuanvu, line 737: Hopefully one day we won't need to do this ugly string check for validation in FBC.

dinhxuanvu (commented):
/lgtm

Issue #240, opened 2021-12-14
Error info is ambiguous when versions is wrong in the ImageSetConfiguration


afflom:
Description of problem:
Error info is ambiguous when versions is wrong in the ImageSetConfiguration

Version-Release number of selected component (if applicable):

How reproducible:

Steps to Reproduce:
1. Use the imagesetconfig as follow :
`oc-mirror --config imageset.yaml file://testin --dry-run`
```
apiVersion: tmp-redhatgov.com/v1alpha1
kind: ImageSetConfiguration
mirror:
  ocp:
    channels:
      - name: stable-4.9
        versions:
          - 4.9
    graph: true
```

Actual result:
1. The error info is ambiguous when versions is wrong in the ImageSetConfiguration
[root@localhost 1206]# oc-mirror --config imageset.yaml file://testin --dry-run
error: decode tmp-redhatgov.com/v1alpha1, Kind=ImageSetConfiguration: json: cannot unmarshal number into Go struct field ReleaseChannel.mirror.ocp.channels.versions of type string

Expect result:
1. Show clear info to help user to correct the configure and at the some time oc-mirror should be able to parse either 4.9.0 or 4.9 since the cli accepts either way

https://bugzilla.redhat.com/show_bug.cgi?id=2029367

PR #494, opened 2022-06-27
Bug 2029367: fix(config): Parse X.Y format minVersion and maxVersion

d10n:
# Description

Error info is ambiguous when minVersion and maxVersion are wrong in the
ImageSetConfiguration, especially when using yaml.

The yaml unmarshaler parses unquoted X.Y values as a number, but users
intend that to be a semver version string.

The k8s yaml unmarshaler serializes to JSON and deserializes from JSON
in order to use the json marshaler's struct tags.

The json unmarshaler can't read a number into a string field, so a
custom UnmarshalJSON method is needed.

The custom json unmarshaling code would be simpler if MinVersion and
MaxVersion were extracted to a dedicated struct and inlined back into
ReleaseChannel, but there are a bunch of usages that would need to be
updated and I chose to prefer slightly-more-complicated unmarshaling
over clobbering history in a bunch of other places.

Bug 2029367
Fixes github issue #240

## Type of change

# How Has This Been Tested?

This is a client-side configuration parsing bug, so I wrote unit tests. Run `make test`

**Test Configuration**:
* Toolchain: go version go1.18.3 linux/amd64

# Checklist:

- [x] I have made corresponding changes to the documentation (I don't think any change is needed)

### DIFF
--- a/pkg/api/v1alpha2/types_config.go
+++ b/pkg/api/v1alpha2/types_config.go
@@ -107,9 +107,9 @@ func (r *ReleaseChannel) UnmarshalJSON(data []byte) error {
 			if err := json.Unmarshal(value, &minVersionString); err != nil {
 				minVersionString = string(value)
 			}
-			ver, err := semver.ParseTolerant(minVersionString)
+			ver, err := semver.Parse(minVersionString)
 			if err != nil {
-				return fmt.Errorf("unable to parse config; minVersion must be a version if present: %w", err)
+				return fmt.Errorf("unable to parse config; minVersion must be in valid semver format if present: %w", err)
 			}
 			r.MinVersion = ver.String()
 		case "maxVersion":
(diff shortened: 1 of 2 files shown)


jpower432:
@d10n I can see why you went this route given the accepted results in the linked issue, but since that was for a different API version, I am not sure it makes sense to accept the `X.Y` format. The major and minor versions are already inferred by the channel name. The minimum and maximum versions are calculated when no user input is given. I would personally prefer to just error out with a more detailed message when an invalid `semver` input is given just because it doesn't seem like the benefit is worth the complexity of accepting another input format. @d10n WDYT?

@afflom @dinhxuanvu Any thoughts?

d10n:
Thanks @jpower432 and @dinhxuanvu - I agree to prefer stricter input with better error messages. I've pushed up the change, which just happens to need the same structure of code to get the error message out.

sherine-k:
@deejross @jpower432 @d10n 
Should this PR be merged still? or is this just a stale PR?
thank you both for your help

sherine-k:
@d10n, @jpower432,
Do we close this PR? or is it still needed?
Thanks for your help

jpower432:
> @d10n, @jpower432, Do we close this PR? or is it still needed? Thanks for your help

@sherine-k If I recall correctly, the issue that was keeping it from being merged before was that it was not linked to a valid issue. However, I don't believe there is a valid bug open for this currently since the linked issue was closed due to staleness. It may need to be confirmed that the issue still exists. I will leave the final decision to @d10n.

Issue #453, opened 2022-05-12
The defaultChannel for an operator should be valid when filtering a package using min/maxVersion  


joehuizenga:
The current implementation of oc-mirror allows an ISC to specify a min/max version of a package

There are situations where specifying a range results in an invalid defaultChannel being specified in the “olm.package” object which makes the catalog invalid 

```
apiVersion: mirror.openshift.io/v1alpha2
kind: ImageSetConfiguration
storageConfig:
  local:
    path: /root/mirror
mirror:
  operators:
    - catalog: icr.io/cpopen/ibm-operator-catalog:fbc-latest
      full: false
      packages:
        - name: ibm-mq
          minVersion: '1.6.0'
          maxVersion: ‘1.7.0’
```

```
oc-mirror --config imageset-ibm-mq.yaml file://output --dry-run --skip-cleanup
```

produces an index.json that has the defaultChannel set to a channel
```
{
    "schema": "olm.package",
    "name": "ibm-mq",
    "defaultChannel": "v1.8",
…
}
```

This RFE is asking that oc-mirror enhance the operator min/maxVersion functionality to also determine/set the defaultChannel for a package that specifies a range.

Possible highlevel logic:

get the maxVersion for a package from the ISC (1.7.0)
find the olm.channel  that has an entry where package.maxVersion = olm.channel.entries[].name  //name includes this version (ibm-mq.v1.7.0)
  maxVersionChannelName = v1.7
get the current defaultChannel(v1.8) if != to one you found above (v1.7), replace it

{
    "schema": "olm.package",
    "name": "ibm-mq",
    "defaultChannel": "v1.7”,
     …
}

open questions:
what to do if there are multiple olm.channels that contain the same package/version? which one should be chosen as the default?
```
{
    "schema": "olm.channel",
    "name": "v1.7",
    "package": "ibm-mq",
    "entries": [
        {
            "name": "ibm-mq.v1.7.0",
            "skipRange": ">=1.0.0 <1.7.0"
        }
    ]
}

{
    "schema": "olm.channel",
    "name": “joes-channel”,
    "package": "ibm-mq",
    "entries": [
        {
            "name": "ibm-mq.v1.7.0",
            "skipRange": ">=1.0.0 <1.7.0"
        }
    ]
}
```

joehuizenga:
[...]

PR #497, opened 2022-07-20
Migration of the diff logic from operator-registry + setting of defaultChannel fix by adding channel's Priority property

pawicao-ibm:
# Description

This PR is a Work In Progress and serves the purpose of presenting the suggestion of changes towards the diff logic. Previously present in the opeator-registry project, the diff logic was suggested to be moved to oc-mirror together with the changes proposed by @joehuizenga - of setting a defaultChannel by adding a priority property to a channel.

Relevant discussions:
- https://github.com/openshift/oc-mirror/issues/453
- https://github.com/operator-framework/operator-registry/pull/969

Fixes [#453](https://github.com/openshift/oc-mirror/issues/453)

~~_This PR comes in parallel with changes to operator-registry. In order to test the migrated diff logic, operator-registry dependency must be set to this one: https://github.com/operator-framework/operator-registry/pull/1006 and vendor has to be updated_~~

_EDIT: following the suggestions https://github.com/operator-framework/operator-registry/pull/1006 the operator-registry PRs were separated into two. That means that in order to test oc-mirror, a dependency to operator-registry must be set to a version of the project which includes both of the changes. It can be for example this temporary branch: https://github.com/pawicao-ibm/operator-registry/tree/oc-mirror-testing_

## Type of change

- [ ] This change requires a documentation update (?)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:


pawicao-ibm:
As agreed, the diff logic should be moved to this repository and this PR presents a suggestion for such a migration.

Some questions that arose:

- the pkg structure - migrating the diff logic I suggested `action` and `diff` - is that okay?
- there are some changes that would have to made in the operator registry as well - since the diff parts of code are now here, some variables (schemas) are refernced from another project and therefore have to be public. Additionally, since the diff logic is getting out of operator-registry, does it also mean that `opm diff` command should be removed?
- To follow other parts of code I used `klog` for logging - check out the verbosity and if it matches how it's supposed to work
- **Sorting by name** is being done when there is no information about channel priority found. Right now we are sorting by comparing channel names as strings if the priority comparison fails. I am afraid that this way of comparison can not always produce the results we are expecting. For example if a channelName is v1.10 it will be sorted as smaller than v1.9 if we compare it with strings and not with any versioning package comparisons. This thought may be invalid though, cause I am not sure if the channels names can even go up to having two digits as the “minor” version. I’m not sure how the versioning works here. Can channel names also be something like stable and so on here? If so, that can also give us inaccurate results if we sort it like that. What do you think? The initial thought we had with @joehuizenga was to error it out in that case.

I'm looking forward for other comments as well. I'm new to both of these projects so there may be some concepts i haven't fully understood while preparing this PR and I'd appreciate feedback

@dinhxuanvu @joehuizenga @joelanford

dinhxuanvu:
I haven't had the time to review this PR in a depth but from the initial review, it looks promising. A quick question here is that do you have plan to hook this code up into oc-mirror code as well or this PR is strictly for the code migration?

jpower432:
> As agreed, the diff logic should be moved to this repository and this PR presents a suggestion for such a migration.
> 
> Some questions that arose:
> 
> * the pkg structure - migrating the diff logic I suggested `action` and `diff` - is that okay?

Overall this looks good to me. Just a thought on the package structure. I am not sure that `action` makes as much sense in the context of `oc-mirror` since diff is not a top-level `oc-mirror` action. To keep the migration simple and keep the boundary between the high-level and low-level diff logic that existed in `operator-registry,` it might make more sense if the `diff` package was moved under the operator package (this is where libraries related to low-level catalog processing are located for `oc-mirror`) and the files in the `action` package were moved directly under the operator package for (possible) later restructuring. WDYT @pawicao-ibm?

pawicao-ibm:
Thank you very much for your replies!

As for your refactoring suggestions @jpower432 it couldn't be done exactly like that since with that we were getting into circular imports between `pkg/v1alpha2` and `pkg/operator`, having the high-level diff logic directly in the `pkg/operator`. Therefore for now I added the high-level diff logic into `pkg/operator/diff` package and the diff includers and diff generators files into `pkg/operator/diff/include`. It certainly makes more sense than having a `pkg/action` package here. Nevertheless I am not sure if a name `include` is the right one here, I didn't really know how to name it, so further feedback would be really appreciated

jpower432:
> Thank you very much for your replies!
> 
> As for your refactoring suggestions @jpower432 it couldn't be done exactly like that since with that we were getting into circular imports between `pkg/v1alpha2` and `pkg/operator`, having the high-level diff logic directly in the `pkg/operator`. Therefore for now I added the high-level diff logic into `pkg/operator/diff` package and the diff includers and diff generators files into `pkg/operator/diff/include`. It certainly makes more sense than having a `pkg/action` package here. Nevertheless I am not sure if a name `include` is the right one here, I didn't really know how to name it, so further feedback would be really appreciated

@pawicao-ibm I think that name is probably fine for now IMO. I see where the circular import came from since this PR is adding the `DiffIncludeConfig` type that `oc-mirror` is doing conversions for in `v1alpha2`. I think what you ended up doing actually makes more sense.

jpower432:
@dinhxuanvu @deejross I just wanted to point out that `oc-mirror` does not use all of the functionality that the `diff` logic provides (specifically diffing two models), and it may have to be trimmed down to ensure there is no unused code. Not sure if the preference would be to have that trimmed down in this PR or a subsequent PR. Thoughts?

deejross:
Personally, I prefer to remove unneeded code when possible. My first reaction is to make a separate PR unless there are very minor changes required to this one in order to accomplish, but I'll leave that for you to decide.

dinhxuanvu:
> @dinhxuanvu @deejross I just wanted to point out that `oc-mirror` does not use all of the functionality that the `diff` logic provides (specifically diffing two models), and it may have to be trimmed down to ensure there is no unused code. Not sure if the preference would be to have that trimmed down in this PR or a subsequent PR. Thoughts?

My original proposal was to relocate the diff code and so revamp/refactor it as it is quite bloated and was written a bit hastily without a lot of thoughts for code sanity. As Jenn already said, there are a few of features that are not needed. I'm on board with asking for some code revamp as a part of this PR if possible. However, we can make this task as a progressive process as things can be changed over a series of PRs instead just one. Just my 2 cents.

jpower432:
Thanks for the feedback. I agree that capturing this requirement in a separate issue will do to limit the changes since I don't think the refactor will be particularly minor.

joehuizenga:
>there are some changes that would have to made in the operator registry as well - since the diff parts of code are now here, some variables (schemas) are refernced from another project and therefore have to be public. Additionally, since the diff logic is getting out of operator-registry, does it also mean that opm diff command should be removed?

@dinhxuanvu I know we had some discussion on `opm diff` remove or deprecate,  I thought the direction was to `deprecate` which would mean that operator-registry now includes code from oc-mirror until opm diff gets removed.  Is this the direction we want to go?  

>Sorting by name is being done when there is no information about channel priority found. Right now we are sorting by comparing channel names as strings if the priority comparison fails

I am ok erroring out here

>Thanks for the feedback. I agree that capturing this requirement in a separate issue will do to limit the changes since I don't think the refactor will be particularly minor.

agree, new issue to remove unused seems like a good path forward

>Therefore for now I added the high-level diff logic into pkg/operator/diff package and the diff includers and diff generators files into pkg/operator/diff/include

diff/include. first thought , it seems strange but struggling with a better name internal? common?   No strong opinion here

pawicao-ibm:
Following the suggestion from @joehuizenga , `diff/include` package was renamed to `diff/internal`.

Also following the suggestions https://github.com/operator-framework/operator-registry/pull/1006 the operator-registry PRs were separated into two. That means that in order to test oc-mirror, a dependency to operator-registry must be set to a version of the project which includes both of the changes. It can be for example this temporary branch: https://github.com/pawicao-ibm/operator-registry/tree/oc-mirror-testing

A really important matter here that remains is the way all of that will be merged - these changes seem to be dependent on each other so changes in operator-registry and here would have to be synced if it comes to merging.

The work here seems to be ready other than that, I will remove the WIP tag in that case

pawicao-ibm:
@jpower432 we are moving forward with both of the operator-registry PRs. One of them is merged already, the other should be merged soon.

Is this one here okay to be approved afterwards or there are any changes that you would like us to make?

dinhxuanvu:
Both PRs on registry have been merged. Feel free to pin to the latest commit on registry repo until a new version is released.

pawicao-ibm:
@jpower432 @deejross the operator-registry dependency is now set at the latest commit. Eveything in this PR should be now easily testable. Waiting for review/inputs from your side

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
 		var outputHasDefault bool
 		outputPkg.DefaultChannel, outputHasDefault = outputPkg.Channels[newPkg.DefaultChannel.Name]
 		if !outputHasDefault {
-			// Create a name-only channel since oldModel contains the channel already.
-			outputPkg.DefaultChannel = copyChannelNoBundles(newPkg.DefaultChannel, outputPkg)
+			// Set the defaultChannel using the priority of a channel when the default got filtered out
+			// If no channels with the Priority property, raise an error
+			if err := setDefaultChannel(outputPkg); err != nil {
+				return outputModel, err
```
jchunkins, line 195: you return the outputModel if there’s an error, unlike the rest of the function which returns nil upon error. Seems like this line should return nil for the model.

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
+			}
 		}
 	}
 
 	return outputModel, nil
 }
 
+type ChannelPriorityPropList []property.Channel
```
jchunkins, line 203: is this something you want to expose publicly? It only seems to be used in the private function `setDefaultChannel`

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
+		for chname, channel := range outputPkg.Channels {
+			if chname == chosenChannelName {
+				outputPkg.DefaultChannel = channel
+				klog.V(0).Infof("Newly assigned default channel from existing channels by Priority is %s\n", chosenChannelName)
+			}
+		}
+		return nil
+	}
+
+	return fmt.Errorf("unable to set a default channel for package: %s - no remaining channels in the filtered output have the Priority property", outputPkg.Name)
```
jchunkins, line 240: - variable `p` is too short… variables should be more descriptive
- variable `p` creates a resulting slice that is as long as your output channel, but it **_should not_** be. The size of the slice `p` is not determined solely by the number of channels. Each channel can have zero or more properties, and some properties may not be “olm.channel”. Lines 210 through 219 selectively add an item at the beginning of the slice, _potentially_ leaving “empty” channels at the end (e.g. if there were other properties that are not “olm.channel”, you could end up with some slice entries with `property.Channel {ChannelName: "", Priority: 0}`). On the other hand, if you have multiple “olm.channel” entires, you could assign an index at line 215 that exceeds the size of the slice. I assume your logging at lines 225 to 227 would potentially show an empty channel name with a zero priority for some of the items. 
  - I think at the very least you should start with an empty slice for variable `p` and only append values that match type “olm.channel” to avoid adding “invalid” channels in your slice. 
  - It may be useful to highlight in the log any channels that had a duplicate priority value since that's a sign that someone did something incorrectly.

jchunkins (changes requested):

pkg/operator/diff/internal/diff.go
jpower432: I would agree with you here, @jchunkins. Though the package itself can only be imported by the parent package due to its naming.

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
+type ChannelPriorityPropList []property.Channel
+
+// setDefaultChannel sets the new default channel of a package if the old default channel got filtered out.
+// Throws an error if there are no channels with the Priority property
+func setDefaultChannel(outputPkg *model.Package) error {
+	p := make(ChannelPriorityPropList, len(outputPkg.Channels))
+	i := 0
+	for _, channel := range outputPkg.Channels {
+		var channelPriority property.Channel
+		for _, prop := range channel.Properties {
```
jpower432, line 212: It may be better to import the constant from `operator-registry`. Just in case the value changes for this property type https://github.com/operator-framework/operator-registry/blob/ce3bb44f7618726474a78f95852e30a2bfb0346a/alpha/property/property.go#L141.

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
 	}
 
 	return outputModel, nil
 }
 
+type ChannelPriorityPropList []property.Channel
+
+// setDefaultChannel sets the new default channel of a package if the old default channel got filtered out.
+// Throws an error if there are no channels with the Priority property
+func setDefaultChannel(outputPkg *model.Package) error {
```
jpower432, line 207: I think this logic could use a unit test or be integrated into the existing unit tests in diff_test.go. It would help catch the errors in the cases @jchunkins is referring to.

jpower432 (changes requested):

pkg/operator/diff/internal/diff.go
```diff
@@ -186,14 +189,58 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
+	p := make(ChannelPriorityPropList, len(outputPkg.Channels))
+	i := 0
+	for _, channel := range outputPkg.Channels {
+		var channelPriority property.Channel
+		for _, prop := range channel.Properties {
+			if prop.Type == "olm.channel" {
```
jpower432, line 213: It may be better to import the constant from `operator-registry`. Just in case the value changes for this property type https://github.com/operator-framework/operator-registry/blob/ce3bb44f7618726474a78f95852e30a2bfb0346a/alpha/property/property.go#L141.

pawicao-ibm:
@jchunkins @jpower432 thank you very much for your feedback. It has been addressed, please check

pkg/operator/diff/internal/diff.go
```diff
@@ -183,17 +187,74 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
 	// Default channel may not have been copied, so set it to the new default channel here.
 	for _, outputPkg := range outputModel {
 		newPkg := newModel[outputPkg.Name]
+		if newPkg == nil {
```
jpower432, line 190: This is a just a nit, you could do the following here and on the previous line:
```
newPkg, found := newModel[outputPkg.Name]
if !found {
return nil, fmt.Errorf("package %s not present in the diff new model", outputPkg.Name)
}

jpower432 (commented):
Just a nit, but overall looks good to me. Leaving final approval and merge to @deejross.

pawicao-ibm:
@jpower432 I think that when I was re-requesting review earlier on, somehow @deejross was removed from the reviewers list and I'm not able to alter that

In any case, final feedback addressed, ready for final approval

dinhxuanvu:
@deejross @jpower432 The PR looks good for me. I would like to ask if you two can do a round of review and give it a lgtm. Then, I can approve it. Thanks.

jchunkins:
@deejross @jpower432 @dinhxuanvu 

Two minor changes were necessary:
- update the operator-registry reference to an actual release (used v1.26.1)
- improve the error message and logging when we can't actually set the default channel.

After these changes things /lgtm

jchunkins (approved):

jpower432 (approved):
@dinhxuanvu These changes LGTM also.

dinhxuanvu (approved):
/approve

Destorntw:
Hi team,
Hillay from the IGC PS,
In the past I had no problems mirroring a specific operator version from a channel that was not the daft channel. 
Since using the oc-mirror v4.11 binary it would auto-re-direct me to the dafulat channel and won't mirror (output errors) due to me asking for a package version that is not in the "default channel" set up by the catalog and operator maintainers as I understood.

With the advice of a developer from the team (former role) I tried using the oc-mirror v4.12 binary and still had errors (similar but different). 
How can I best work to advance this feature/bug in the cli?
Attaching my errors and the same imageSetConfig.yaml used in both tries in different oc-mirror binaries.

imageSet:
apiVersion: [mirror.openshift.io/v1alpha2](http://mirror.openshift.io/v1alpha2)
kind: ImageSetConfiguration
archiveSize: 50
mirror:
  operators:
  - catalog: [registry.redhat.io/redhat/redhat-operator-index:v4.10](http://registry.redhat.io/redhat/redhat-operator-index:v4.10)
    packages:
    - name: advanced-cluster-management
      channels:
      - name: release-2.4
        minVersion: '2.4.9'
        maxVersion: '2.4.9'

image

image

PR #503, opened 2022-08-30
WRKLDS-456 - Wrapper function for ParseReference

sherine-k:
# Description
This is the first part of the implementation described in [the design document](https://docs.google.com/document/d/1-TESqErOjxxWVPCbhQUfnT3XezG2898fEREuhGena5Q/edit#heading=h.mqgxt28qr45k)
It creates ParseReference function in the image package, that wraps around imagesource.ParseReference, adding support for `oci:` prefix
This new function is used instead of imagesource.ParseReference in pkg/cli/mirror/operator.go

Fixes # [WRKLDS-456](https://issues.redhat.com//browse/WRKLDS-456) - Partial

## Type of change

# How Has This Been Tested?

- pkg/image/image_test.go

# Checklist:

- [NA] I have made corresponding changes to the documentation

### DIFF
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -1,10 +1,18 @@
 package image
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
-	"strings"
+	libgoref "github.com/openshift/library-go/pkg/image/reference"
+	"github.com/openshift/oc/pkg/cli/image/imagesource"
+)
+
+var (
+	DestinationOCI imagesource.DestinationType = "oci"
 )
 
 // GetVersionsFromImage gets the set of versions after stripping a dash-suffix,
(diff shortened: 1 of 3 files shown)


lmzuccarelli (commented):
/lgtm - I will refer the final lgtm and approval to the WRKLDS team

sherine-k:
> /lgtm - I will refer the final lgtm and approval to the WRKLDS team. @sherine-k - could you check why the prow jobs were failing

Yes, I just removed the Draft. Adding it stopped all jobs with a sigterm

jpower432 (commented):
Looks good from my perspective as well. I will defer final approval to WRKLDS team.

pkg/image/image.go
```diff
@@ -31,3 +39,26 @@ func GetTagsFromImage(image string) ([]string, error) {
 	tags, err := remote.List(repo, remote.WithAuthFromKeychain(authn.DefaultKeychain))
 	return tags, err
 }
+
+/* ParseReference is a wrapper function of imagesource.ParseReference
+   It provides support for oci: prefixes
+*/
+func ParseReference(ref string) (imagesource.TypedImageReference, error) {
+	dstType := imagesource.DestinationRegistry
+	switch {
```
ardaguclu, line 48: We are mirroring same conditionals in here https://github.com/openshift/oc/blob/b6f6c9ed655b1c60b636d524372380d55b681af0/pkg/cli/image/imagesource/reference.go#L98. When new case needs to be added, this function should be maintained as well. 

Instead of this, maybe we can pass all other conditions whose are not including `oci:` prefix to the main function. Does that simplify this function?. I mean something like this;

```go
if !strings.HasPrefix(ref, "oci:") {
return imagesource.ParseReference(ref)
}

dstType = DestinationOCI
ref = strings.TrimPrefix(ref, "oci:")
ref = strings.TrimPrefix(ref, "//") //it could be that there is none
ref = strings.TrimPrefix(ref, "/")  // case of full path

dst, err := libgoref.Parse(ref)
if err != nil {
       return imagesource.TypedImageReference{Ref: dst, Type: dstType}, fmt.Errorf("%q is not a valid image reference: %v", ref, err)
}
return imagesource.TypedImageReference{Ref: dst, Type: dstType}, nil
```

sherine-k: Good idea. thnaks @ardaguclu . 
Commit updated.

sherine-k:
Thanks @ardaguclu for the review! :) 
@lmzuccarelli @ardaguclu can one of you add a approved label?
thanks

PR #512, opened 2022-09-19
CFE-538 - WRKLDS-456 - WRKLDS-457: copy images from and to remote registry using local directory (oci format)

lmzuccarelli:
# Description

**Overview**

**Copy**

- Read imagesetconfig file (command line)
- Pull catalog image
- Find FB config layer and untar
- Read filtered catalog.json file to get related images
- Pull filtered operators (fist round, phase 2 to consider all operators i,e no filters)
- Store in oci format in designated local folder

**Mirror**

- Read imagsetconfig file (command line)
- List all relevant operators (filter) on local directory
- Read catalog.json to get reletaed images
- Configure each image to reflect remote url 
- Mirror images from local directory to remote in oci format

JIRA Task - [WORKLOADS-536](https://issues.redhat.com/browse/WRKLDS-536)

**Verification Steps**

Copy workflow
```
$ bin/oc-mirror -c iscfg.yaml file://noo --use-oci-feature --oci-feature-action=copy
2022/09/19 18:53:34 INFO: downloading the catalog image
Copying blob bb05a60d034d done  
Copying blob 6521843dd476 skipped: already exists  
Copying blob 1a6ae3d35ced skipped: already exists  
Copying blob 0ee984cde3db skipped: already exists  
Copying blob 57fbfcc98e2f skipped: already exists  
Copying blob a3c83f4d53b1 skipped: already exists  
Copying config f6d2ed3a9a done  
Writing manifest to image destination
Storing signatures
2022/09/19 18:54:53 INFO: finding file based config (in catalog layers)
noo/blobs/sha256/1a6ae3d35ced1c7654b3bf1a66b8a513d2ee7f497728e0c5c74841807c4b8e77
noo/blobs/sha256/6521843dd4764da8f455aa0734384e4e056cab523fba813770cda18edf0e32c3
noo/blobs/sha256/0ee984cde3db6385b6debbce97f7282b14e829b0b9a7655093010ae433428726
noo/blobs/sha256/57fbfcc98e2f07b2bce9317f1eabab1790187b695932f41b46ad2a6398349748
noo/blobs/sha256/a3c83f4d53b1b0fb6958063db7ff4ba2bb167d9a40f36b4d6d901454e233304d
noo/blobs/sha256/a1aeeeb6653ab711c46eac88a908eb756e3caaf3c12633d300b4dc08def56a42
node-observability-operator true
Copying blob 3735f4762e2e skipped: already exists  
Copying blob 1e09a5ee0038 skipped: already exists  
[...]

PR #517, opened 2022-10-07
Pr 512 follow

sherine-k:
This PR is simply to fix the final comments left by @soltysh on https://github.com/openshift/oc-mirror/pull/512

### DIFF
--- a/pkg/cli/mirror/mirror.go
+++ b/pkg/cli/mirror/mirror.go
@@ -1,6 +1,7 @@
 package mirror
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log"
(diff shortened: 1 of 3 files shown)


pkg/cli/mirror/mirror.go
```diff
@@ -290,284 +290,285 @@ func (o *MirrorOptions) Run(cmd *cobra.Command, f kcmdutil.Factory) (err error)
 	diskToMirror := len(o.ToMirror) > 0 && len(o.From) > 0
 	mirrorToMirror := len(o.ToMirror) > 0 && len(o.ConfigPath) > 0
 
-	if o.UseOCIFeature {
```
soltysh, line 293: o.EI don't think this improves the situation much, I'd suggest something like:
```golang
if o.UseOCIFeature {
    return mirrorOCIImage()
}
return mirrorImage()
```
then copy the contents of those if blocks under those functions and it'll be so much easier to read.

soltysh (changes requested):

pkg/cli/mirror/mirror.go
sherine-k: OK this explains it better. thanks

sherine-k: Change done. Thank you @soltysh

PR #533, opened 2022-12-07
OCPBUGS-3414: fix: Missing 'ImageContentSourcePolicy' and 'CatalogSou…

sherine-k:
…rce' in the oci fbc feature implementation

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/pkg/api/v1alpha2/types_config.go
+++ b/pkg/api/v1alpha2/types_config.go
@@ -114,6 +114,9 @@ type Operator struct {
 	// SkipDependencies will not include dependencies
 	// of bundles included in the diff if true.
 	SkipDependencies bool `json:"skipDependencies,omitempty"`
+	// OriginalRepo is used when the Catalog is an OCI FBC (File Based Catalog) location.
+	// It contains the reference to the original repo on a remote registry
+	OriginalRepo string `json:"originalRepo,omitempty"`
 }
 
 // GetUniqueName determines the catalog name that will
--- a/pkg/api/v1alpha2/types_config.go
+++ b/pkg/api/v1alpha2/types_config.go
@@ -114,9 +114,9 @@ type Operator struct {
 	// SkipDependencies will not include dependencies
 	// of bundles included in the diff if true.
 	SkipDependencies bool `json:"skipDependencies,omitempty"`
-	// OriginalRepo is used when the Catalog is an OCI FBC (File Based Catalog) location.
+	// OriginalRef is used when the Catalog is an OCI FBC (File Based Catalog) location.
 	// It contains the reference to the original repo on a remote registry
-	OriginalRepo string `json:"originalRepo,omitempty"`
+	OriginalRef string `json:"originalRef,omitempty"`
 }
 
 // GetUniqueName determines the catalog name that will
--- a/pkg/cli/mirror/copy.go
+++ b/pkg/cli/mirror/copy.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -184,7 +185,8 @@ func (o *MirrorOptions) bulkImageMirror(ctx context.Context, isc *v1alpha2.Image
 			return err
 		}
 
-		// place related images into the workspace
+		// place related images into the workspace - aka mirrorToDisk
+		// TODO this should probably be done only if artifacts have not been copied
 		result, err := o.generateSrcToFileMapping(ctx, relatedImages)
 		if err != nil {
 			return err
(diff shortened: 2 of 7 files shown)


sherine-k:
@jchunkins @lmzuccarelli 
This is the proposal for generating the catalogSource.

This is still a WIP : unit tests, and other tests are still to be done.

For now the only test that was done was :
**ImageSetConfig**
```
kind: ImageSetConfiguration
mirror:
  operators:
  #rhop
  - catalog: oci:///home/skhoury/go/src/github.com/openshift/oc-mirror/my-oci-catalog/redhat-operator-index
    originalRepo: registry.redhat.io/redhat/redhat-operator-index:v4.12
    packages:
    - name: node-observability-operator
```
**Command**
```
./bin/oc-mirror --config /home/skhoury/go/src/github.com/openshift/oc-mirror/imstcfg-3414.yaml --use-oci-feature --oci-feature-action=mirror docker://localhost:5000 --dest-skip-tls
```
**Result ImageContentSourcePolicy.yaml**
```
---
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  labels:
    operators.openshift.org/catalog: "true"
  name: operator-0
spec:
  repositoryDigestMirrors:
  - mirrors:
    - localhost:5000
    source: registry.redhat.io/redhat
```
**catalogSource-redhat-operator-index.yaml**
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: redhat-operator-index
  namespace: openshift-marketplace
spec:
  image: localhost:5000/redhat-operator-index:v4.12
  sourceType: grpc
```

I'm no expert on the CatalogSource and ImageContentSourcePolicy, so please let me know if these would work on a cluster.

Thank you

sherine-k:
/assign @deejross @zhouying7780 
Hi Ross, Ying, I have a doubt about the contents of ImageContentSourcePolicy and CatalogSource. 
Would you be able to confirm if this is what is expected?
/cc @jchunkins @lmzuccarelli

jchunkins (changes requested):
Here's my findings from debugging the code in this PR:

# Catalog Source

I encountered some unexpected docker reference parsing. Given this example:

foo.com/cp/test/common-services@sha256:ef64abd2c4c9acdc433ed4454b008d90891fe18fe33d3a53e7d6104a4a8bf5c5

I get this result after parsing:

```
github.com/openshift/library-go/pkg/image/reference.DockerImageReference {
    Registry: "foo.com", 
    Namespace: "cp", 
    Name: "test/common-services", 
    Tag: "", 
    ID: "sha256:ef64abd2c4c9acdc433ed4454b008d90891fe18fe33d3a53e7d6104a4a8bf5c5"
}
```

So the parser handles arbitrary depth for namespaces differently from the way I expected it to. According to [these tests](https://github.com/openshift/library-go/blob/master/pkg/image/reference/reference_test.go#L51-L55) this is the expected behavior for this API, so it is what it is. Just for reference see the [distribution BNF naming convention](https://github.com/distribution/distribution/blob/main/reference/reference.go#L6-L26) for how they describe a docker reference. They call each slash separated portion of the docker reference a "path-component". I expected the namespace to be `cp/test` and the image name to be last "path-component" `common-services`, but alas that's not what it does. This has an impact on the final naming of the generated catalog source due to the slash. The current error I am seeing is:

"error writing CatalogSource: open oc-mirror-workspace/results-1670710694/catalogSource-test/common-services.yaml: no such file or directory"

This happens because at [this line](https://github.com/sherine-k/oc-mirror/blob/OCPBUGS-3414/pkg/cli/mirror/manifests.go#L300) the "name" is `test/common-services`. When it gets to [this line](https://github.com/sherine-k/oc-mirror/blob/OCPBUGS-3414/pkg/cli/mirror/manifests.go#L315) the path it wants to write to is "oc-mirror-workspace/results-1670715405/catalogSource-test/common-services.yaml", but because of the extra slash, we get a directory that does not actually exist.

So my solution to this particular problem was to add the following just after [this line](https://github.com/sherine-k/oc-mirror/blob/OCPBUGS-3414/pkg/cli/mirror/manifests.go#L300):

```go
name = strings.ReplaceAll(name, "/", "-")
```

At least this way the slash char is made "safe" for file path use.

As far as the content of the catalog source this looks "correct". 

```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: test/common-services
  namespace: openshift-marketplace
spec:
  image: localhost:5000/tmp/wrklds-536/common-services@sha256:ef64abd2c4c9acdc433ed4454b008d90891fe18fe33d3a53e7d6104a4a8bf5c5
  sourceType: grpc
```

# ICSP

However, the ICSP does not look complete. Unfortunately the generated ICSP is only taking into account the catalog source image. You can see this happening at [this line](https://github.com/sherine-k/oc-mirror/blob/OCPBUGS-3414/pkg/cli/mirror/mirror.go#L711). It makes no attempt to handle the "related images", so you get the following output:

```
---
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  labels:
    operators.openshift.org/catalog: "true"
  name: operator-0
spec:
  repositoryDigestMirrors:
  - mirrors:
    - localhost:5000/tmp
    source: hyc-cp-opencontent-docker-local.artifactory.swg-devops.com/cp
```

This seems to be a "top level" namespace configuration... what I mean by this is the first "path-component" (using docker BNF terminology here) is provided, and no deeper. We follow this style of ICSP in our own instructions for customers:

```
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  name: mirror-config
spec:
  repositoryDigestMirrors:
    - mirrors:
        - cp.stg.icr.io/cp
      source: cp.icr.io/cp
    - mirrors:
        - cp.stg.icr.io/cp
      source: icr.io/cpopen
```

This way we don't need to get specific about every single image reference and can just say (in english) `When you encounter and ref that starts with "cp.icr.io/cp" try to find it at "cp.stg.icr.io/cp" instead, and if that fails fall back to "cp.icr.io/cp"`. If you don't limit it this way, the ICSP can get very long if there's lots of images.

At any rate, if we can get the related images to be included in the ICSP too, then I think this would solve our problem. Honestly I don't get how the regular pathways through oc mirror don't run into this issue too.

jchunkins:
@jpower432 Given what I mentioned in my review above, how are related images handled in regards to ICSP? Just curious on your perspective on what that was not needed.

jpower432:
@jchunkins They are needed. Related images are packed and carried over just like bundle images. There are currently labels as bundle images at this [line ](https://github.com/openshift/oc-mirror/blob/b2b9a28c85efc9da57de00095797ee8f39c5cc24/pkg/cli/mirror/operator.go#L413) and not differentiated during the mapping generation.

So the ICSPs are generated, but the way they are labeled in the code is misleading. This could probably use a comment at the very least since so far we have not found that bundle images need to be handled any differently than related images during the mirroring process.

sherine-k:
@jpower432 
>  Related images are packed and carried over just like bundle images. There are currently labeled as bundle images at this [line ](https://github.com/openshift/oc-mirror/blob/b2b9a28c85efc9da57de00095797ee8f39c5cc24/pkg/cli/mirror/operator.go#L413) 
Do you mean we should use `TypeOperatorBundle` instead of `TypeOperatorRelatedImage` for all related images (regardless of bundle or not) at this [line](https://github.com/openshift/oc-mirror/blob/b2b9a28c85efc9da57de00095797ee8f39c5cc24/pkg/cli/mirror/copy.go#L351)?

soltysh (approved):
/lgtm
/approve

jchunkins (commented):
This already got merged (not sure why), so I am going to send what I had in my pending review. Guess I  will follow up in slack

PR #552, opened 2023-01-24
[graph-data] use new cincinnati endpoint to download tarball

PratikMahajan:
# Description

move to the new endpoin hosted on redhat domain api.openshift.com 
to download the tarball for openshift update service (cincinnati) graph-data.

Also changes the existing logic and extracts the tarball before creating the 
container image.

As this is an init container or will be used with cincinnati plugin, we're using 
a stripped down ubi image (ubi-micro)

Fixes [OTA-831](https://issues.redhat.com//browse/OTA-831)

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/pkg/cli/mirror/cincinnati_graph_image.go
+++ b/pkg/cli/mirror/cincinnati_graph_image.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	archiver "github.com/mholt/archiver/v3"
 	"github.com/openshift/library-go/pkg/image/reference"
 	"github.com/openshift/oc/pkg/cli/image/imagesource"
 	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
@@ -28,10 +29,11 @@ import (
 
 const (
 	// Base image to use when build graph image
-	graphBaseImage = "registry.access.redhat.com/ubi8/ubi:latest"
+	graphBaseImage = "registry.access.redhat.com/ubi8/ubi-micro:latest"
 	// URL where graph archive is stored
 	graphURL       = "https://api.openshift.com/api/upgrades_info/graph-data"
 	outputFile     = "cincinnati-graph-data.tar.gz"
+	graphDataDir   = "/var/lib/cincinnati/graph-data/"
 	getDataTimeout = time.Minute * 60
 )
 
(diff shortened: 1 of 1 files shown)


pkg/cli/mirror/cincinnati_graph_image.go
```diff
@@ -167,3 +177,55 @@ func downloadGraphData(ctx context.Context, dir string) error {
...
+			}
+		case tar.TypeReg:
+			outFile, err := os.Create(fPath)
+			if err != nil {
+				return fmt.Errorf("failed to create file while extracting tar: %v", err)
+			}
+			if _, err := io.Copy(outFile, tarReader); err != nil {
+				return fmt.Errorf("failed to copy extracted file: %v", err)
+			}
+			outFile.Close()
```
petr-muller, line 220: technically this is a leak on errors, but I don't think it matters that much in `oc`

pkg/cli/mirror/cincinnati_graph_image.go
```diff
@@ -1,6 +1,8 @@
 package mirror
 
 import (
+	"archive/tar"
```
petr-muller, line 4: Seems like we vendor (and even [directly use](https://github.com/openshift/oc-mirror/blob/main/go.mod#L17)) https://github.com/mholt/archiver, maybe we just want to use that and not write our own `untar`?

pkg/cli/mirror/cincinnati_graph_image.go
PratikMahajan: using archiver library instead of this function

pkg/cli/mirror/cincinnati_graph_image.go
PratikMahajan: moved to mholt/archiver

pkg/cli/mirror/cincinnati_graph_image.go
```diff
@@ -87,13 +89,20 @@ func (o *MirrorOptions) buildGraphImage(ctx context.Context, dstDir string) (ima
 
 	// unpack graph data archive and build image
 	graphToFile := filepath.Join(dstDir, config.GraphDataDir, outputFile)
-	add, err := builder.LayerFromPath(".", graphToFile)
+	graphDataFolder := filepath.Join(dstDir, config.GraphDataDir, "/graph-data")
+
+	err = archiver.Unarchive(graphToFile, graphDataFolder)
```
LalatenduMohanty, line 94: Before this change oc-mirror also handled tar.gz file, I am not sure why we need to unarchive the new tar.gz file?

PratikMahajan: the old logic extracted the tar file when the container was started using Entrypoint. 
In cases when the entrypoint is not used, we wont be able to extract the contents of tar file. (for example using dkrv2 plugin of cincinnati). 
This change makes the container image compatible with both cases, i.e using as `init` container and using with dkrv2 plugin.

LalatenduMohanty (approved):
/lgtm

zhouying7780:
Pre-merge test done , no issue found

dinhxuanvu (approved):
/approve

PR #568, opened 2023-02-28
WIP First pass for multi architecture catalogs

jchunkins:
# Description

- Numerous changes to read/write single and multi arch images. Each code path has to handle the parent/child hierarchy of indexes, usually via recursion.
- In operator.go
  - Made the operator rendering code "group together" the DeclarativeConfig and IncludeConfig and keep track of them on a per platform basis. Introduced this and other new data structures to pass between the various functions in `pkg/cli/mirror`  as well as passing them to other packages like `pkg/metadata`. Introduced `pkg/cli/mirror/operatorcatalog` to contain structs which are used in the aforementioned packages to avoid circular dependencies.
  - Make writeLayout() attempt to get platform information from image configuration and fall back to linux/amd64 if it can't
  - Make writeConfigs() write content in new single / multi arch aware directory structure
- In catalog_images.go
  - Make the rebuildCatalogs() function handle single / multi arch catalogs
- In fbc_operators.go
  - Make extractDeclarativeConfigFromImage() function (which replaced other existing functions using the go-containerregistry API instead.
  - Make getManifest() function multi arch aware
- In util.go
  - Add getCatalogMetadataByPlatform() function to capture metadata about platform specific catalogs and returns it to the caller. This information is now passed around various locations in the code.
  - Add getDigestFromOCILayout() function to get the hash associated with a given platform within an OCI layout
- metadata.go
  - Add ASCII art to describe the new layout structure for the workspace source directory
- pkg/image/builder
  - Make the image builder single / multi arch aware
  - Allow the API to provide an optional (i.e. can be nil) matcher. A matcher is satisfies a given criteria (e.g. a specific platform) to allow the builder to only update images that match the provided matcher criteria.
- update bunch of test cases and introduce new tests
- added docs as to clarify pieces as I understood them, and some of these changes are unrelated to the core purpose of this PR
- updated vendoring with the `make sanity` target

Fixes # https://issues.redhat.com/browse/CFE-760

## Type of change

# How Has This Been Tested?

- [ ] Test With Docker (mirror to mirror)
  - Use a ImageSetConfiguration which points to a "fat manifest" catalog:
  ```yaml
  kind: ImageSetConfiguration
  storageConfig:
    local:
      path: /tmp/localstorage
  mirror:
    operators:
    - catalog: icr.io/cpopen/ibm-zcon-zosconnect-catalog@sha256:6f02ecef46020bcd21bdd24a01f435023d5fc3943972ef0d9769d5276e178e76
  ```
  - Start a local docker registry to act as the destination
  ```
  docker run -d -p 5000:5000 --name registry registry:2
  ```
  - Run oc mirror like this (no errors should occur):
  ```
  oc mirror -c "/path/to/ImageSetConfiguration.yml" --dest-skip-tls --dest-use-http docker://localhost:5000
  ```
  - Throw away your local docker registry
  ```
  docker stop registry
  docker rm registry
  ```
- [ ] Test With Docker (mirror to disk)
  - Use a ImageSetConfiguration which points to a "fat manifest" catalog:
  ```yaml
  kind: ImageSetConfiguration
  storageConfig:
    local:
      path: /tmp/localstorage
  mirror:
    operators:
    - catalog: icr.io/cpopen/ibm-zcon-zosconnect-catalog@sha256:6f02ecef46020bcd21bdd24a01f435023d5fc3943972ef0d9769d5276e178e76
  ```
  - Run oc mirror like this (no errors should occur):
  ```
  oc mirror -c "/path/to/ImageSetConfiguration.yml" file://tmp/mirrortodisktest
  ```
- [ ] Test With OCI
  - Pull a docker "fat manifest" image into a location on the file system:
  ```
  skopeo copy --all --format v2s2 docker://icr.io/cpopen/ibm-zcon-zosconnect-catalog@sha256:6f02ecef46020bcd21bdd24a01f435023d5fc3943972ef0d9769d5276e178e76 oci:///Users/jhunkins/temp/ocmirror
  ```
  - Use a ImageSetConfiguration which points to this newly downloaded image like this:
  ```yaml
  kind: ImageSetConfiguration
  storageConfig:
    local:
      path: /tmp/localstorage
  mirror:
    operators:
    - catalog: oci:///Users/jhunkins/temp/ocmirror
  ```
  - Start a local docker registry to act as the destination
  ```
  docker run -d -p 5000:5000 --name registry registry:2
  ```
  - Run oc mirror like this (no errors should occur):
  ```
  oc mirror -c "/path/to/ImageSetConfiguration.yml" --use-oci-feature --oci-insecure-signature-policy --dest-skip-tls --dest-use-http docker://localhost:5000
  ```
  - Throw away your local docker registry
  ```
  docker stop registry
  docker rm registry
  ```

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:


pkg/cli/mirror/catalog_images.go
```diff
@@ -40,6 +41,56 @@ func (o *MirrorOptions) unpackCatalog(dstDir string, filesInArchive map[string]s
...
+
+• dstDir: the path to where the config.SourceDir resides
+
+# Returns
+
+• image.TypedImageMapping: the source/destination mapping for the catalog (this contains all images for single or multiple architectures)
+
+• error: non-nil if error occurs, nil otherwise
+*/
 func (o *MirrorOptions) rebuildCatalogs(ctx context.Context, dstDir string) (image.TypedImageMapping, error) {
```
lmzuccarelli, line 94: Is it possible to add some unit tests , the cognitive complexity is high so we need to grasp this better

pkg/cli/mirror/fbc_operators_test.go
```diff
@@ -140,7 +140,8 @@ func TestGetManifest(t *testing.T) {
 				require.EqualError(t, err, c.err)
```
lmzuccarelli, line 140: nit: unit test fails - change unmarshall -> unmarshal

pkg/config/load_test.go
```diff
@@ -184,27 +185,58 @@ mirror:
...
+			expected: "registry.com/ns/foo:v1.3",
+		},
+		{
+			expected: "registry.com/ns/baz:v1.3",
+		},
+		{
+			expected: "registry.com/ns/baz:v1.2",
+		},
+		{
+			expected: "tmp",
```
lmzuccarelli, line 217: These tests fail - could you update to expected value i.e for tmp it should be /tmp etc

pkg/cli/mirror/catalog_images.go
```diff
@@ -88,29 +214,55 @@ func (o *MirrorOptions) rebuildCatalogs(ctx context.Context, dstDir string) (ima
...
+			platform := *NewOperatorCatalogPlatform(platformString)
+			hash, err := getDigestFromOCILayout(ctx, layout.Path(filepath.Join(fullPathToRepoDir, config.LayoutsDir)), platform)
+			if err != nil {
+				return err
+			}
+			info.fullPathToRepoDir = fullPathToRepoDir
+			info.imageDetails = append(info.imageDetails, imageDetails{
+				indexJsonPath: fpath,
+				platform:      platform,
+				hash:          *hash,
```
lmzuccarelli, line 246: This line is failing in the e2e tests

```goroutine 1 [running]:
github.com/openshift/oc-mirror/pkg/cli/mirror.(*MirrorOptions).rebuildCatalogs.func2({0xc0003e1b00, 0x86}, {0x45656e0, 0xc000cc41a0?}, {0x0?, 0x0})
	/home/lzuccarelli/go/src/github.com/openshift/oc-mirror/pkg/cli/mirror/catalog_images.go:246 +0xb38
path/filepath.walk({0xc0003e1b00, 0x86}, {0x45656e0, 0xc000cc41a0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:433 +0x123
path/filepath.walk({0xc00044be80, 0x7b}, {0x45656e0, 0xc000cc40d0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc00044bd00, 0x75}, {0x45656e0, 0xc00099d2b0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc0005bd180, 0x61}, {0x45656e0, 0xc00099d1e0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc000c8a060, 0x53}, {0x45656e0, 0xc00099d110}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc0005b2460, 0x49}, {0x45656e0, 0xc00099d040}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc000db64b0, 0x2e}, {0x45656e0, 0xc00099cea0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.walk({0xc0009a02a0, 0x25}, {0x45656e0, 0xc00099cdd0}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:457 +0x285
path/filepath.Walk({0xc0009a02a0, 0x25}, 0xc00114e3f8)
	/usr/local/go/src/path/filepath/path.go:520 +0x6c
github.com/openshift/oc-mirror/pkg/cli/mirror.(*MirrorOptions).rebuildCatalogs(0xc000288b60, {0x455d250, 0xc0000560e0}, {0xc0009a02a0, 0x25})
	/home/lzuccarelli/go/src/github.com/openshift/oc-mirror/pkg/cli/mirror/catalog_images.go:131 +0x2b8```

File: pkg/cli/mirror/catalog_images.go [hunk 5]
```diff
@@ -88,29 +214,55 @@ func (o *MirrorOptions) rebuildCatalogs(ctx context.Context, dstDir string) (ima
 			if err != nil {
 				return fmt.Errorf("error parsing index dir path %q as image %q: %v", fpath, img, err)
 			}
-			ctlgRef.Ref = sourceRef.Ref
+
+			destRef := image.TypedImage{}
+			destRef.Type = imagesource.DestinationRegistry
```
sherine-k, line 219: Are we sure this will work also for MirrorToDisk & DiskToMirror workflows?

pkg/config/load_test.go
jchunkins: I've updated the tests and added some more scenarios. I've not changed `GetUniqueName` so the results just simply match the value that came out. This is used in a bunch of places, so getting this right is going to be critical. It has to satisfy what was stated in https://issues.redhat.com//browse/CFE-764 and also has to work for renderDC code paths as well as storage code paths.

pkg/cli/mirror/util.go
```diff
@@ -113,3 +128,326 @@ func (o *MirrorOptions) checkErr(err error, acceptableErr func(error) bool, logM
...
+• map[OperatorCatalogPlatform][]name.Digest: If no error occurs, returns a map whose key is Platform
+and value is a slice of one or more image digest references for that architecture.
+If an error occurs the map will always be initialized (i.e non-nil) but could have partial results.
+
+• error: non-nil if an error occurs, nil otherwise
+*/
+func getImageDigests(ctx context.Context, imageRef string, layoutPath *layout.Path, insecure bool) (map[OperatorCatalogPlatform][]name.Reference, error) {
+
+	// initialize return values
+	digestsMap := map[OperatorCatalogPlatform][]name.Reference{}
```
sherine-k, line 198: Could there be more than one manifest per platform?
[]name.Reference contains the reference to manifest for that specific platform, not all digests. There's something I'm missing

pkg/cli/mirror/util.go
```diff
@@ -113,3 +128,326 @@ func (o *MirrorOptions) checkErr(err error, acceptableErr func(error) bool, logM
...
+
+# Returns
+
+• map[OperatorCatalogPlatform][]name.Digest: If no error occurs, returns a map whose key is Platform
+and value is a slice of one or more image digest references for that architecture.
+If an error occurs the map will always be initialized (i.e non-nil) but could have partial results.
+
+• error: non-nil if an error occurs, nil otherwise
+*/
+func getImageDigests(ctx context.Context, imageRef string, layoutPath *layout.Path, insecure bool) (map[OperatorCatalogPlatform][]name.Reference, error) {
```
sherine-k, line 195: Is it a possibility to return a map using `imageInfo` (like in catalog_images), so that, instead of keeping just the digest?
Is it possible to find a map that can be used and completed (with create.go's `createOlmArtifactsForOCI, getImageDigests, renderDC, plan, etc)?
I was wondering if this would help reduce the burden of "rediscovering" the image contents at multiple places, sometimes through reading manifests, other times through walking the FS

pkg/cli/mirror/catalog_images.go
jchunkins: I see the issue... single arch images won't have a platform associated with the image, and this code needs to handle this situation so it conditionally uses a matcher depending on if a hash was found. I've fixed the code and added a test. Since I can't be sure what e2e test was run here, I am going to leave this conversation open for now.

pkg/cli/mirror/util.go
jchunkins: I did it this way to be more defensive about the references we discover along the way. There's two scenarios:

1. somebody duplicates a manifest entry for a given platform
2. nested "image indexes" are discovered resulting in a duplicate digest for a given platform

Neither of these are very likely, and I can totally see the confusion this causes. I can change this to a single reference rather than a slice if you'd prefer.

jchunkins: I've refactored things, and as a result this is now a single reference rather than a slice.

pkg/cli/mirror/util.go
jchunkins: I took this idea and refactored the render code path to standardize on `map[OperatorCatalogPlatform]CatalogMetadata` so that we can capture new information in this map as the code goes through `getImageDigests` -> `renderDCFunc` -> `plan`. To do this, renderDCFunc now uses this map as an in/out parameter.

pkg/config/load_test.go
```diff
@@ -184,27 +185,78 @@ mirror:
   - catalog: registry.com/ns/baz:v1.2
     targetTag: v1.3
   - catalog: registry.com/ns/baz:v1.2
+  - catalog: oci:///tmp
```
sherine-k, line 188: Some unit tests for GetUniqueName are in pkg/api/v1alpha2/types_config.go
Do you think it makes sense to move these tests there?
TBH I don't understand what load_test.go is for.

pkg/cli/mirror/operator.go
```diff
@@ -65,8 +65,14 @@ func (o *OperatorOptions) PlanFull(ctx context.Context, cfg v1alpha2.ImageSetCon
...
 // PlanDiff plans only the diff between each old and new catalog image pair
 func (o *OperatorOptions) PlanDiff(ctx context.Context, cfg v1alpha2.ImageSetConfiguration, lastRun v1alpha2.PastMirror) (image.TypedImageMapping, error) {
-	f := func(ctx context.Context, reg *containerdregistry.Registry, ctlg v1alpha2.Operator) (*declcfg.DeclarativeConfig, v1alpha2.IncludeConfig, error) {
-		return o.renderDCDiff(ctx, reg, ctlg, lastRun)
+	// Wrapper renderDCDiff so it satisfies the renderDCFunc function signature.
+	f := func(
+		ctx context.Context,
+		reg *containerdregistry.Registry,
+		ctlg v1alpha2.Operator,
+		digestsToProcess map[OperatorCatalogPlatform]CatalogMetadata,
```
sherine-k, line 73:
```suggestion
		archSpecificSubCatalogs map[OperatorCatalogPlatform]CatalogMetadata,
```

pkg/cli/mirror/operator.go
```diff
@@ -82,7 +88,55 @@ func (o *OperatorOptions) complete() {
...
+/*
+CatalogMetadata represents the combination of a DeclarativeConfig, and its associated
+IncludeConfig, which is either provided by the user directly or generated from the contents
+of DeclarativeConfig
+*/
+type CatalogMetadata struct {
+	dc         *declcfg.DeclarativeConfig // a DeclarativeConfig instance
+	ic         v1alpha2.IncludeConfig     // a IncludeConfig instance
+	catalogRef name.Reference             // the reference used to obtain DeclarativeConfig and IncludeConfig
+}
```
sherine-k, line 111: I like this

aguidirh:
Hi @jchunkins,

Could you please convert this PR from draft to WIP ? Adding WIP in the title will prevent it to be merged.

This way we know where the tests are failing and we can fix them.

pkg/cli/mirror/operator.go
jchunkins: I wend with a slightly different name `catalogMetadataByPlatform`

pkg/cli/mirror/create.go
```diff
@@ -8,17 +8,19 @@ import (
...
 	"github.com/openshift/oc-mirror/pkg/api/v1alpha2"
+	oc "github.com/openshift/oc-mirror/pkg/cli/mirror/operatorcatalog"
 	"github.com/openshift/oc-mirror/pkg/config"
 	"github.com/openshift/oc-mirror/pkg/image"
 	"github.com/openshift/oc-mirror/pkg/metadata/storage"
 )
 
 // Create will plan a mirroring operation based on provided configuration
-func (o *MirrorOptions) Create(ctx context.Context, cfg v1alpha2.ImageSetConfiguration) (v1alpha2.Metadata, image.TypedImageMapping, error) {
+func (o *MirrorOptions) Create(ctx context.Context, cfg v1alpha2.ImageSetConfiguration) (v1alpha2.Metadata, image.TypedImageMapping, map[string]map[oc.OperatorCatalogPlatform]oc.CatalogMetadata, error) {
```
sherine-k, line 23: Does it make sense to include `map[string]map[oc.OperatorCatalogPlatform]oc.CatalogMetadata` in `MirrorOptions` in order to pass it more seemlessly accross methods?

pkg/cli/mirror/fbc_operators.go
```diff
@@ -38,7 +41,6 @@ const (
 	configPath          string = "configs/"
 	catalogJSON         string = "/catalog.json"
 	relatedImages       string = "relatedImages"
-	configsLabel        string = "operators.operatorframework.io.index.configs.v1"
```
sherine-k, line 41: It is safer to use the configsLabel: the configs folder is not always called this way as we learned a few months back

sherine-k: Never mind this comment

pkg/cli/mirror/mirror.go
```diff
@@ -720,7 +721,7 @@ func (o *MirrorOptions) mirrorToMirrorWrapper(ctx context.Context, cfg v1alpha2.
 			return err
 		}
 		workspace := filepath.Join(o.Dir, config.SourceDir)
-		if err = metadata.UpdateMetadata(ctx, sourceBackend, &meta, workspace, o.SourceSkipTLS, o.SourcePlainHTTP); err != nil {
+		if err = metadata.UpdateMetadata(ctx, sourceBackend, &meta, workspace, o.SourceSkipTLS, o.SourcePlainHTTP, allCatalogs); err != nil {
```
sherine-k, line 724: Since we're passing the map to UpdateMetadata, we should check that the workflow MirrorToDisk + DiskToMirror is not impacted by this change.
If it is working, we can even extend this nice idea, and pass the map to rebuildCatalogs, thus completely bypassing the necessity to walk the folder to "discover" the catalogs that need rebuilding.

pkg/image/image.go
```diff
@@ -92,6 +92,7 @@ func ParseReference(ref string) (TypedImageReference, error) {
 
 	dstType := DestinationOCI
 
+	// TODO: this assumes you can parse it into a docker image ref (and you can’t do that since its just a path).
```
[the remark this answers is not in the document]
jchunkins: I get what you're saying... DockerImageReference has to be provided, but it is not really a valid docker reference for an OCI path. The DockerImageReference is constructed with the values that come from the result of `v1alpha2.ParseImageReference`, but this content is going to be variable depending on the path that's feed into `v1alpha2.ParseImageReference`.

In my next push, there's a test case called `TestV1A2ParseImageReferenceOCIRefs` in `image_test.go` which covers all of the bases of what `v1alpha2.ParseImageReference` does with an OCI path (or at least I think it does).

Since `image.ParseImageReference` function is called all over the place, it's hard to talk about specifically, but one example is in create.go in the `run` function. It iterates over the `[]v1alpha2.Operator` slice, and uses the resulting `parsedRef.Ref.Name` as an image reference. If you look in the test case I mentioned, that equates to the `expRepo` field in the test case values. Essentially the last segment of the path is used as the reference. In some edge cases, the parsing results is an empty string (e.g. if you've been given `/` as the path or if the path is relative without any directory name provided). These edge cases are likely to result in an error (and maybe that's OK).

To flesh out what I am talking about:
1. Start with "oci:///Users/jhunkins/temp/ocmirror" as a reference
2. This becomes "ocmirror"
3. Now "ocmirror" gets converted into a go container registry reference and stored in `CatalogMetadata.CatalogRef`. That effectively looks something like this:

```
index.docker.io/library/ocmirror@sha256:dc1b9846d7994450e74b9cde2e621f8c1d98cdf1debd591db291785dd3fc6446
```

In the end we have various ways to prevent `CatalogMetadata.CatalogRef` from being used as a real image reference. We just need to be vigilant to not accidentally use `image.ParseImageReference` results in ways that don't take into account the fact that it could be a bogus reference for OCI refs. My comment was intended to drive this conversation. I can change the comment to just be a warning that states this is generating a bogus docker image reference for OCI and leave it at that.

pkg/cli/mirror/create.go
jchunkins: I considered doing this, but I figured if I did that, we might need to use some form of synchronization (i.e. sync.Mutex or sync.Map) if this code ever used multi threading. Passing the map around like this avoids this issue since it is not shared. Also, as it is right now, testing individual functions is a bit easier since I don't have to construct MirrorOptions for those tests.

I've already done the painful part of refactoring things, so it is easier to leave the code as-is. If you'd prefer it, I can change it to storing it in MirrorOptions.

pkg/cli/mirror/fbc_operators.go
```diff
@@ -300,96 +302,181 @@ func addCatalogToMapping(catalogMapping image.TypedImageMapping, srcOperator v1a
...
+	returnPath := filepath.Join(extractedImageDir, configsPrefix)
 
-	//Use the label in the config layer to determine the
-	//folder containing the related images, when untarring layers
-	cfgDirName, err := getConfigPathFromConfigLayer(imagePath, string(manifest.ConfigInfo().Digest))
-	if err != nil {
-		return "", err
-	}
-	// iterate through each layer
+	tr := tar.NewReader(mutate.Extract(img))
```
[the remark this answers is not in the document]
jchunkins: Yes, indeed. Seemed like a better way to go. I did hit a scenario where it is difficult to determine if an error occurs during extraction (e.g. we have a bad layer test case) but I'd not expect that to happen in normal situations. I've handled this with sanity check after extraction to see if the output directory exists and has content.

pkg/config/load_test.go
jchunkins: This test loads the config once and then tests each entry with GetUniqueName. So it tests the loading functionality in a round about way. The other test in pkg/api/v1alpha2/types_config.go just uses static config objects. Leaving here for now

sherine-k:
Hi @zhouying7780 
This PR introduces a new feature - multi-arch support for catalogs.
It also refactors the way catalogs are mirrored (for **all workflows**: mirrorToMirror, mirrorToDisk + diskToMirror, oci **as well as non-oci**). 
So can you help us validate this PR against regressions, and also add needed test cases to cover the multi-arch scenarios?
Thank you

pkg/cli/mirror/mirror.go
```diff
@@ -605,7 +606,7 @@ func (o *MirrorOptions) mirrorToMirrorWrapper(ctx context.Context, cfg v1alpha2.
 	if err := bundle.MakeWorkspaceDirs(o.Dir); err != nil {
 		return err
 	}
-	meta, mapping, err := o.Create(ctx, cfg)
+	meta, mapping, allCatalogs, err := o.Create(ctx, cfg)
 	if err != nil {
```
Prashanth684, line 610: this feature probably needs oc-mirror to grow a new option to mirror the manifestlist and keep the legacy behavior as such.

zhouying7780:
I tested this pr , and this is the result , when I use the `oc image info` to check the mirrored image , I found the mirror image is not correct : 
### oc image info localhost:5000/mulitoperator/cpopen/ibm-zcon-zosconnect-catalog:6f02ec --insecure  --filter-by-os=linux/amd64
Warning: the default reading order of registry auth file will be changed from "${HOME}/.docker/config.json" to podman registry config locations in the future version of oc. "${HOME}/.docker/config.json" is deprecated, but can still be used for storing credentials as a fallback. See https://github.com/containers/image/blob/main/docs/containers-auth.json.5.md for the order of podman registry config locations.
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS DIGEST
  /  sha256:777486079eb21efd26af3d50332f5a70e1fd24a95245227dac41aab6d8871832

zhouying7780:
if we mirrored *manifestlist.DeserializedManifestList type , then we should mirror  linux/amd64 and  linux/s390x
if we only mirror linux/amd64 , then we should not mirror the *manifestlist.DeserializedManifestList type

zhouying7780:
And I think we should support to filter the architectures like the ocp release image . This is also needed by customer. This could be as new feature request.

zhouying7780:
And also we need to support list operators which is multi-arch.

jchunkins:
@zhouying7780 It's not entirely clear to me what scenario you ran, but I suspect that my change in commit https://github.com/openshift/oc-mirror/pull/568/commits/97ffd13fb74e7d00df1cf86374a5862f8fa2f2f6 resolves the issue you hit. I'd be interested to know if that fixes the issue or not. If it does not, I'd like to know the exact scenario you ran so I can look at it more closely.

pkg/api/v1alpha2/types_config.go
```diff
@@ -172,17 +172,28 @@ func (o Operator) GetUniqueName() (string, error) {
 		}
 		uniqueName = path.Join(uniqueName, name)
 	}
-	if tag != "" {
-		uniqueName = uniqueName + ":" + tag
+
+	if tag == "" && id == "" {
+		// if we get here without a tag or digest, default to latest tag
+		uniqueName = uniqueName + ":" + "latest"
```
sherine-k, line 178: OCPBUGS-2633 is one case among many where the use of latest by default makes oc-mirror fail.

pkg/cli/mirror/mirror.go
jeffdyoung: +1

sherine-k:
Hi @jchunkins 
As discussed last Thursday, this PR impacts all other oc-mirror users on execution time. Our analysis this week showed us it might not be aligning to what the rest of the CNCF is doing.
There seems to be a big misunderstanding about the expectations around multi-arch support.

With multi-arch, the idea is to get the same software on different hardware.
For operator catalogs that are built downstream, all image references (operators, operands, other related images) are manifest-lists. The philosophy is: The runtime will auto-select the correct arch depending on the hardware running underneath. 
image

The current version of oc-mirror in fact already mirrors ALL the images from ALL platforms available in each of these manifest-lists. We verified this this week. 

Although this PR will not have significant impact on mirroring volume, performance will be heavily impacted. 
But more importantly, with this PR, we would allow a way to get slightly different software depending on the hardware.  
Doing this would be a stark difference to how the rest of Kube and the CNCF ecosystem looks at this.
All packaging, from plain YAML, kustomize, Helm to OLM (CSV, bundle, catalog) is arch-less. The only multi-arch thing is the images themselves containing differently compiled binaries. 

Maybe it's important to understand better your use case. It seems that the catalog content you deal with is not the same between a catalog's platformA manifest, and platformB manifest. 
Can you elaborate more?

One thing to add is that OLM on cluster does not expect the architecture of the bundle to matter. Because bundles and catalogs are arch-less in OLM.

pkg/cli/mirror/options.go
```diff
@@ -14,29 +14,29 @@ import (
...
+	SkipMissing                bool   // If an input image is not found, skip them.
+	SkipMetadataCheck          bool   // Skip metadata when publishing an imageset
+	SkipPruning                bool   // If set, will disable pruning globally
+	ContinueOnError            bool   // If an error occurs, keep going and attempt to complete operations if possible
+	IgnoreHistory              bool   // Ignore past mirrors when downloading images and packing layers
+	MaxPerRegistry             int    // Number of concurrent requests allowed per registry
+	UseOCIFeature              bool   // Use the new oci feature for oc mirror (oci formatted copy)
+	OCIRegistriesConfig        string // Registries config file location (used only with --use-oci-feature flag)
+	OCIInsecureSignaturePolicy bool   // If set, OCI catalog push will not try to push signatures
 	MaxNestedPaths             int
```
lmzuccarelli, line 40: +100 for adding useful comments on each flag

jchunkins:
Closing this in favor of the replacement PR https://github.com/openshift/oc-mirror/pull/611

PR #623, opened 2023-04-28
OCPBUGS 11922: Limit the nested repository path while mirroring the images

lmzuccarelli:
# Description

This fix addresses the issue when mirroring to repositories with limited (restricted paths, typically  oc-mirror fails with 401 unauthorized for some registries when nested path exceeds the expected maximum path-components

The fix consists of having oc-mirror use the flag --max-nested-paths. 

For example a repository to mirror to could be something like

`docker://registry.gitlab.com/xxx/yyy`

When mirroring the nested paths could expand to 

`docker://registry.gitlab.com/xxx/yyy/zzz` 

in this case mirroring will fail with 401 unauthorized if we add one more nested path as indicated

To avoid this scenario oc-mirror will use the --max-nested-paths flag and then calculate the registry+username (in the example for `docker://registry.gitlab.com/xxx/yyy` the count is 3 -> registry + user namespace)

If the max nested paths is set to 4, namespaces and component names will then have all instances of "/" converted to "-"

This will allow for successful mirroring.

Fixes # OCPBUGS-11922

## Type of change

# How Has This Been Tested?

Unit tests have been implemented to cover edge cases.

To ensure that the mirroring works correctly here  is a manual test case to validate the mapping.tx and ImageContentSourcePolicy files

ImageSetConfig to use

```
kind: ImageSetConfiguration
storageConfig:
  local:
    path: /tmp/lmz
mirror:
  platform: {}
  operators:
    - catalog: redhat.registry.io/redhat/certified-operator-index:v4.12
      packages: 
      - name: gitlab-runner-operator
        channels: 
        - name: stable
  additionalImages: []
  helm: {}
```

Command line to execute (change the source-use-http flag as I used a local copy of the operator)

```
oc-mirror --config=./isc.yaml docker://localhost.localdomain:5000/ocpbugs-11922/test-paths --continue-on-error --source-use-http --dest-use-http --max-nested-paths 4
```
Once this completes execute the following command

```
oc adm catalog mirror localhost.localdomain:5000/redhat/certified-operator-index:v4.12 localhost.localdomain:5000/ocpbugs-11922/test-release -a /home/lzuccarelli/.docker/config.json  --index-filter-by-os='linux/amd64'  --manifests-only  --to-manifests=./local-tmp   --max-components=4  --skip-verification=false --insecure=true
```
Compare the output found in `oc-mirror-workspace/results-xxxx/mapping.txt` to 
`./local-tmp/mapping.txt | grep git`

Compare the output found in `oc-mirror-workspace/results-xxxx/imageContentSourcePolicy.yaml` with that of 
`./local-tmp/imageContentSourcePolicy.yaml | grep git`

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:


pkg/cli/mirror/mirror.go
```diff
@@ -445,10 +445,9 @@ func (o *MirrorOptions) mirrorMappings(cfg v1alpha2.ImageSetConfiguration, image
 			Type: srcRef.Type,
 		}
 
-		dstTIR := imagesource.TypedImageReference{
-			Ref:  dstRef.Ref,
-			Type: dstRef.Type,
-		}
+		// OCPBUGS-11922
+		dstTIR := o.processNestedPaths(&dstRef)
+
```
sherine-k, line 450: Although this works, and mirrors correctly the images, the original TypedImageMapping (`images`) is not modified, and is not returned by this method. 
Therefore, the mapping.txt that is written L818 doesn't contain the renamed images, and nore will the ImageContentSourcePolicy that is generated from it.

pkg/cli/mirror/mirror.go
```diff
@@ -816,7 +815,7 @@ func (o *MirrorOptions) mirrorToDiskWrapper(ctx context.Context, cfg v1alpha2.Im
 
 	mappingPath := filepath.Join(o.Dir, mappingFile)
 	if o.DryRun {
-		if err := writeMappingFile(mappingPath, mapping); err != nil {
+		if err := o.writeMappingFile(mappingPath, mapping); err != nil {
```
sherine-k, line 818: The mapping file that we have at this point doesn't contain the image renaming

sherine-k (changes requested):
Hi @lmzuccarelli 
We have a few areas where we have doubts still, and we want to  work a bit on this PR, also using the feedback from the customer tests.

sherine-k:
@zhouying7780 
If you prefer, since this merged, you can now test it directly from main.
We will wait for your heads up in order to start backporting.

PR #630, opened 2023-05-09
CFE-805: Check StorageConfig's sequence for both OCI and non-oci cases, since …

sherine-k:
…metadata is used in both


sherine-k:
This PR is a duplicate of #649 
Closing in favor of that PR

PR #634, opened 2023-05-16
OCPBUGS-11922: paths not needed in ICSPBuilder interface

sherine-k:
# Description
This is a follow up of #623 
Handing the maxNestedPaths as argument all the way to GenerateICSP is not needed anymore. 
This is rather handled by passing the scope to the function, when maxNestedPaths is set, as done in https://github.com/openshift/oc-mirror/blob/5ae42b8e206e972d74b4fc2232bc0b82552915cc/pkg/cli/mirror/mirror.go#L552

Fixes # OCPBUGS-11922
## Type of change


aguidirh (commented):
/lgtm

PR #644, opened 2023-05-26
adds Jeremy Peterson to OWNERS file

aguidirh:


soltysh:
Overriding valid-bug, this is only updating owners

PR #651, opened 2023-05-31
CFE-825: As a oc-mirror user, I would like mirrored operator catalogs to have valid caches

sherine-k:
# Description

This PR implements the changes described in [CFE-825](https://issues.redhat.com/browse/CFE-825).
This is a Work In Progress. 

List of Done:
* Extracts the opm binary from the catalog
   * Accounts for remote catalogs
   * Accounts for local oci catalogs
   * Accounts for multi-arch local oci catalogs
     * by extracting binaries from the manifest that corresponds to the machine's platform and os
 * Assumes that the opm binaries are suffixed opm

List of ToDo:
* If opm version is 1.25 or greater, opm serve --cache-only should work. Otherwise, the cache generation should not be done
* Implement logic for the case where the catalog doesn't have the cache feature implemented
* (Implement?) and test on different platforms, to validate that the opm binary extraction is correct
* Update test-catalog images used by E2E tests in quay.io/redhatgov
* To be discussed: how to unit test easily this code - requires a fair load of mocking, and changes in function signatures to inject mocks.
* To be discussed: how to improve performance: for each catalog, an extraction (and pull of image) is done using operator-registry code to render declarativeConfig. Then another extraction (and pull) is done to extract opm binary (uses crane)
* Tests against official redhat catalogs in various versions >=4.11

Fixes [CFE-825](https://issues.redhat.com/browse/CFE-825).

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?
 * Tested manually with registry.redhat.io/redhat/redhat-operator-index:v4.13
 * Tested manually with OCI local catalog multi-arch copied from above catalog
 * In both cases, run the mirrored catalogs with
 ```
podman run -it localhost:5000/cfe-825/redhat/redhat-operator-index:v4.13 serve /configs --cache-only --cache-dir=/tmp/cache --cache-enforce-integrity
```

# Checklist:

### CODE REVIEW

File: pkg/cli/mirror/catalog_images.go [hunk 1]
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		}
+
+		var buf bytes.Buffer
+		_, err = buf.ReadFrom(tr)
+		if err != nil {
+			return err
+		}
+
+		targetFileName := filepath.Join(outDir, header.Name)
+		bytes := buf.Bytes()
```
jchunkins, line 415: Not sure where else to put this commentary, so including it here. There's a long line of different base images that have been used over time, and the opm executable could literally be anywhere. Here's a brief description of upstream and downstream images, to try to highlight some of the complexity you're going to run into.

## Upstream images

### Upstream images used over time (not an exhaustive list)

quay.io/operator-framework/upstream-opm-builder
quay.io/operator-framework/upstream-registry-builder 
quay.io/operator-framework/opm:latest                <- this is the current image


### Upstream possible paths (some of these could have been a symbolic link)

/opm      <- opm at root is one possibility!
/bin/opm

## Downstream image

### Downstream (not an exhaustive list)

registry.redhat.io/openshift4/ose-operator-registry:v4.13

### Downstream paths for 4.13

bin/darwin-amd64-opm -> /bin/registry/darwin-amd64-opm
bin/opm -> /bin/registry/opm
bin/windows-amd64-opm -> /bin/registry/windows-amd64-opm

---

Symbolic links are going to be the most difficult things to handle here. In theory you could have a opm symbolic link within the bin directory pointing to something outside the bin directory. None of the examples I show above do this, but it is something that COULD happen. Also the opm executable could be at the root.

At any rate, the section of code I have highlighted here should be checking if the number of bytes is zero. The reason for this is because according to the [tar read docs](https://pkg.go.dev/archive/tar#Reader.Read):

> Calling Read on special types like TypeLink, TypeSymlink, TypeChar, TypeBlock, TypeDir, and TypeFifo returns (0, io.EOF) regardless of what the Header.Size claims.

So your byte buffer could have zero bytes if the file is a symbolic link. In the downstream path examples above, you'd attempt to write out `bin/darwin-amd64-opm` with zero bytes and `/bin/registry/darwin-amd64-opm` which is a real executable which should have N number of bytes. So there's no point in writing out symbolic links since they won't really be useful.

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		if err != nil {
+			return err
+		}
+		opmBinaryExtracted = targetFileName
+	}
+	// check for the folder (it should exist if we found something)
+	_, err = os.Stat(opmBinaryExtracted)
+	if errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("opm binary not found after extracting %q from catalog image %v", opmBinaryExtracted, srcRef)
+	}
```
jchunkins, line 440: Since you're looping through the tar file and finding potentially more than one opm file, this is effectively only checking the last one you found. Maybe that's OK? Also the comment says a folder, but it actually checking a file

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		return "", fmt.Errorf("error finding the extracted opm binary %s while preparing to run opm to regenerate cache: %v", opmCmdPath, err)
+	}
+	err = os.Chmod(opmCmdPath, 0744)
+	if err != nil {
+		return "", fmt.Errorf("error changing permissions to the extracted opm binary while preparing to run opm to regenerate cache: %v", err)
+	}
+	return opmCmdPath, nil
+}
+
+func extractOPMBinary(ctx context.Context, srcRef image.TypedImageReference, outDir string, insecure bool) error {
```
jchunkins, line 323: function docs might be good

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
 	return nil
 }
 
+func findOpmCmd(artifactDir string) (string, error) {
```
jchunkins, line 283: function docs might be good

pkg/cli/mirror/catalog_images.go
```diff
@@ -195,22 +209,53 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
 		if err != nil {
 			return fmt.Errorf("error creating deleted layer: %v", err)
 		}
 
-		// Delete must be first in the slice
-		// so that the /configs directory is deleted
-		// and then add back with the new FBC.
-		layers := []v1.Layer{deleted, add}
+		// white out layer /tmp
+		// TODO : any label in the manifests that might indicate more accurately where the cache is?
```
jchunkins, line 225: I don't think you're going to find labels for where the cache is, but you do have the docker command that tells the image how the image should be started. 

The CMD values come from the docker file that builds the image. An example of how docker files are created for catalog images can be found in [generate_dockerfile.go](https://github.com/operator-framework/operator-registry/blob/v1.25.0/alpha/action/generate_dockerfile.go). That link is the very first release of operator-registry that contains the `--cache-dir` argument. There's also examples in operator-sdk project too, but I've not included them here.

For example:
```
ENTRYPOINT ["/bin/opm"]
CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
```

You could obtain the `cmd` value from the image config and attempt to find the `--cache-dir` argument and get its value. You should probably attempt to handle both variations where the equals sign is used versus separate arguments, just to be safe:

CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
CMD ["serve", "/configs", "--cache-dir", "/tmp/cache"]

If you process `cmd` and can't find the argument, you could just assume `/tmp` as a fallback.

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+	opmBin := "opm"
+
+	if runningOS != "linux" {
+		opmBin = strings.Join([]string{runningOS, runningArch, opmBin}, "-")
+	}
+	opmCmdPath := ""
+	binaryDir := filepath.Join(wd, artifactDir, config.OpmBinDir)
+	err = filepath.Walk(binaryDir, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			fmt.Printf("prevent panic by handling failure accessing a path %q: %v\n", path, err)
```
jchunkins, line 299: Did you really want this to show up in standard out or just log it?

pkg/cli/mirror/catalog_images.go
```diff
@@ -195,22 +209,53 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		absConfigPath, err := filepath.Abs(filepath.Join(artifactDir, config.IndexDir))
+		if err != nil {
+			return fmt.Errorf("error getting absolute path for catalog's index %v: %v", filepath.Join(artifactDir, config.IndexDir), err)
+		}
+		absCachePath, err := filepath.Abs(filepath.Join(artifactDir, config.TmpDir))
+		if err != nil {
+			return fmt.Errorf("error getting absolute path for catalog's cache %v: %v", filepath.Join(artifactDir, config.TmpDir), err)
+		}
+		//TODO call opm serve /configs –-cache-dir /tmp/cache –-cache-only
+		cmd := exec.Command(opmCmdPath, "serve", absConfigPath, "--cache-dir", absCachePath, "--cache-only")
```
jchunkins, line 244: In slack you mentioned that Joe said you should do this:

> I think you’ll want to run opm serve <configDir> --cache-dir=<cacheDir> --cache-only --cache-enforce-integrity 

This seems to be missing the `--cache-enforce-integrity`

jchunkins: TODO can be removed from the comment I think

pkg/cli/mirror/catalog_images.go
```diff
@@ -195,22 +209,53 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		fmt.Printf("%s\n", cmd.String())
```
jchunkins, line 245: Did you want this to go to standard out or just log?

pkg/cli/mirror/catalog_images.go
```diff
@@ -223,8 +268,10 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
 				containertools.ConfigsLocationLabel: "/configs",
 			}
 			cfg.Config.Labels = labels
-			cfg.Config.Cmd = []string{"serve", "/configs"}
-			cfg.Config.Entrypoint = []string{"/bin/opm"}
+			// Although it was prefered to keep the entrypoint and command as it was
+			// we couldnt guarantie that the cache-dir was /tmp/cache, and therefore
+			// we had to specify the command for the newly built catalog
+			cfg.Config.Cmd = []string{"serve", "/configs", "--cache-dir=/tmp/cache"}
```
jchunkins, line 274: See my other comment about cache dir discovery. I wonder if this needs to be more selective about the arguments you update here. This just completely overwrites the command with hard coded values... what if future versions of opm need different arguments to start?

jchunkins: If we're overriding the --cache-dir argument always, then maybe the discovery part does not matter? Something to ponder.

jchunkins (changes requested):

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +280,167 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+				if err != nil {
+					return err
+				}
+				childIndexManifest, err := childIndex.IndexManifest()
+				if err != nil {
+					return err
+				}
+
+				// at this point, find the first image and store it for later if possible
+				//TODO extract the child index that corresponds to this machine's architecture
```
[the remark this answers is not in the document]
sherine-k: Correct. Sorry, I failed to erase some of my initial `TODO` comments

lmzuccarelli:
@jchunkins - thanks for comprehensive review

lmzuccarelli:
@sherine-k - Once we get feedback from QE lets get this PR merged. You have addressed most of John's suggestions.

pkg/cli/mirror/catalog_images.go
sherine-k: Done

pkg/cli/mirror/catalog_images.go
jchunkins: Sorry I don't think I was clear enough. I think that this code should find `--cache-dir` and **_only modify its contents_** and leave the rest of the arguments alone. As I mentioned in https://github.com/openshift/oc-mirror/pull/651#discussion_r1284725550 there's two ways the argument could be potentially specified:

"--cache-dir=/tmp/cache"
"--cache-dir", "/tmp/cache"

The reason for this is to protect the "serve" arguments in case they change in the future (e.g. adding a new positional argument or flag argument). Hard coding `[]string{"serve", "/configs", "--cache-dir=/tmp/cache"}` would break any future changes that might be made to the serve command.

My second comment regarding "discovery" was about trying to see what the current value for `--cache-dir` is and using that path when we change `cfg.Config.Cmd`. In my mind that's optional because if we just change to `/tmp/cache` always then the discovery portion is not necessary.

jchunkins:
@sherine-k I added a new comment in https://github.com/openshift/oc-mirror/pull/651#discussion_r1285957697 but the comment thread was marked as resolved and the buttons for marking it as unresolved are not present. I wanted to make sure you saw this reply.

pkg/cli/mirror/catalog_images.go
```diff
@@ -25,6 +31,23 @@ import (
 	"github.com/openshift/oc-mirror/pkg/image/builder"
 )
 
+const (
+	opmCachePrefix           = "/tmp/cache"
+	opmBinarySuffix          = "opm"
+	opmBinaryDir             = "usr/bin/registry"
+	cacheLocationPlaceholder = "cacheLocation.txt"
```
jchunkins, line 38: Seems like this should go into pkg/config/metadata.go as a constant and update the "tree" diagram in that file

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +310,249 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+	}
+	return opmCmdPath, nil
+}
+
+// extractOPMAndCache is usually called after rendering catalog's declarative config.
+// it uses crane modules to pull the catalog image, select the manifest that corresponds to the
+// current platform. It then extracts from that image any files that are suffixed `*opm` for later
+// use upon rebuilding the catalog: This is because the opm binary can be called `opm` but also
+// `darwin-amd64-opm` etc.
+func extractOPMAndCache(ctx context.Context, srcRef image.TypedImageReference, outDir string, insecure bool) error {
```
jchunkins, line 361: Nit: the  `outDir` seems to be the catalog source directory... maybe it should be called that for clarity?

pkg/cli/mirror/catalog_images.go
```diff
@@ -195,22 +218,74 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
-		// and then add back with the new FBC.
-		layers := []v1.Layer{deleted, add}
+		if withCacheRegeneration {
+			// read the location of the cache
+			dat, err := os.ReadFile(filepath.Join(artifactDir, cacheLocationPlaceholder))
+			if err != nil {
+				return fmt.Errorf("unable to determine location of cache for image %s. Cache generation failed: %v", ctlgRef, err)
+			}
+			cacheLocation := string(dat)
+			cacheLocationElmts := strings.Split(strings.TrimPrefix(cacheLocation, string(os.PathSeparator)), string(os.PathSeparator))
```
jchunkins, line 252: Nit: I've not tested this, but doesn't `filepath.SplitList()` handle this?

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +310,249 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+			return fmt.Errorf("opm binary not found after extracting %q from catalog image %v", opmBinaryExtracted, srcRef)
+		}
+	}
+
+	return nil
+}
+
+// getPtfImageFromOCIIndex takes an oci local image located in `fbcPath` and finds the image that
+// corresponds to the current platform and OS within the manifestList or imageIndex
+func getPtfImageFromOCIIndex(fbcPath string, architecture string, os string) (v1.Image, error) {
```
jchunkins, line 466: Nit: is `Ptf` in this function name supposed to be platform? Took a while for me to reason out what this meant.

pkg/cli/mirror/catalog_images.go
```diff
@@ -233,6 +310,249 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
...
+		defer cfl.Close()
+	} else {
+		return err
+	}
+
+	_, err = cfl.Write([]byte(cachePath))
+	if err != nil {
+		return err
+	}
+	cfl.Close()
```
jchunkins, line 409: nit: isn't this handled at line 400 in the defer?

pkg/config/metadata.go
```diff
@@ -87,6 +87,12 @@ const (
 	// IndexDir is the location of the
 	// file-based catalog json file.
 	IndexDir = "index"
+	// TmpDir is the location of the
+	// catalog cache directory.
+	TmpDir = "tmp/cache"
```
jchunkins, line 92: nit: Might be good to update the workspace tree diagram with this new directory

pkg/cli/mirror/catalog_images.go
```diff
@@ -223,8 +298,10 @@ func (o *MirrorOptions) processCatalogRefs(ctx context.Context, catalogsByImage
 				containertools.ConfigsLocationLabel: "/configs",
 			}
 			cfg.Config.Labels = labels
-			cfg.Config.Cmd = []string{"serve", "/configs"}
-			cfg.Config.Entrypoint = []string{"/bin/opm"}
+			// Although it was prefered to keep the entrypoint and command as it was
+			// we couldnt guarantie that the cache-dir was /tmp/cache, and therefore
+			// we had to specify the command for the newly built catalog
+			cfg.Config.Cmd = []string{"serve", "/configs", "--cache-dir=/tmp/cache"}
```
jchunkins, line 304: Still seems like this update of the cmd with --cache-dir needs to be conditional based on if we added a new cache layer. It probably needs to take withCacheRegeneration into account

jchunkins (changes requested):

pkg/cli/mirror/catalog_images.go
sherine-k: Done, good catch

pkg/cli/mirror/catalog_images.go
sherine-k: Sure . renamed

pkg/cli/mirror/catalog_images.go
sherine-k: great catch! Thank you!

PR #657, opened 2023-06-22
MIXEDARCH-265: Enable mirroring of the multi release image

jeffdyoung:
# Description

Enable mirroring the 'multi' release payload.

## Type of change

## Passed Local tests

- [x]  make test-unit
- [x]  make test-e2e

Verified that I could mirror multi and single arch payloads to quay.io:

`./bin/oc-mirror --config mirror.yaml docker://quay.io/jeffdyoung`

mirror.yaml:
```
kind: ImageSetConfiguration
storageConfig: 
 registry:
   imageURL: quay.io/jeffdyoung/oc-mirror
mirror:
  platform:
    architectures:
      - "multi"
      - "arm64"
      - "amd64"
    channels:
    - name: stable-4.13
      type: ocp
    - name: fast-4.13
      type: ocp
```

### DIFF
--- a/pkg/cli/mirror/release.go
+++ b/pkg/cli/mirror/release.go
@@ -365,6 +365,7 @@ func (o *ReleaseOptions) getMapping(opts *release.MirrorOptions) (image.TypedIma
 
 	opts.IOStreams.Out = file
 	opts.ToMirror = true
+	opts.KeepManifestList = true
 	if err := opts.Validate(); err != nil {
 		return nil, err
 	}
--- a/docs/design/imageset-configuration.md
+++ b/docs/design/imageset-configuration.md
@@ -4,6 +4,7 @@ Design: oc-mirror Imageset Configuration
 - [Design: oc-mirror Imageset Configuration](#design-oc-mirror-imageset-configuration)
 - [Content Types](#content-types)
   - [Platforms](#platforms)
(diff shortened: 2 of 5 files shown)


pkg/cli/mirror/release.go
```diff
@@ -365,6 +365,7 @@ func (o *ReleaseOptions) getMapping(opts *release.MirrorOptions) (image.TypedIma
 
 	opts.IOStreams.Out = file
 	opts.ToMirror = true
+	opts.KeepManifestList = true
```
sherine-k, line 368: I'm wondering if there would be any impact for setting this to true for all architectures other than multi.
I don't think there is an impact... Just thought to mention it in case anyone thinks of anything.

sherine-k (approved):
Hi @jeffdyoung 
This PR looks fairly simple. 
I left a small comment for something I'm not certain about. 
Thank you so much

pkg/cli/mirror/release.go
jeffdyoung: @sherine-k I tested mirroring `multi` release image in parallel with the single arch `amd64` and `arm64` release images, and didn't observe errors or any difference in the mirroring behavior.

Maybe @ardaguclu or @soltysh can confirm?

PR #661, opened 2023-06-29
[release-4.13] OCPBUGS-16372: A variety of changes needed for correct operation with multi…

lmzuccarelli:
… architecture catalogs (#611)

* A variety of changes needed for correct operation

Functional Changes

pkg/cli/mirror/fbc_operators.go

- remove findFBCConfig & replace with extractDeclarativeConfigFromImage This is necessary because the layer processing did not handle whiteout files. Switched to go-containerregistry which handles this correctly.
- modify getManifest so that it is fat manifest aware, and returns the first manifest in the list, or for single images, returns that manifest
- Removed functions that are no longer necessary: GetCatalogConfigPath, getConfigPathFromConfigLayer, UntarLayers

pkg/cli/mirror/create.go

- make createOlmArtifactsForOCI manifest list aware and discovers the first image in the list, or for single images, uses that image directly
- calls replacement function extractDeclarativeConfigFromImage
- stores full path to the artifact in a map so this path does not need to be recalculated later

pkg/cli/mirror/operator.go

- make use of operatorCatalogToFullArtifactPath map to obtain the full path to the olm_artifacts directory and its contents and remove getOperatorCatalogRef() since it is no longer necessary
- make writeLayout() use the image config file to discover the platform when adding images, otherwise fall back to linux/amd64

pkg/cli/mirror/options.go

- add operatorCatalogToFullArtifactPath map[string]string

pkg/image/image.go

- replace getManifest with getFirstDigestFromPath which now gets the first digest for manifest lists, or the image digest for single image
- change logging behavior in ParseReference() when path does not exist and invalidate the ID in that case.

Non-functional Changes

pkg/cli/mirror/mirror.go

- removed redundant AsRespoitory() call since RepositoryName() already does this
- use constant for olm_artifacts directory

pkg/cli/options.go

- use constant for workspace

all files
- add documentation and change formatting for long function signatures
- if any file used deprecated io/ioutil, replace with io or os package as necessary
- update unit tests as necessary (the bulk of the files are related to new test cases)

* reverted test case result

- The original PR had concerns about using latest tag, so did not bring this change over, therefore this test case had to be reverted too.

* bring over remaining changes

This commit brings over the remaining changes needed for correct manifest list handling.

internal/testutils/testutils.go

- update and create new functions for creating single/multiarch images
- created *WithURL function variants to allow pre-existing registry in tests because the defaults server won't allow external calls while debugging

pkg/cli/mirror/catalog_images.go

- in rebuildCatalogs, make sure that the catalog source docker reference uses lower case values for "name" and "namespace" which is required to make valid docker references. This covers the case where a OCI filepath uses uppercase letters in its path.

pkg/cli/mirror/create.go

- add missing call to strip protocol from ctlg.OCIFBCPath when constructing a layoutPath

pkg/cli/mirror/mirror.go

- add missing initialization of operatorCatalogToFullArtifactPath

pkg/image/image.go

- in ParseReference, make sure that the docker reference uses lower case values for "name" and "namespace" which is required to make valid docker references. This covers the case where a OCI filepath uses uppercase letters in its path.

pkg/image/builder/image_builder.go

- make image builder handle single arch and multi arch images

pkg/image/builder/image_builder_test.go

- update test cases for image builder

update go modules and vendoring

* fix test case error message

* Change getFirstDigestFromPath based on feedback

- make function return error when number of manifests > 1
- do basic checks to ensure digest is valid before returning
- return image digest instead of config hash
- adjust documentation

* fix test case

* fix uninitialized error

* fix test case after rebase

* Fix OCPBUGS-13198

When copying the OCI layout, make sure we only copy files/folders that are part of the OCI layout specification.

The bug resulted because the copy destination was inside the source directory, which lead to an infinite circular copy operation when using the github.com/otiai10/copy package. This fix avoids this because it copies only the OCI layout files/folders and ignores others.

* refactor inner function to standalone per feedback

* refactor processImageIndex to be a method

- based on review, make inner function processImageIndex into a method of ImageBuilder
- add/update documentation

* remove unused getManifest func and test case

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:


lmzuccarelli:
/hold
Pending approval from PM , QE and Staff Eng

zhouying7780:
Done pre-merge regression test , no issue found

lmzuccarelli:
> /label qe-approved

@zhouying7780 thank you for the feedback !!!

lmzuccarelli:
We have received the go ahead from PM to merge this PR

Reference https://redhat-internal.slack.com/archives/C04TWUT9Y0Y/p1689100996267769

PR #730, opened 2023-11-10
DO_NOT_MERGE - CFE-977: Implementation of a blob gatherer for images in the local cache

sherine-k:
# Description

This PR is part of a larger user story [CFE-977](https://issues.redhat.com/browse/CFE-977) - Generate smaller archives

In this PR, we focus on identifying those blobs (both layers, config and manifests of various formats (docker-v2 or oci), both single arch and multi arch) that form a container image that is saved to oc-mirror's local cache.

The implementation relies on reading the storage folder of that local cache. 
While I agree that this is rather intrusive into the way the local cache registry stores the blobs on disk, it is a fairly easy implementation.

Another implementation might follow in another PR that could interact with the local cache registry instead of looking directly on disk.

For the moment, the use of BlobGatherer is not fully integrated with the MirrorToDisk workflow. Following PRs will be in charge of that.

Fixes # [CFE-979](https://issues.redhat.com//browse/CFE-979)

## Type of change

# How Has This Been Tested?

Unit tests for BlobGatherer where implemented

## Expected Outcome
Tests should pass


v2/pkg/archive/blob-gatherer.go
```diff
@@ -0,0 +1,97 @@
...
+	if strings.Contains(imgRef, ":") {
+		nameAndTag := strings.Split(imgRef, ":")
+		imgRef = nameAndTag[0]
+	}
+	return imgRef, nil
+}
+
+func isImageByDigest(imgRef string) bool {
+	return strings.Contains(imgRef, "@")
+}
```
aguidirh, line 87: Since this code is very similar with what is under v2/pkg/image, is it possible to reuse it?

v2/pkg/archive/blob-gatherer_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	for i, imgRef := range imgRefs {
+		actualIsByDigest := isImageByDigest(imgRef)
+
+		if actualIsByDigest != expectedisByDigest[i] {
+			t.Errorf("isImageByDigest() returned unexpected value for %q: got %v, want %v", imgRef, actualIsByDigest, expectedisByDigest[i])
+		}
+	}
+}
```
aguidirh, line 59: If possible to reuse the code under v2/pkg/image, moving these tests also under v2/pkg/image would be great

v2/pkg/archive/blob-gatherer.go
sherine-k: I can move this under v2/pkg/image, along with its unit test.

v2/pkg/archive/blob-gatherer_test.go
sherine-k: Good suggestion

v2/pkg/mirror/mirror.go
```diff
@@ -348,7 +348,3 @@ func parseMultiArch(multiArch string) (copy.ImageListSelection, error) {
 		return copy.CopySystemImage, fmt.Errorf("unknown multi-arch option %q. Choose one of the supported options: 'system', 'all', or 'index-only'", multiArch)
 	}
 }
-
-func HelloFromV2() string {
```
aguidirh, line 352: Thanks for removing this func.

sherine-k:
Closed in favor of #732

PR #733, opened 2023-11-20
PROJQUAY-6345: Skipping mirror failure if manifest not found during prune

bcaton85:
# Description

The pruning image process during the mirror operation has the possibility of failing when using RedHat Quay. In Quay, when tags are deleted the manifests are automatically eligible for garbage collection. When pruning manifest lists `oc mirror` will attempt to delete the child manifests by digest as well as the parent manifest. Since the child manifests are automatically eligible for garbage collection, they return 404 when `oc mirror` attempts to delete them directly. Since the garbage collection process is already handled within Quay, this change prevents a failure if the manifest returns a 404 during the image prune.

Fixes # [(issue)](https://issues.redhat.com/browse/PROJQUAY-6345)

## Type of change

# How Has This Been Tested?

Reproduced through steps provided in the mentioned issue.

## Expected Outcome
When `--skip-pruning-missing-manifests` is set to true the mirror will not fail if a 404 is returned when pruning manifests.

### DIFF
--- a/pkg/cli/mirror/prune.go
+++ b/pkg/cli/mirror/prune.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -162,7 +163,9 @@ func (o *MirrorOptions) pruneImages(deleter imageprune.ManifestDeleter, manifest
 
 				for _, manifest := range manifests {
 					err := deleter.DeleteManifest(k, manifest)
-					if err != nil {
+					if structuredErr, ok := err.(*transport.Error); ok && structuredErr.StatusCode == http.StatusNotFound {
+						klog.Infof("Manifest %s not found in repo %s", manifest, k)
+					} else if err != nil {
 						err = fmt.Errorf("repo %q manifest %s: %w", k, manifest, err)
 						errorsCh <- err
 					}
--- a/pkg/cli/mirror/prune_test.go
+++ b/pkg/cli/mirror/prune_test.go
@@ -550,6 +550,20 @@ func TestPruneImages(t *testing.T) {
 				"repo1|digest5",
 			},
 		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/cli/mirror/options.go
```diff
@@ -73,6 +74,7 @@ func (o *MirrorOptions) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.OCIRegistriesConfig, "oci-registries-config", o.OCIRegistriesConfig, "Registries config file location (it works only with local oci catalogs)")
 	fs.BoolVar(&o.OCIInsecureSignaturePolicy, "oci-insecure-signature-policy", o.OCIInsecureSignaturePolicy, "If set, OCI catalog push will not try to push signatures")
 	fs.BoolVar(&o.SkipPruning, "skip-pruning", o.SkipPruning, "If set, will disable pruning globally")
+	fs.BoolVar(&o.SkipPruningMissingManifests, "skip-pruning-missing-manifests", o.SkipPruningMissingManifests, "If set, pruning will not fail if it's unable to find the manifest")
```
dmage, line 77: I would say it should be the default behavior. Why would one want the pruner to fail when it gets 404 on a delete request?

bcaton85: I would agree with this - @sherine-k @lmzuccarelli any objections?

lmzuccarelli: @bcaton85 - no objections

bcaton85: okay let me update

bcaton85: updated

sherine-k (commented):
/lgtm
Thanks for the PR! @bcaton85

sherine-k:
/retitle PROJQUAY-6345: Skipping mirror failure if manifest not found during prune

lmzuccarelli:
@bcaton85 - Thanks for this bug fix

PR #758, opened 2023-12-11
WIP: OCPBUGS-23158: Attempt a fix for operator filtering when only minVersion is set

sherine-k:
# Description

This is an attempt to make oc-mirror take `skips` into account when calculating the bundles to filter from an operator catalog. 
This PR fixes the specific case described in [OCPBUGS-23158](https://issues.redhat.com/browse/OCPBUGS-23158)

**:warning: IMPORTANT NOTE: this PR doesn't pretend to include `skips` in all cases of upgrade graph calculation for operators. It only does so for the case where channel filtering is applied, with minVersion/maxVersion.** 

## Type of change

# How Has This Been Tested?

Unit tests reproducing the use case from the issue are added.
MirrorToDisk with the imagesetconfig from the issue description passes. 

## Expected Outcome
Unit test and manual tests pass.

### DIFF
--- a/pkg/operator/diff/internal/diff_include.go
+++ b/pkg/operator/diff/internal/diff_include.go
@@ -389,6 +389,39 @@ func makeUpgradeGraph(ch *model.Channel) map[string][]*model.Bundle {
 	return graph
 }
 
+func getPredecessorOf(ch *model.Channel, bundleName string) *model.Bundle {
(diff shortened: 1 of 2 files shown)


sherine-k:
Closing the ticket OCPBUGS-23158 as a WONT DO, deferred to 4.16.

PR #765, opened 2023-12-14
OCPBUGS-21854: cincinnati-graph-data.tar.gz is created only for succe…

sherine-k:
…ssful API call

# Description

PARTIAL FIX - DONT MERGE
Fixes [OCPBUGS-21854](https://issues.redhat.com/browse/OCPBUGS-21854)

## Type of change

# How Has This Been Tested?

The cincinnati-graph-data.tar.gz gets created [here](https://github.com/openshift/oc-mirror/blob/90202e830bf7a243eb473506b79c1f067805af35/pkg/cli/mirror/create.go#L143-L145).
Using a debugger, I shut down the internet connection before the HTTP request is issued, and can see that the file cincinnati-graph-data.tar.gz is not created anymore under `cmd/oc-mirror/oc-mirror-workspace/src/cincinnati`

## Expected Outcome
Please describe the outcome expected from the tests

### DIFF
--- a/pkg/cli/mirror/cincinnati_graph_image.go
+++ b/pkg/cli/mirror/cincinnati_graph_image.go
@@ -151,16 +151,6 @@ func (o *MirrorOptions) buildGraphImage(ctx context.Context, srcSignatureDir str
 
 // downloadsGraphData will download the current Cincinnati graph data
 func downloadGraphData(ctx context.Context, dir string) error {
-	// TODO(jpower432): It would be helpful to validate
-	// the source of this downloaded file before processing
-	// it further
-	graphArchive := filepath.Join(dir, outputFile)
-	out, err := os.Create(filepath.Clean(graphArchive))
-	if err != nil {
-		return err
-	}
-	defer out.Close()
-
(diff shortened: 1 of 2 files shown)


sherine-k:
Hi @obockows 
Is it possible to test this fix for your client? 
PS: As explained in the Jira ticket, I'm not certain this will completely fix the issue. Empty file or no file will both cause an issue for building the graph image.
In my opinion, it's more important to understand the conditions for which the mirroring continues in spite of issues when downloading the graph data. Can you please assist us with this.
Thank you

obockows:
@sherine-k first of all sorry for lack of prompt answer and thank you for your effort. 
Of course, we can test the fix for the customer. You can prepare the binary and I will send it and they will try. 
Regarding your thought:
`In my opinion, it's more important to understand the conditions for which the mirroring continues in spite of issues when downloading the graph data. Can you please assist us with this.`
I agree, I am pretty sure there is some corner case that leads to such behavior. They were mirroring directly to their registry. I believe it would be worth it if you added more verbose output / additional help around the methods/functions that create the tar file - it would be just used for debugging purposes of course.

PR #778, opened 2024-01-12
CLID-14: Add study results in docs

sherine-k:
TLDR; The team will use a semver sorting to filter the bundles from the catalog. 

A soon to be expected introspection tool for catalogs [RFE-4930](https://issues.redhat.com//browse/RFE-4930) would be the corner stone towards a long term more accurate solution for filtering bundles.


docs/design/v2/operator-filtering-study.md
```diff
@@ -0,0 +1,139 @@
...
+
+### S1 - SemVer
+
+Most bundle versions are semver compatible. The bundles included in a channel can simply be sorted by version and any bundles that fall in the range selected in the ImageSetConfig are taken. 
+
+Such an algorithm will result in idempotent results compared to V1 results. Nevertheless, the upgrade graph information conveyed by the `replaces`, `skip` and `skipRange` is not taken into account. 
+This algorithm is totally different from the one used by OLM. There is no guarantee here that the selected list of bundles would allow OLM to perform an upgrade for the selected operator. There may also be a much shorter path between the minVersion and the maxVersion requested. oc-mirror will, in this case, mirror more bundles than what a cluster would need to perform the upgrade. 
+
+### S2 - Call `render-graph`
```
lmzuccarelli, line 40: I honestly thought that we were not going to have a dependency on opm render-graph - as per Jordan Keister's suggestion in our discussion with  the OLM team

lmzuccarelli: I'm just saying this as it adds unnecessary complexity

sherine-k: Agreed. In S2, we explain what the consequences would be if we were to call render-graph from oc-mirror, and explain why this is not a viable solution. 

Are you saying that S2 should not be presented and/or be presented same as S4?

PR #790, opened 2024-02-02
[release-4.14] OCPBUGS-28871: Capability to override default channel …

lmzuccarelli:
This fix address the issues found [here](https://issues.redhat.com/browse/OCPBUGS-385) .

The TL;DR is that dev teams can change the default channel from one release to the next, this causes customers who have set CICD pipelines to have to change their ImageSetConfig each time a version changes.
This fix allow the use of a field in the ImageSetConfig to override the default channel and specify any channel and use the HEAD (latest version of that channel).

It means that the customer needs to be aware of the caveats not using the default channel when mirroring. This will mean we need to update our documents accordingly. We will work with the docs team as to find the appropriate wording and track (link) the docs issue to this.

Fixes # [OCPBUGS-385](https://issues.redhat.com/browse/OCPBUGS-385)

## Type of change

# How Has This Been Tested?

This has been tested locally and seems to working as expected.
We will need thorough testing from QE to ensure this fix works.

## Expected Outcome
The ImageSetConfig used is as follows

mirror:
  operators:
  - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.12
    packages:
    - name: elasticsearch-operator
      overrideDefaultChannel: true
      channels:
      - name: stable

NB the overrideDefaultChannel flag is at the package level , also there can only be one channel specified.

Use the following cli to execute

bin/oc-mirror --config test-isc.yml file://<some-path>
The head of the current channel "stable" is version "elasticsearch-operator.v5.8.0"

To verify this

execute the command as shown above
update the ImageSetConfig to use the defaultChannel (HEAD is the same elasticsearch-operator.v5.8.0)
execute the oc command but ensure the contents are save in a different folder
compare folders
The elasticsearch-operator was the operator of interest in the listed JIRA, I also tested using the cluster-logging operator and also got the expected results (also expecting v5.8.0 version).

Maybe QE could look at a random set of operators to verify this on


.gitignore
```diff
@@ -17,6 +17,7 @@ v2/cmd/mirror/logs/
 v2/cmd/mirror/working-dir/
 v2/launch.json
 v2/logs/
+v2/test/results
```
sherine-k, line 20:
```suggestion
v2/tests/results
```

PR #800, opened 2024-02-20
CLID-27: Update to include oci (disk based layout) for v2

lmzuccarelli:
# Description

This updates the operator collector functionality to include file based oci index image layouts, utilizing the current bundle filtering.

## Type of change

# How Has This Been Tested?

Ensure an oci format layout is on disk

```
skopeo copy docker://registry.redhat.io/redhat/redhat-operator-index:v4.14 oci:redhat-index --remove-signatures
```

Execute oc-mirror using the following ImageSetConfig

```
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: oci://redhat-index
      packages:
      - name: aws-load-balancer-operator
```
Use the following cli

```
bin/oc-mirror --v2 --config clid-27.yaml file://clid-27
```

## Expected Outcome
The  mirrorToDisk functionality should pass

Untar the `clid-27/mirror_000001.tar` and verify that the albo as well as the redhat-index (catalog) is included

Second check was with setting the TargetTag

```
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: oci://redhat-index
      targetTag: "v4.14"
      packages:
      - name: aws-load-balancer-operator
```
Use the following cli

```
bin/oc-mirror --v2 --config clid-27.yaml file://clid-27
```

Third check was with setting the TargetCatalog

```
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: oci://redhat-index
      targetCatalog: "test-catalog:v4.14"
      packages:
      - name: aws-load-balancer-operator
```
Use the following cli

```
bin/oc-mirror --v2 --config clid-27.yaml file://clid-27
```

Verify the contents as in the previous step/s

As a final step verify the diskToMirror functionality

```
bin/oc-mirror --v2 --config clid-27.yaml --from file://clid-27 docker://localhost:5000/clid-27
```


v2/pkg/operator/local_stored_collector.go
```diff
@@ -220,17 +241,36 @@ func (o LocalStorageCollector) prepareM2DCopyBatch(log clog.PluggableLoggerInter
...
-			if err != nil {
-				o.Log.Error("%s", err.Error())
-				return nil, err
-			}
-			src = imgSpec.ReferenceWithTransport
-
-			if imgSpec.IsImageByDigest() {
-				dest = dockerProtocol + strings.Join([]string{o.LocalStorageFQDN, imgSpec.PathComponent + ":" + imgSpec.Digest[:hashTruncLen]}, "/")
+			if strings.HasPrefix(img.Image, ociProtocol) {
+				transportAndPath := strings.Split(img.Image, "://")
```
sherine-k, line 245: Same as above

v2/pkg/operator/local_stored_collector.go
```diff
@@ -220,17 +241,36 @@ func (o LocalStorageCollector) prepareM2DCopyBatch(log clog.PluggableLoggerInter
...
+				if len(transportAndPath) == 2 {
+					// generate a random hex string to use as a tag
+					var rnd = rand.New(rand.NewSource(time.Now().UnixNano()))
+					hex := strconv.FormatUint(rnd.Uint64(), 16)
+					var hexTag string
+					if len(hex) > hashTruncLen {
+						hexTag = hex[:hashTruncLen]
+					} else {
+						hexTag = hex
+					}
```
sherine-k, line 256: Wouldn't this produce a different tag each time we run MirrorToDisk? 
If the user is not using targetTag, they might even have different tags in the enclave registry for the same image.
We could use something similar to [this](https://github.com/openshift/oc-mirror/blob/main/v2/pkg/archive/image-blob-gatherer.go#L38-L60) in order to get the digest of the image, and deduce the tag from there. WDYT?
Such a function might have its place in the `image` package?

v2/pkg/operator/local_stored_collector.go
```diff
@@ -220,17 +241,36 @@ func (o LocalStorageCollector) prepareM2DCopyBatch(log clog.PluggableLoggerInter
...
+					dest = dockerProtocol + strings.Join([]string{o.LocalStorageFQDN, transportAndPath[1] + ":" + hexTag}, "/")
+				} else {
+					dest = dockerProtocol + strings.Join([]string{o.LocalStorageFQDN, img.Image}, "/")
```
sherine-k, line 259: Are we sure this is going to be accepted by the registry? ie no tag nor digest? 
Shouldn't we use the digest from the manifest instead here?
Maybe also (at least for oci), check if `targetTag` is provided: I bet the users of the oci feature would be looking for this.

v2/pkg/operator/local_stored_collector.go
```diff
@@ -59,24 +63,41 @@ func (o *LocalStorageCollector) OperatorImageCollector(ctx context.Context) ([]v
...
-			dest := ociProtocolTrimmed + dir
-			err = o.Mirror.Run(ctx, src, dest, "copy", &o.Opts, *writer)
-			writer.Flush()
-			if err != nil {
-				o.Log.Error(errMsg, err)
+
+		// CLID-27 ensure we pick up oci:// (on disk) catalogs
+		if strings.Contains(op.Catalog, ociProtocol) {
+			// truncate the prefix to get the directory
+			ociDir := strings.Split(op.Catalog, ociProtocol)
```
sherine-k, line 70: Suggestion: Here you can use imgSpec, err := image.ParseRef(img.Image) instead of HasPrefix.
imgSpec gives you the pathComponents as well as the transport.
Normally we performed unit tests to ensure it works for oci transport.

sherine-k (commented):
Hey @lmzuccarelli 
Nice work! I left a few suggestions. Let me know what you think

aguidirh (commented):
Could you please add the following line to your ImageSetConfig otherwise is going to fail:

`apiVersion: mirror.openshift.io/v1alpha2`

aguidirh (commented):
```
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: oci://redhat-index
      packages:
      - name: aws-load-balancer-operator
```

I have followed all the steps and mirrorToDisk worked fine, but when I tried to do diskToMirror using the tar generated by the previous step, I got the following: 

```
2024/02/20 14:47:11  [ERROR]  : [Worker] errArray Invalid destination name docker://docker://localhost:7000/redhat-index: invalid reference format
2024/02/20 14:47:11  [ERROR]  : [Worker] error in batch - refer to console logs
```

In my destination registry only the albo was created, redhat-operator-index is missing there:

```
{
  "repositories": [
    "albo/aws-load-balancer-controller-rhel8",
    "albo/aws-load-balancer-operator-bundle",
    "albo/aws-load-balancer-rhel8-operator",
    "openshift4/ose-kube-rbac-proxy"
  ]
}
```

aguidirh:
This is also a valid ImageSetConfig:

```
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: oci:///home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/redhat-index
      packages:
      - name: aws-load-balancer-operator
```

It failed during mirrorToDisk with the following:

```
2024/02/20 15:07:21  [ERROR]  : [Worker] errArray Invalid destination name docker://localhost:55000//home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/redhat-index:3ffabd67c14a: invalid reference format
2024/02/20 15:07:21  [ERROR]  : [Worker] error in batch - refer to console logs 
```

v2/pkg/operator/local_stored_collector.go
aguidirh: Same comment as @sherine-k here

v2/pkg/operator/local_stored_collector.go
```diff
@@ -59,24 +63,41 @@ func (o *LocalStorageCollector) OperatorImageCollector(ctx context.Context) ([]v
...
+			if len(ociDir) > 0 {
```
aguidirh, line 71: Since the code is checking the index [1] the len should be > 1 instead of zero. But I believe it is not needed this conditional in case image.ParseRef is used. Probably only a check if the field is not empty.

v2/pkg/operator/local_stored_collector.go
aguidirh: Yes, agree. Let's say the customer using this oci catalog wants to mirror only devworkspace-operator and one month later the customer decided to mirror also aws-load-balancer based on the same oci catalog on disk. This is going to generate two different tags even if the base oci catalog was the same one. 

I think it is safer to use the digest from the manifest of the oci catalog as suggested above and keep it in the image package.

v2/pkg/operator/local_stored_collector.go
aguidirh: I believe we should use the digest from the manifest on line 257 (instead of a random one) and here on line 259 we should try to find the tag (checking if the image contains a tag or if it is present in the targetTag field).

lmzuccarelli:
@aguidirh @sherine-k - So  I have gone about the tagging using annotations in the RelateImage.Name field
i.e as this example 

`"aws-load-balancer-rhel8-operator-16e9ffed36107527a37713ac5bd34a7bc20f042269a81077429fb5884914c4d0-annotation"`

Basically annotate if we have TargetTag set else find the manifest digest and use that. In the actual prepare for copy just check for the annotation and use it appropriately.

I think this is a simpler solution , I also have update the tests - checked that it covers 81% of the code

v2/pkg/operator/local_stored_collector.go
```diff
@@ -40,9 +42,11 @@ type LocalStorageCollector struct {
...
 
 	var (
-		allImages []v1alpha3.CopyImageSchema
-		label     string
-		dir       string
+		allImages      []v1alpha3.CopyImageSchema
+		label          string
+		dir            string
+		imageReference string
+		imageName      string
```
sherine-k, line 49: Is it probably safer to reset these 2 to empty at the beginning of the for loop?

PR #814, opened 2024-03-27
WIP: CLID-54: Mirror to mirror workflow

sherine-k:
# Description

This simply chains mirrorToDisk and diskToMirror


lmzuccarelli:
@sherine-k - nice I like this approach its simple and uses our existing flow, yes it will take a bit longer , but we have caching to assist and also if things go wrong we take "off" from the place of error on the next run

PR #828, opened 2024-04-11
CLID-85: Change string replacement in registry config to use templates

lmzuccarelli:
# Description

Update to change string replacement with text/template

## Type of change

# How Has This Been Tested?

Standard unit tests

## Expected Outcome

TestExecutorSetupLocalStorage - should pass

### DIFF
--- a/v2/pkg/cli/executor.go
+++ b/v2/pkg/cli/executor.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
 	"k8s.io/kubectl/pkg/util/templates"
@@ -430,8 +431,8 @@ func (o *ExecutorSchema) setupLocalStorage() error {
 version: 0.1
 log:
   accesslog:
-    disabled: $$PLACEHOLDER_ACCESS_LOG_OFF$$
-  level: $$PLACEHOLDER_LOG_LEVEL$$
+    disabled: {{ .LogAccessOff }}
+  level: {{ .LogLevel }}
   formatter: text
   fields:
     service: registry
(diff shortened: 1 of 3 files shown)


v2/pkg/cli/executor.go
```diff
@@ -450,27 +451,30 @@ http:
       #path: /etc/registry
 `
 
-	if _, err := os.Stat(o.LocalStorageDisk); err != nil {
-		// something went wrong
-		return fmt.Errorf("error using the local storage folder for caching")
-	}
```
sherine-k, line 456: this can be saved. it has value

v2/pkg/cli/executor.go
```diff
@@ -450,27 +451,30 @@ http:
...
-	configYamlV01 = strings.Replace(configYamlV01, "$$PLACEHOLDER_ROOT$$", o.LocalStorageDisk, 1)
-	configYamlV01 = strings.Replace(configYamlV01, "$$PLACEHOLDER_PORT$$", strconv.Itoa(int(o.Opts.Global.Port)), 1)
-	if o.Opts.Global.LogLevel != "trace" {
-		configYamlV01 = strings.Replace(configYamlV01, "$$PLACEHOLDER_LOG_LEVEL$$", o.Opts.Global.LogLevel, 1)
-	} else {
-		configYamlV01 = strings.Replace(configYamlV01, "$$PLACEHOLDER_LOG_LEVEL$$", "info", 1)
```
sherine-k, line 462: trace is not a log level understood , so you need some sort of translation for it. that is why, we cannot simply use: 
```
rc := RegistryConfig{
		LocalStorageDisk: o.LocalStorageDisk,
		LocalStoragePort: int(o.Opts.Global.Port),
		LogLevel:         o.Opts.Global.LogLevel,
		LogAccessOff:     o.Opts.Global.LogLevel != "debug",
	}
```

sherine-k (commented):
Hi @lmzuccarelli 
This is a much nicer approach!
 Beware some pitfalls that I highlighted here.

aguidirh (approved):
Thanks for this change @lmzuccarelli :tada: 

Using templates here was a wise choice.

PR #839, opened 2024-04-24
CLID-62,CLID-63: moves pkg dir to internal dir

aguidirh:
# Description

As oc-mirror is not a library, all the packages in v2 were moved under internal folder. In this way we prevent other go mod creating an undesired oc-mirror dependency.

## Type of change

# How Has This Been Tested?

The following tests were executed:

`All automated tests passing`

`Mirror To Disk`

`Disk To Mirror`

`Mirror To Mirror`

`Delete`

## Expected Outcome

All flows above should be executed without errors and the containers images should be pulled/pushed/deleted according to the flow executed.


lmzuccarelli:
Verified locally  using mirror to disk workflow, also double check  make build, make test etc

lmzuccarelli (approved):

PR #867, opened 2024-06-05
OCPBUGS-33121: Fix problems with versions and verbose flags

lmzuccarelli:
# Description

This fix addresses the issue faced in using the --v2 (used to invoke v2 of oc mirror) and accidentally using -v2 (used for verbose logging)

This fix will only be for 4.17 onwards until we deprecate v1

Fixes # OCPBUGS-33121

## Type of change

# How Has This Been Tested?

Locally and unit tests

## Expected Outcome
console message should show 

```
error: detected a v2 ImageSetConfiguration, please use --v2 instead of -v2
```

### DIFF
--- a/pkg/cli/mirror/mirror.go
+++ b/pkg/cli/mirror/mirror.go
@@ -280,6 +280,9 @@ func (o *MirrorOptions) Validate() error {
 	if mirrorToMirror || mirrorToDisk {
 		cfg, err := config.ReadConfig(o.ConfigPath)
 		if err != nil {
+			if strings.Contains(err.Error(), "config GVK not recognized") && o.LogLevel == 2 {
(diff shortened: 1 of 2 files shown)


aguidirh:
Usually before removing flags on GA version of oc-mirror like v1, it is necessary to deprecate them and let the customers know in the docs they need to remove the flag from their automatons / scripts. It gives them time to change their side.

Removing flags directly without deprecating could cause breaking automation which uses the flag removed.

Suggestion:

-Based on the apiVersion of the image set configuration (v2alpha1 for example) it would be possible to identify that the customer wants to use v2 and print a message `the correct flag is --v2 and not -v2` and stop the flow so the customer can fix the mistake.

PR #924, opened 2024-09-12
CLID-232,OCPBUGS-38339,OCPBUGS-38343,OCPBUGS-38233: Create release signature configmap

lmzuccarelli:
# Description

Add the release signature configmap for oc-mirror v2 (json format compatible with v1) to the cluster-resources folder of oc-mirror

This fix also addresses the bug fixes for OCPBUGS-38339 and OCPBUGS-38343

## Type of change

# How Has This Been Tested?

Use the following imagesetconfig 

```
kind: ImageSetConfiguration
mirror:
  platform:
    channels:
    - name: stable-4.16
      minVersion: 4.16.0
      maxVersion: 4.16.0
```

Execute oc-mirror mirror to disk workflow

```
bin/oc-mirror --config clid-232.yaml file://clid-232 --v2
```
Now execute the disk to mirror workflow 

```bash
bin/oc-mirror --config clid-232.yaml --from file://clid-232  docker://localhost:5000/clid-232 --v2 --tls-verify=false
```

## Expected Outcome
Check the working-dir/cluster-resources folder there should be a files with naming convention ``` signature-configmap.json and signature-configmap.yaml```  

To verify the contents of this field execute the following bash cli The ``` working-dir/signatures/ ```  folder has the same digest

```
cat working-dir/signatures/4.16.0-x86_64-sha256-3717338045df06e31effea46761b2c7e90f543cc4f00547af8158dd6aea868c3 | base64
```

The value output should match the contents of the config map (BinaryData section)

Verify the bug fixes 

OCPBUGS-38339

In the jira ticket follow the steps to re-produce the bug, there should be no panic but an error with appropriate message

OCPBUGS-38343

Once the tar file has been produced in the (mirror-to-disk) workflow, untar the contents and there should be a json file (with configmap structure in json) in the directory "working-dir/cluster-resources"

OCPBUGS-38233

Once the tar file has been produced in the (mirror-to-disk) workflow, untar the contents and there should be a yaml file (with configmap structure in yaml format) in the directory "working-dir/cluster-resources"


v2/internal/pkg/api/v2alpha1/type_configmap.go
sherine-k, line 1: Hi @lmzuccarelli
This is ConfigMap from the [kubernetes API](ttps://github.com/kubernetes/kubernetes/blob/856475e5fffe3d99c71606d6024f5ed93e37eebc/pkg/apis/core/types.go#L6123), right?
Suggestion to put this rather under v2/internal/pkg/api/kubernetes/core - a bit like we did with the catalogsource_types.go

v2/internal/pkg/release/signature.go
```diff
@@ -169,14 +170,40 @@ func (o SignatureSchema) GenerateReleaseSignatures(ctx context.Context, images [
...
+			cm.BinaryData[index] = b64
+			jsonData, err := json.Marshal(cm)
+			if err != nil {
+				o.Log.Error("%v", err)
+			}
+			// write to cluster-resources directory
+			ferr = os.WriteFile(o.Opts.Global.WorkingDir+clusterResourcesDir+index+".json", jsonData, 0644)
+			if ferr != nil {
+				o.Log.Error("%v", ferr)
+			}
```
sherine-k, line 206: Hi @lmzuccarelli,
I was wondering if this makes more sense in the `clusterresources` module, where we generate all the manifests (separation of concerns).
WDYT?

go.work.sum
sherine-k, line 1: Hi @lmzuccarelli
I think these dependencies are probably because of a previous attempt to include the kubernetes API module for configmap? 
It feels weird to have all these dependencies for such little changes in the signature.go imports

sherine-k (commented):
Hi @lmzuccarelli,
Thanks so much for the PR, I left a few comments, especially for go.work.sum.
Let' s discuss them together.

v2/internal/pkg/release/signature.go
```diff
@@ -169,14 +169,19 @@ func (o SignatureSchema) GenerateReleaseSignatures(ctx context.Context, images [
...
 				o.Log.Error("%v", ferr)
 			}
 			imgs = append(imgs, img)
+
+			cr := clusterresources.New(o.Log, o.Opts.Global.WorkingDir, o.Config, "")
+			err = cr.GenerateSignatureConfigMap(digest, id, data)
+			if err != nil {
+				o.Log.Error("%v", err)
+			}
+
```
aguidirh, line 184: Currently all the clusterresources are being generated in the executor.go, the generation of the SignaturesConfigMap could follow the same standard.

Here is an example: https://github.com/openshift/oc-mirror/blob/8d037bac34651ea92e52f46757854d1b53b8ebac/v2/internal/pkg/cli/executor.go#L781

Having the generation of the SignatureConfigMap in the executor could bring some benefits:

* The SignatureConfigMap will be generated only if the workflow finished successfully (in the collector it will be always generated)
* If the customer is running a dry run, having it in the executor.go will avoid the generation of the SignatureConfigMap (in the collector it will be always generated)
* Having the clusterresources being generated at only one place is easy to maintain the code in the long term
* Since the executor already has the dependency on the clusterresources pkg it is possible to avoid an additional dependency of the clusterresources pkg in the collector pkg
* Having the release collector calling the clusterresources (write results) pkg directly will break the architecture defined here: https://github.com/openshift/oc-mirror/blob/main/v2/assets/architecture.png where the executor should be the responsible to call the clusterresources (write results)

aguidirh (changes requested):

v2/internal/pkg/clusterresources/clusterresources.go
```diff
@@ -604,3 +607,61 @@ func toRFC1035(r rune) rune {
...
+			Namespace: signatureNamespace,
+			Labels: map[string]string{
+				signatureLabel: "",
+			},
+		},
+		BinaryData: make(map[string]string),
+	}
+	// read the signatures directory
+	sigDir := filepath.Join(o.WorkingDir, signatureDir)
+	signatures, err := os.ReadDir(sigDir)
```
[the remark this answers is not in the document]
lmzuccarelli: Yes - actually its simpler than that I''ll just add to the hashmap created for the BinaryData field . Good catch

kasturinarra:
Completed testing CLID-232 and found two issues, listing the same here:
---------------------------------------------------------------------------------------------------------
1. After d2m the folder path shown for signature configmap is missing a / see here `024/09/20 11:39:57  [INFO]   : CLID-232/working-dir/cluster-resourcessignature-configmap.json file created, 2024/09/20 11:39:57  [INFO]   : CLID-232/working-dir/cluster-resourcessignature-configmap.yaml file created` -> this is a bug and i will raise it separately
2. Binary data contains sha256-3717338045df06e31effea46761b2c7e90f543cc4f00547af8158dd6aea868c3-1: only after that it contains the actual content when run cat working-dir/signatures/4.16.0-x86_64-sha256-3717338045df06e31effea46761b2c7e90f543cc4f00547af8158dd6aea868c3 | base64 and base64 also contains a = at the end
cc: @lmzuccarelli

lmzuccarelli:
@kasturinarra - thanks for the feedback , I'll update and fix asap.

lmzuccarelli:
@kasturinarra - I have updated as requested and addressed the version-arch prefix problem

sherine-k (approved):
Thanks @lmzuccarelli

PR #936, opened 2024-10-10
CLID-247: changes the tlsVerify default to true and adds helm tls support

aguidirh:
# Description

This PR changes the tlsVerify default to true and adds helm tls support.

Fixes # ([CLID-247](https://issues.redhat.com/browse/CLID-247))

## Type of change

# How Has This Been Tested?

Run all the workflows (mirrorToDisk, diskToMirror and mirrorToMirror) for releases, operators, additional images and helm charts.

When doing the mirrorToMirror or diskToMirror `--dest-tls-verify=false` should be included in case the target registry does not have the TLS enabled

## Expected Outcome
All the flows (mirrorToDisk, diskToMirror and mirrorToMirror) should continue working for releases, operators, additional images, helm charts.

Also the delete should continue working (except for helm which is not implemented yet)

### DIFF
--- a/v2/internal/pkg/cli/executor.go
+++ b/v2/internal/pkg/cli/executor.go
@@ -41,7 +41,6 @@ import (
 	"github.com/openshift/oc-mirror/v2/internal/pkg/release"
 	"github.com/openshift/oc-mirror/v2/internal/pkg/version"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


v2/internal/pkg/cli/executor.go
```diff
@@ -359,13 +354,13 @@ func (o *ExecutorSchema) Complete(args []string) error {
 		o.Log.Debug("destination %s ", rootDir)
 		// destination is the local cache, which is HTTP
 		// nolint: errcheck
-		o.destFlagSet.Set("dest-tls-verify", "false")
+		o.Opts.DestImage.TlsVerify = false
```
lmzuccarelli, line 357: @aguidirh - much better, I was also looking at this for the rebuild catalogs stuff, I found it ugly to have to do a key/value set for a simple true false flag

PR #948, opened 2024-11-11
CLID-264: removes the default values from the struct definition

aguidirh:
# Description

This PR removes some unnecessary zero values definition on the BuildOptions struct.

Fixes [CLID-264](https://issues.redhat.com/browse/CLID-264)

## Type of change

# How Has This Been Tested?

MirrorToDisk and DiskToMirror

## Expected Outcome
Both workflows above should pass and the rebuilt catalog should be in the destination registry.

### DIFF
--- a/v2/internal/pkg/imagebuilder/rebuild_catalog.go
+++ b/v2/internal/pkg/imagebuilder/rebuild_catalog.go
@@ -117,9 +117,7 @@ RUN rm -fr /tmp/cache/* && /bin/opm serve /configs --cache-only --cache-dir=/tmp
 		return catalogCopyRefs, err
 	}
 
-	buildOptions.DefaultMountsFilePath = ""
-
-	o.Logger.Trace("containerfile %s", string(contents.Bytes()))
+	o.Logger.Trace("containerfile %s", contents.String())
 
 	buildStoreOptions, err := storage.DefaultStoreOptions()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


v2/internal/pkg/imagebuilder/rebuild_catalog.go
```diff
@@ -153,16 +151,12 @@ RUN rm -fr /tmp/cache/* && /bin/opm serve /configs --cache-only --cache-dir=/tmp
...
-		SignBy:                 "",
-		ManifestType:           "application/vnd.oci.image.manifest.v1+json",
-		AddCompression:         []string{},
-		ForceCompressionFormat: false,
-		MaxRetries:             &retries,
+		Store:              buildStore,
+		SystemContext:      destSysContext,
+		ImageListSelection: cp.CopyAllImages,
+		RemoveSignatures:   true,
+		ManifestType:       buildah.OCIv1ImageManifest, //TODO currently red hat catalog is docker v2 format, converting it to oci is not going to cause problems?
```
sherine-k, line 158: I don't think it will... This is already what v1 does when rebuilding catalogs: upon pulling, it is converted to oci

sherine-k (approved):

v2/internal/pkg/imagebuilder/rebuild_catalog.go
aguidirh: Great, so probably it will not be a problem. Thanks @sherine-k.

PR #1047, opened 2025-01-28
CLID-300: Update catalog filtering doc

sherine-k:
This is a simple doc update. Takes into account the fact that filtering by bundle selection has been removed


aguidirh (approved):
Thanks for updating this file @sherine-k

PR #1072, opened 2025-02-11
OCPBUGS-50503: oc-mirror fails with error `unknown image: reference n…

lmzuccarelli:
…ame is empty`

# Description

This fix addresses the issue when using digest based release images in the ImageSetConfig.

Github / Jira issue: [OCPBUGS-50503](https://issues.redhat.com/browse/OCPBUGS-50503)

## Type of change

# How Has This Been Tested?

Tested locally using this ImageSetConfig

```
kind: ImageSetConfiguration
mirror:
  platform:
    graph: true
    release: quay.io/openshift-release-dev/ocp-release@sha256:e0907823bc8989b02bb1bd55d5f08262dd0e4846173e792c14e7684fbd476c0d

```

Execute a mirror-to-mirror 

```
bin/oc-mirror --config release-digest.yaml --workspace file://release-test docker://localhost:5000/release-test --v2 --dest-tls-verify=false
```

Executed m2d and d2m 

## Expected Outcome

In all flows oc-mirror should execute without any errros

### DIFF
--- a/v2/internal/pkg/release/local_stored_collector.go
+++ b/v2/internal/pkg/release/local_stored_collector.go
@@ -181,6 +181,7 @@ func (o *LocalStorageCollector) ReleaseImageCollector(ctx context.Context) ([]v2
 		}
 
 		for _, releaseImg := range releaseImages {
+
 			releaseRef, err := image.ParseRef(releaseImg.Image)
 			if err != nil {
 				return []v2alpha1.CopyImageSchema{}, fmt.Errorf(errMsg, err.Error())
@@ -409,7 +410,7 @@ func (o *LocalStorageCollector) ReleaseImage(ctx context.Context) (string, error
 	if len(o.Releases) == 0 {
 		releaseImages, _, err := o.identifyReleases(ctx)
 		if err != nil {
-			return "", fmt.Errorf("[release collector] could not establish the destination for the release image: %v", err)
+			return "", fmt.Errorf("[release collector] could not find release images (from disk cache): %v", err)
 		}
 		o.Releases = []string{}
 		for _, img := range releaseImages {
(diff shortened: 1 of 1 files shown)


v2/internal/pkg/release/local_stored_collector.go
```diff
@@ -424,7 +425,18 @@ func (o *LocalStorageCollector) ReleaseImage(ctx context.Context) (string, error
...
+		// OCPBUGS-50503
+		// This could be a digest also
+		imgSpec, err := image.ParseRef(o.Releases[0])
+		if err != nil {
+			return "", fmt.Errorf("[release collector] could not parse release image %s", o.Releases[0])
+		}
+		releaseTag := ""
+		if imgSpec.IsImageByDigestOnly() {
+			releaseTag = imgSpec.Digest
+		} else {
```
aguidirh, line 437: Having the digest value in the releaseTag would impact the first two switch cases in `prepareTag` func (line 566). 

It would be better to add only tag values to releaseTag var and let the `prepareTag` to handle the scenario where the image is by digest only.

v2/internal/pkg/release/local_stored_collector.go
```diff
@@ -424,7 +425,18 @@ func (o *LocalStorageCollector) ReleaseImage(ctx context.Context) (string, error
...
-		releaseTag := o.Releases[0][:strings.LastIndex(o.Releases[0], ":")]
+		// OCPBUGS-50503
+		// This could be a digest also
+		imgSpec, err := image.ParseRef(o.Releases[0])
+		if err != nil {
+			return "", fmt.Errorf("[release collector] could not parse release image %s", o.Releases[0])
+		}
+		releaseTag := ""
+		if imgSpec.IsImageByDigestOnly() {
+			releaseTag = imgSpec.Digest
```
sherine-k, line 436: If we really want the releaseTag, we simply need to recover the second returned value from identifyReleases.
This gives us the releaseFolder from which we can read the `image-references` file, which contains the tag. 

More important is that this fix only works if we are calling this method `ReleaseImage` to generate the UpdateService. Because the UpdateService only wants the URL (without digest nor tag) to add in the updateService.yaml.

A comment at the beginning of this method could be important so that we don' t use this method for anything else.

The reason why the digest cannot be used in releaseTag for other scenarios is because we are not mirroring the release image with a digest as the tag! we are mirroring using sha256-tag. so any cluster trying to pull the release image by a tag that resembles the digest would fail to do so

lmzuccarelli: @sherine-k Alex and I agreed that this code is repeated in 2 places (i.e the call to ReleaseImage) in executor jus t called when graph is set to true, we need  seriously to refactor the "spaghetti code" in the executor  and elsewhere.

PR #1078, opened 2025-02-14
no-jira: Disable go workspace and build v2 as a binary

r4f4:
# Description

* We shouldn't be importing the v2 module into the v1 (logically wrong). Having them as separate modules will allow us to bump their dependencies independently, enabling faster pace development in the v2 code.
* The distribution/distribution/v3 library needs to be pinned to a version before the `ocischema.DeserializedImageIndex` type as introduced. That is because `oc` and `library-go` cannot currently handle that type and nor do we know when it will.

Due to way go workspace works, we could not downgrade the distribution version we used in v1 because we need a more recent version for v2 (needed by containers/image). Removing the workspace allows us to have distinct dependency versions thus making it possible to have distinct development cadences for v1 and v2.

To make it possible for users to still invoke --v1 or --v2 from the same binary, we use the `//go:embed` directive to embed the v2 binary into the v1 oc-mirror, extract and execute the binary at runtime. It incurs a small time penalty but it has the same affect as importing the v2 module into v1 without the downsides.

Github / Jira issue: OCPBUGS-44103

## Type of change

# How Has This Been Tested?

```yaml
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.16
      packages:
        - name: openstack-operator
          channels:
            - name: stable-v1.0
              minVersion: 1.0.4
              maxVersion: 1.0.4
```
```bash
$ ./bin/oc-mirror --config isc.yaml file://ocpbugs-44103
```

## Expected Outcome

No `error: the manifest type *ocischema.DeserializedImageIndex is not supported`


cmd/oc-mirror/main.go
```diff
@@ -15,3 +39,47 @@ func checkErr(err error) {
 		kcmdutil.CheckErr(err)
 	}
 }
+
+func isV2() bool {
+	return len(os.Args) > 0 && slices.Contains(os.Args[:], "--v2")
+}
+
+func runOcMirrorV2Cmd(args []string) error {
```
lmzuccarelli, line 47: @r4f4 - This is clever,  it also ensures as you mentioned separating v1 and v2 go.mods

r4f4:
Update: fixed a typo when reading the embed file and fixed an issue in the `--help` output by giving the full args slice to the v2 process otherwise the program name will be missing:
```
Usage:
  --v2
-c <image set configuration path> [--from | --workspace] <destination prefix>:<destination location> --v2
 [flags]
  --v2 [command]
```

cmd/oc-mirror/main.go
```diff
@@ -1,11 +1,36 @@
...
 	kcmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
+//go:embed data/*
+var mirrorv2 embed.FS
+
 func main() {
+	if isV2() {
+		if err := runOcMirrorV2Cmd(os.Args); err != nil {
+			klog.Errorf("%s", err.Error())
```
r4f4, line 26: Need to experiment with this a bit more but I think we just want to print the error if it's not an `exec.ExitError`.

v2/Makefile
```diff
@@ -1,14 +1,51 @@
...
+
+GO_BUILD_FLAGS = -tags "json1 $(BTRFS_BUILD_TAG) $(LIBDM_BUILD_TAG) $(LIBSUBID_BUILD_TAG)"
+
 all: clean test build
 
 build: 
-	mkdir -p build
-	go build -o build ./... 
+	mkdir -p $(GO_BUILD_BINDIR)
+	go build $(GO_MOD_FLAGS) $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) -race -o $(GO_BUILD_BINDIR) ./...
```
sherine-k, line 48: hi @r4f4 
does this mean we will have 2 binaries for oc-mirror? 
In this case, should we also be adding these binaries in the various Dockerfiles for delivery?

cmd/oc-mirror/main.go
```diff
@@ -15,3 +40,48 @@ func checkErr(err error) {
...
+			// Kill children if parent is dead
+			Pdeathsig: syscall.SIGKILL,
+			Setpgid:   true,
+		},
+		Args:   args,
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	}
+
+	return cmd.Run()
```
sherine-k, line 86: Hi @r4f4 
this is an interesting way to call the v2. 
I'm not at all an expert... I wonder how this impacts the way oc-mirror would run in the various situations:
* as a standalone binary
* as an oc plugin
* inside a pod / pipeline

maybe not at all...

v2/Makefile
r4f4: No, it's still just one binary in the end.
We could consider 2 distinct released binaries in the future so that `oc-mirror` is v2 and users need to use `oc-mirror-v1` for the deprecated functionality.

cmd/oc-mirror/main.go
r4f4: There should be no impact I know of. If there is, it's probably a bug in my implementation.

r4f4:
/retitle disable go workspace and pin distribution/distribution/v3

I'm removing the reference to the bug since it's being fixed by https://github.com/openshift/oc-mirror/pull/1080

aguidirh:
@r4f4 one question I have is about the FIPS. 

Today we have rhel8 and rhel9 binaries only because we needed the FIPS compliance in rhel9. 

Do you think that even embeding one binary inside the other they (probably the build team) will be able to still embed the FIPS feature in the binary? 

I do not know how they enable the FIPS on the binary, so that's why I decided to ask this question.

r4f4:
> @r4f4 one question I have is about the FIPS.
> 
> Today we have rhel8 and rhel9 binaries only because we needed the FIPS compliance in rhel9.
> 
> Do you think that even embeding one binary inside the other they (probably the build team) will be able to still embed the FIPS feature in the binary?
> 
> I do not know how they enable the FIPS on the binary, so that's why I decided to ask this question.

As long as the v2 binary respects all the flags set during build time, it should be fine. It would be like building 2 fips-capable binaries and then concatenating them into a single binary. Both will be dynamically linked against the system/rhel openssl libs. The difference is that the v2 is extracted during runtime.

aguidirh:
Got it, thanks for the explanation @r4f4.

cmd/oc-mirror/main.go
sherine-k: I was also thinking about impacts on the debugger (dev only).
When we previously tried to use buildah as the solution to build catalogs, there was also an exec.Cmd involved, and debuggings were harder (but not impossible) to do

r4f4: > I was also thinking about impacts on the debugger (dev only).
> When we previously tried to use buildah as the solution to build catalogs, there was also an exec.Cmd involved, and debuggings were harder (but not impossible) to do

Then you'll be happy to know this has no impact! For local dev/debug you can run the V2 binary directly. The run-v2-from-v1 is just a convenience for users, devs don't have to go through the wrapper.

sherine-k:
Hey, I'm all for giving this a last round of approval from QE and adopt this for 4.19.
@nidangavali , @kasturinarra WDYT?
If we could do some basic v1 and v2 tests on various rhel8/rhel9/fips_compliant machines, it could provide us with the data we need.

kasturinarra:
Can some one help explain what the bug is about and what are we trying to achieve ?

r4f4:
> Can some one help explain what the bug is about and what are we trying to achieve ?

Initially this was an attempt to solve the `DeserializedImageIndex` issue by separating the dependencies of v1 and v2; in that case, we could use an older `distribution/distribution` version for v1 which didn't trigger the issue but a newer one for v2.

Even though the `DeserializedImageIndex` issue has been resolved by a different approach, having v1 and v2 share the same version of deps when v1 is being deprecated is going to hold us back sooner or later. So the idea in this PR is still valid for the future.

The idea here is that we generate 2 separate binaries, one for v1 and another for v2. For ease of transitioning, the published `oc-mirror` binary will be the v1 with an embedded v2. Nothing should change for users.

r4f4:
Update: rebased on top of master to fix merge conflicts.

r4f4:
/retitle Disable go workspace and build v2 as a binary

Update: rebased and resolved merge conflicts.

lmzuccarelli:
Let's wait for @aguidirh  to add approve

r4f4:
Update: rebased to fix merge conflicts.

aguidirh:
Hi @lmzuccarelli,

I just added the LGTM, could you please add the missing labels in order to get this PR merged?

lmzuccarelli:
/retitle no-jira: Disable go workspace and build v2 as a binary

PR #1086, opened 2025-02-19
CLID-321: Add annotations to resources generated by oc-mirror v2

sherine-k:
# Description

This PR helps add an annotation to all cluster resources generated by oc-mirror.
Right now the annotation is `author: oc-mirror` , but let's discuss what is the right annotation to use. 

This way, on a cluster, when users need to remove/update/view IDMS/ITMS/CatalogSource generated by oc-mirror they could use:
```bash
oc get idms jsonpath='{.items[?(@.metadata.annotations.author=oc-mirror)].metadata.name}'
```

Github / Jira issue: [CLID-321](https://issues.redhat.com/browse/CLID-321)

## Type of change

# How Has This Been Tested?

I tested this only for additionalImages.
```yaml
kind: ImageSetConfiguration
mirror:
  additionalImages: 
  - name: registry.redhat.io/ubi9/ubi-micro:latest
```

## Expected Outcome
```yaml
---
apiVersion: config.openshift.io/v1
kind: ImageTagMirrorSet
metadata:
  annotations:
    author: oc-mirror
  name: itms-generic-0
spec:
  imageTagMirrors:
  - mirrors:
    - sherinefedora:5000/ubi9
    source: registry.redhat.io/ubi9
status: {}
```

### DIFF
--- a/v2/internal/pkg/clusterresources/clusterresources.go
+++ b/v2/internal/pkg/clusterresources/clusterresources.go
@@ -6,10 +6,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 	"unicode"
 
 	confv1 "github.com/openshift/api/config/v1"
@@ -22,6 +24,7 @@ import (
 	"github.com/openshift/oc-mirror/v2/internal/pkg/image"
 	clog "github.com/openshift/oc-mirror/v2/internal/pkg/log"
 	"github.com/openshift/oc-mirror/v2/internal/pkg/parser"
+	"github.com/openshift/oc-mirror/v2/internal/pkg/version"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -129,7 +132,8 @@ func (o *ClusterResourcesGenerator) generateITMS(mirrorsByCategory []categorized
 				Kind:       "ImageTagMirrorSet",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "itms-" + catMirrors.category.toString() + "-0",
+				Name:        "itms-" + catMirrors.category.toString() + "-0",
+				Annotations: generateOcMirrorAnnotations(),
 			},
 			Spec: confv1.ImageTagMirrorSetSpec{
 				ImageTagMirrors: []confv1.ImageTagMirrors{},
@@ -294,6 +298,10 @@ func (o *ClusterResourcesGenerator) generateCatalogSource(catalogRef string, cat
 			generateWithoutTemplate = true
 			o.Log.Error("error generating catalog source from template. Fall back to generating catalog source without template: %v", err)
 		}
+		if obj.ObjectMeta.Annotations == nil {
+			obj.ObjectMeta.Annotations = map[string]string{}
+		}
+		maps.Copy(obj.ObjectMeta.Annotations, generateOcMirrorAnnotations())
 	}
 	if generateWithoutTemplate || catalogSourceTemplateFile == "" {
 		obj = ofv1alpha1.CatalogSource{
@@ -302,8 +310,9 @@ func (o *ClusterResourcesGenerator) generateCatalogSource(catalogRef string, cat
 				Kind:       "CatalogSource",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      catalogSourceName,
-				Namespace: "openshift-marketplace",
+				Name:        catalogSourceName,
+				Namespace:   "openshift-marketplace",
+				Annotations: generateOcMirrorAnnotations(),
 			},
 			Spec: ofv1alpha1.CatalogSourceSpec{
 				SourceType: "grpc",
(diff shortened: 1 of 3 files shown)


v2/internal/pkg/clusterresources/clusterresources.go
```diff
@@ -127,6 +127,9 @@ func (o *ClusterResourcesGenerator) generateITMS(mirrorsByCategory []categorized
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "itms-" + catMirrors.category.toString() + "-0",
+				Annotations: map[string]string{
```
aguidirh, line 130: Thanks for adding this annotation. It will be easier to find cluster resources generated by oc-mirror in this way.

As we discussed, do you think it could be benificial adding also the oc-mirror version (v1 or v2) ?

aguidirh (commented):
Thanks for adding this annotation Sherine.

Only one comment from my side. 

Overall LGTM

v2/internal/pkg/clusterresources/clusterresources.go
sherine-k: Yes! 
Actually, this PR just shows the how.
I think we need to discuss what annotations to add, key names, etc.
v1 / v2 is definitely an idea, oc-mirror exact version too, date/time of generation...

sherine-k: @aguidirh @lmzuccarelli @r4f4 @kasturinarra 
Do you think of any annotations we could add?

lmzuccarelli: agree with @aguidirh

kasturinarra: I think version and oc-mirror v1 or v2 should be good i believe, if customers or users requests more later we could further look into it.

aguidirh:
Hi @sherine-k,

As we discussed above, is it possible to add the oc-mirror version to the annotation?

sherine-k:
Hi @aguidirh 

> As we discussed above, is it possible to add the oc-mirror version to the annotation?

I had a bit of a struggle with the annotation key name, and maybe your suggestions will help here.
In `createdBy`, right now the value is `oc-mirror --v2` (see [here for ex](https://github.com/openshift/oc-mirror/blob/8dfb3ac03271bd59ecefeb24c0bca550bf51bdad/v2/internal/pkg/clusterresources/clusterresources.go#L136))

I'm not an absolute fan of this either... My goal was to avoid confusions:
* `version` as an annotation name might be confused with the version of the resource, so I avoided it completely
* `oc-mirror_version` was also added to have the exact version of the binary used to generate this
* so I was struggling to find another annotation name for `v1`/`v2`

Any ideas?

aguidirh:
Hi @sherine-k,

It is fine to add the version together in the same annotation. Sorry I reviewed in my IDE and I forgot to pull the latest commit which contained the version.

Is there any advantage of keeping `--v2` instead of only `v2` ?

sherine-k:
> Is there any advantage of keeping --v2 instead of only v2 ?

None... I just thought it would match the flag... 
We could also use `oc-mirror v2`, `oc-mirror_v2`, `oc-mirror-v2`?
I'll try to look at other examples from other components for ideas...

v2/internal/pkg/clusterresources/clusterresources.go
```diff
@@ -130,6 +132,11 @@ func (o *ClusterResourcesGenerator) generateITMS(mirrorsByCategory []categorized
...
+					"createdBy":         "oc-mirror --v2",
+					"createdAt":         time.Now().UTC().Format(time.RFC850),
+					"oc-mirror_version": version.Get().GitVersion,
```
r4f4, line 138: This is not going to return the proper git version atm because we're not setting it during build:
```fish
❯ ./bin/oc-mirror --v2 version

2025/03/11 10:27:23  [INFO]   : 👋 Hello, welcome to oc-mirror
2025/03/11 10:27:23  [INFO]   : ⚙️  setting up the environment for you...
WARNING: This version information is deprecated and will be replaced with the output from --short. Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"", Minor:"", GitVersion:"v0.0.0-unknown", GitCommit:"", GitTreeState:"", BuildDate:"", GoVersion:"go1.23.6", Compiler:"gc", Platform:"linux/amd64"}
```
I'll fix this in a separate bug.

v2/internal/pkg/clusterresources/clusterresources.go
```diff
@@ -304,6 +317,11 @@ func (o *ClusterResourcesGenerator) generateCatalogSource(catalogRef string, cat
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      catalogSourceName,
 				Namespace: "openshift-marketplace",
+				Annotations: map[string]string{
+					"createdBy":         "oc-mirror --v2",
+					"createdAt":         time.Now().UTC().Format(time.RFC850),
+					"oc-mirror_version": version.Get().GitVersion,
+				},
```
r4f4, line 324: Should we move this to an utility function since it repeats quite a bit?

v2/internal/pkg/clusterresources/clusterresources.go
```diff
@@ -832,3 +845,11 @@ func (o *ClusterResourcesGenerator) GenerateSignatureConfigMap(allRelatedImages
 
 	return nil
 }
+
+func generateOcMirrorAnnotations() map[string]string {
+	return map[string]string{
+		"createdBy":         "oc-mirror v2",
+		"createdAt":         time.Now().UTC().Format(time.RFC850),
+		"oc-mirror_version": version.Get().GitVersion,
```
r4f4, line 853: The version issue should be fixed by https://github.com/openshift/oc-mirror/pull/1109

sherine-k:
@aguidirh @r4f4 @lmzuccarelli 
This PR is ready and code is modified to accomodate suggestions you made.
Thanks

PR #1155, opened 2025-05-20
WRKLDS-1592: Use internal implementation for auth.APIVersions

tchap:
# Description

This repository uses `github.com/distribution/distribution/v3/registry/client/auth` package, which is no longer public. Breaking a public contract should not happen, but it did. This issue would prevent future dependency upgrade when needed, so it must be handled eventually.

The only function actually being used is [auth.APIVersions](https://github.com/distribution/distribution/blob/a44f1fb058533f3f840037ebf29c0fea377ebbc6/internal/client/auth/api_version.go#L26-L38), which is rather simple, so the chosen solution is to reimplement the function in `pkg/metadata/storage/registry.go` as `apiVersions`, hence removing the dependency.

Github / Jira issue: 

https://issues.redhat.com/browse/WRKLDS-1592

## Type of change

# How Has This Been Tested?

I added unit tests, which might feel as a bit extra work, then run `make test`.
Let me know in case they are not needed or there is a better way to test this, thanks.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require golang.org/x/sys v0.31.0 // indirect
 require (
 	github.com/containerd/errdefs v0.3.0
 	github.com/distribution/distribution/v3 v3.0.0-beta.1
+	github.com/google/go-cmp v0.7.0
 	github.com/mitchellh/hashstructure/v2 v2.0.2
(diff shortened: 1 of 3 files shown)


tchap:
I am going to actually close this as the changes for supporting the current distribution/v3 and library-go are underway and  `auth.APIVersions` will be included in `library-go` it seems.

PR #1161, opened 2025-05-26
OCPBUGS-56953: Allow custom values for helm charts to be provided

siegy22:
# Description

Some helm charts don't render without providing values. Which forces one to use additionalImages a lot.
Also a handy feature to reduce the additionalImages list, by setting the custom image (flavour) directly in the helm chart.

One example would be OpenBao, that provides ubi based images, but doesn't use them by default.

Before:

```yaml
kind: ImageSetConfiguration
mirror:
  # ...
  additionalImages:
    - quay.io/openbao/openbao-ubi:2.2.0
  helm:
    repositories:
      - name: bao
        url: https://openbao.github.io/openbao-helm
        charts:
          - name: openbao
            version: 0.12.0 # You'd have to manually check the matching image
                            # when bumping the chart's version
```

After:

```yaml
kind: ImageSetConfiguration
mirror:
  # ...
  helm:
    repositories:
      - name: bao
        url: https://openbao.github.io/openbao-helm
        charts:
          - name: openbao
            version: 0.12.0
            values:
              server:
                image:
                  repository: "openbao/openbao-ubi"
```

## Type of change

- [x] This change requires a documentation update on openshift docs (https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/disconnected_environments/mirroring-in-disconnected-environments#about-installing-oc-mirror-v2)

# How Has This Been Tested?

- Added a testcase for backwards compatibility
- Added a testcase for the new functionality
- Tested locally with the example above

## Expected Outcome
In the manual test the `openbao/openbao-ubi:2.2.0` should be included in the mirrored tar file or in the target registry

### DIFF
--- a/docs/design/imageset-configuration.md
+++ b/docs/design/imageset-configuration.md
@@ -90,7 +90,6 @@ OKD is supported by the mirroring process but not the discovery `list` tools.
 ## Helm Charts
 
 - Private repositories are currently not supported.
-- Helm charts that require alterations to the `values.yaml` to render are not currently supported.
 
 
 
--- a/docs/examples/imageset-config-helm.yaml
+++ b/docs/examples/imageset-config-helm.yaml
@@ -14,6 +14,14 @@ mirror:
         charts:
           - name: podinfo
             version: 5.0.0
+            # Provide values if needed. Image overrides are respected
+            # and will result in mirroring ghcr.io/example/podinfo:6.9.0 instead
+            # of the default ghcr.io/stefanprodan/podinfo:5.0.0
+            # values:
+            #   image:
+            #     repository: ghcr.io/example/podinfo
+            #     tag: 6.9.0
+
     #local:
     #  - name: podinfo-local
     #    path: /home/user/podinfo-6.0.0.tgz


siegy22:
You're absolutely right that values can vary significantly between charts, and subcharts or more complex configurations may benefit from separate values files. That said, for the primary issue this PR addresses — some Helm charts not rendering unless a minimal set of values is provided (especially for image overrides) — inline values seem to be the most practical short-term solution.

In most real-world use cases we’ve seen so far, only a few keys (like image.repository or tag) need to be overridden to get a chart to render correctly and include the appropriate images in the mirror. Using a full external values.yaml might be overkill in those situations and would make the ImageSetConfiguration file more fragmented and harder to read at a glance.

This feature mainly helps reduce the need to maintain large additionalImages lists and avoid brittle image guessing — making it easier to work with charts like OpenBao that offer alternative images but don’t default to them.

We could certainly explore support for file-based values overrides later if there's a stronger demand for complex configurations or subchart handling. But for now, inline values offer a good balance of usability and functionality.

siegy22:
What do you mean it depends on the helm API? The whole cli depends on the helm API on how to render charts. The only thing that's changed is that we allow to pass custom values to it.

And I also don't see the need for multiple values files because you as the user are in control of the values file? Or why would you want to chain multiple files for a single chart? I guess to be more DRY it would be a thing? But this could also be added separately in a later change to allow to pass files, instead of inling them by maybe adding a `values_file(s)` option to the chart in the ImagesetConfiguration.

PR #1162, opened 2025-05-27
OCPBUGS-55076: adds code to handle signature errors in the archive pkg

aguidirh:
# Description

When running `m2d` with `--remove-signatures=false`, failing to add a release signature (except the graph image which does not have a signature) to the archive should stop the workflow.

Github / Jira issue: [OCPBUGS-55076](https://issues.redhat.com/browse/OCPBUGS-55076)

## Type of change

# How Has This Been Tested?

Developed unit tests for each scenario.

## Expected Outcome
Unit tests should pass.

### DIFF
--- a/v2/internal/pkg/archive/image-blob-gatherer.go
+++ b/v2/internal/pkg/archive/image-blob-gatherer.go
@@ -122,7 +122,8 @@ func multiArchBlobs(ctx context.Context, in internalImageBlobGatherer) (map[stri
 
 		singleArchBlobs, err := singleArchBlobs(ctx, singleIn)
 		if err != nil {
-			if errors.As(err, &SignatureBlobGathererError{}) {
+			var sigErr *SignatureBlobGathererError
+			if errors.As(err, &sigErr) {
 				sigErrors = append(sigErrors, err)
 			} else {
 				return nil, err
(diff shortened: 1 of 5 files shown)


v2/internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -184,27 +185,27 @@ func imageSignatureBlobs(ctx context.Context, in internalImageBlobGatherer) ([]s
...
 	if ref, err = image.ParseRef(in.imgRef); err != nil {
-		return nil, SignatureBlobGathererError{SigError: err}
+		return nil, &SignatureBlobGathererError{SigError: err}
 	}
 	ref = ref.SetTag(tag)
 
 	manifestBytes, mime, err := imageManifest(ctx, in.sourceCtx, ref.ReferenceWithTransport, nil)
 	if err != nil {
-		return nil, SignatureBlobGathererError{SigError: err}
+		return nil, &SignatureBlobGathererError{SigError: err}
```
sherine-k, line 198: Hi @aguidirh ,

I would really distinguish this error (and maybe the ones before it) from the following ones, and have `handleSignatureErrors` take into account the difference.

This specific error means: the registry doesn't have that tag (HTTP 404). 
The errors that follow (L 203 and 208) mean that the image has a signature but we're unable to read its digest or get its blobs, so really an error.

Taking examples: 

- If I have images from a helm chart, which are not signed, `imageManifest()` is going to be in error, but it's ok to skip that error, because most probably policies on the cluster are in place to consume that image without signature verification. 
- On the other hand, if I have an operator bundle image, which has a signature tag, but that for some reason we're unable to get the `sigBlobs` , I'm not sure we should skip such an error... What if cluster image policies expect to verify image signature for operator images, and that deployment fails?

WDYT @aguidirh ?

aguidirh: The initial request was to stop the workflow only in case oc-mirror v2 was not able to add release signatures to the archive (tar), any other signature image type like operators, helm or additional images should continue.

Since we are very close to 4.19 GA and we would like to have all the bugs related to the signature merged, I would  suggest to handle this kind of scenario in some kind of exit code feature for the archive pkg (there is no exit code feature on this pkg right now).

What do you think @lmzuccarelli ?

lmzuccarelli: @aguidirh - yes I agree with you Alex, lets go with what we have and then add a bugfix (if this becomes urgent) for z stream release in 4.19 or a feature to add in 4.20

PR #1164, opened 2025-05-27
OCPBUGS-56486: v2/config: allow `delete` and `mirror` words

r4f4:
# Description

We should be able to accept the `delete` and `mirror` words in the configuration as long as they are not part of the config root keys. This should fix the issue with images making the config be rejected because they contain one of those keywords in their name.

Github / Jira issue:  OCPBUGS-56486

## Type of change

# How Has This Been Tested?

With the following ISC:
```yaml
kind: ImageSetConfiguration
mirror:
  additionalImages:
   - name: delete.delete/delete/delete/delete1:delete
```

## Expected Outcome

Config is not rejected just because it contains the words `mirror` or `delete`.

### DIFF
--- a/v2/internal/pkg/config/load.go
+++ b/v2/internal/pkg/config/load.go
@@ -3,83 +3,89 @@ package config
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


aguidirh (approved):
/approve

PR #636, opened 2023-05-16
OCPBUGS-13332: Create rfc 1035 compliant catalog source name

jchunkins:
# Description

Create catalog source metadata.name values that are compliant with RFC 1035. The generated name is also suitable for naming the catalog source file as well.

Fixes # OCPBUGS-13332

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] run oc mirror

1. Copy single architecture catalog in OCI format
    ```
    skopeo --override-os linux copy --format v2s2 docker://icr.io/cpopen/ibm-zcon-zosconnect-catalog@sha256:6f02ecef46020bcd21bdd24a01f435023d5fc3943972ef0d9769d5276e178e76 oci:///tmp/611/oci-index
    ```
1. Create ISC in /tmp/611/oci-index/isc.yaml
    ```yaml
    kind: ImageSetConfiguration
    apiVersion: mirror.openshift.io/v1alpha2
    storageConfig:
      local:
        path: /tmp/localstorage
    mirror:
      operators:
      - catalog: oci:///tmp/611/oci-index
    ```
1. change current working directory to  "/tmp/611"
1. run oc mirror with destination as a local registry
    ```
    oc mirror --config /tmp/611/oci-index/isc.yaml docker://localhost:5000/single-oci --dest-skip-tls --dest-use-http --include-local-oci-catalogs
    ```
1. RESULT:
    - note that resulting name complies with RFC 1035 and is of the form `cs-<generated name>[-0]` where the generated name varies depending on the input provided in the ISC. The name starts with `cs-` to ensure that the name starts with a alpha character, and an optional suffix of `-0` is used only if the last character of the generated name ends with a non alphanumeric character. This `-0` suffix should rarely be necessary, but is provided in order to be compliant with RFC 1035. Note that the name is used in the catalog source metadata.name and as part of the file name.
        - Example file name 
          `/tmp/611/oc-mirror-workspace/results-1684276766/catalogSource-cs-611-oci-index.yaml`
        - Example catalog source:
          ```apiVersion: operators.coreos.com/v1alpha1
          kind: CatalogSource
          metadata:
[...]

PR #1174, opened 2025-06-04
CLID-370: clusterresources: remove `cs-` prefix

r4f4:
# Description

Users are complaining that the `cs-` prefix breaks automations.

Github / Jira issue: [CLID-370](https://issues.redhat.com//browse/CLID-370)

## Type of change

# How Has This Been Tested?

m2d + d2m
```
2025/06/05 10:23:50  [INFO]   : 📄 Generating CatalogSource file...
2025/06/05 10:23:50  [INFO]   : data/cs-prefix/working-dir/cluster-resources/redhat-operator-index-v4-14.yaml file created
```

## Expected Outcome

Catalog source without the `cs-` prefix

### DIFF
--- a/v2/internal/pkg/clusterresources/clusterresources.go
+++ b/v2/internal/pkg/clusterresources/clusterresources.go
@@ -282,7 +282,7 @@ func (o *ClusterResourcesGenerator) generateCatalogSource(catalogRef string, cat
 
 	pathComponents := strings.Split(catalogSpec.PathComponent, "/")
 	catalogRepository := pathComponents[len(pathComponents)-1]
-	catalogSourceName := "cs-" + catalogRepository + "-" + csSuffix
+	catalogSourceName := catalogRepository + "-" + csSuffix
 	// maybe needs some updating (i.e other unwanted characters !@# etc )
 	catalogSourceName = strings.Replace(catalogSourceName, ".", "-", -1)
 	errs := validation.IsDNS1035Label(catalogSourceName)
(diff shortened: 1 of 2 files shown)


sherine-k:
> If I remember correctly, this `cs-` prefix was a customer request on [CLID-10](https://issues.redhat.com/browse/CLID-10) / [CFE-824](https://issues.redhat.com/browse/CFE-824). But I do not remember the reasons behind that.
> 
> @sherine-k do you remember the details about adding this prefix to the catalog source name? I would like to make sure we are not going to break the customer who did this request.

Yes, @aguidirh , I found a bit of context here:
The cs- was introduced as a fix for [OCPBUGS-13332](https://issues.redhat.com/browse/OCPBUGS-13332), and PR #636 explains that the catalog source custom resource name should not start by a numeric character. 

[OCPSTRAT-1842](https://issues.redhat.com/browse/OCPSTRAT-1842) doesn't ask for removing the cs- prefix, because this ensure we have a valid custom resource name. Instead, it asks for the possibility to specify another name for the custom resource (following the naming conventions).

r4f4:
> > The RFE/customer only asks for the removal of the `cs-` prefix. The custom name is our own invention.
> 
> There's a very interesting thread of discussions about whether to remove the prefix or to add an option in [RFE-5555](https://issues.redhat.com/browse/RFE-5555). I wasn't involved in the discussions, but if I understand correctly what I read, they seem to be in favor of adding an option in the ISC.

I don't see the point on adding a config option to ISC if the file can be easily renamed after generation.

aguidirh:
My concern is that only removing the `cs-` probably would make [OCPBUGS-13332](https://issues.redhat.com/browse/OCPBUGS-13332) to come back. 

The PR #636 by @jchunkins in oc-mirror v1 fixed the issue in a different way, should we follow the same strategy? Or should we fix it in a different way in oc-mirror v2?

Issue #1030, opened 2025-01-20
Add support for mirroring Helm charts with both operator and operand images


pgodowski:
As of today, `oc-mirror` supports mirroring the container images referenced from Helm Charts, as the result of the Helm support initial work: https://github.com/openshift/oc-mirror/issues/83

However, when the Helm Chart packages the operator-managed workloads, `oc-mirror` is mirroring only the operator-controller image, but not the operand images. The reason for it is that `oc-mirror` does the image reference looking within the rendered Helm YAML manifests, assuming that some of those YAMLs would represent `Deployment`, `Pod`, `StatefulSet`, etc.

In operator-managed workloads, what gets initially deployed via Helm chart though is the operator-controller Deployment/StatefulSet, but the operand is deployed by operator-controller dynamically, after operator's CustomResource reconcilation.

This enhancement request is to provide some convention for Helm chart authors to follow by listing the operand images and then `oc-mirror` to get the list of such operand images for mirroring purposes.

There are some existing community efforts in this space:
- [helm-images](https://github.com/nikhilsbhat/helm-images/)  - which can extract image reference from both workload-controller manifests (`Deployment`, etc), but also from `ConfigMap`
- [helm-mirror](https://github.com/openSUSE/helm-mirror/tree/master) - which extract image reference from within any manifest file, as long as it matches the specific regular expression (like `image:`).

There are some suggestions for implementation choices:

- use ConfigMap which has the list of images
- use of ImageSetConfigration packages within Helm chart, which has the `additionalImages` section populated with the list of operand images
- use the `images.txt` file which was supposed to be a Helm convention, but never materialised, see https://github.com/helm/helm/issues/7754

pgodowski:
Explored this a little bit more and came up with two tactical approaches which can be done with the existing `oc-mirror v2` implementation - those are obviously stop-gap options, yet some of the concepts there could be taken into consideration in the actual design of the final solution.

## Tactical option 1 - additionalImages in ImageSetConfiguration

```yaml
apiVersion: mirror.openshift.io/v1alpha2
kind: ImageSetConfiguration
...
additionalImages:
- name: quay.io/acme/operator-image:1.0
- name: quay.io/acme/operand-primary-image:1.0
- name: quay.io/acme/operand-secondary-image:1.0
```

## Tactical option 2 - additional CR in Helm, listing images

[...]

PR #1178, opened 2025-06-10
CLID-372: Enhance Helm Chart Support for operator and operand images

aguidirh:
# Description

This PR enhance the oc-mirror v2 to discover and mirror container images referenced within environment variables of deployment templates. It skips the environment variables which does not contain a valid container image on its value.

Github / Jira issue: 

- [OCPSTRAT-1967](https://issues.redhat.com/browse/OCPSTRAT-1967)
- [CLID-371](https://issues.redhat.com/browse/CLID-371)
- [CLID-372](https://issues.redhat.com/browse/CLID-372)
- #1030 

## Type of change

# How Has This Been Tested?

With the following Image Set Configuration (please change the path of test-mirror-helm-0.3.0.tgz):

```
kind: ImageSetConfiguration
mirror:
  helm:
    local:
     - name: test-mirror-helm
       path: /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/test-mirror-helm-0.3.0.tgz
       imagePaths:
        - "{.spec.template.spec.containers[*].env[*].value}"
```
Run the following `m2d` command:
```
./bin/oc-mirror -c ./alex-tests/alex-isc/clid-372.yaml file://alex-tests/clid-372 --v2
```

## Expected Outcome

This is the expected output (SUCCESS) running the steps above with the code of this PR:
```
2025/06/10 18:06:44  [INFO]   : 👋 Hello, welcome to oc-mirror
2025/06/10 18:06:44  [INFO]   : ⚙️  setting up the environment for you...
2025/06/10 18:06:44  [INFO]   : 🔀 workflow mode: mirrorToDisk 
2025/06/10 18:06:44  [INFO]   : 🕵  going to discover the necessary images...
2025/06/10 18:06:44  [INFO]   : 🔍 collecting release images...
2025/06/10 18:06:44  [INFO]   : 🔍 collecting operator images...
2025/06/10 18:06:44  [INFO]   : 🔍 collecting additional images...
2025/06/10 18:06:44  [INFO]   : 🔍 collecting helm images...
2025/06/10 18:06:44  [INFO]   : 🔂 rebuilding catalogs
2025/06/10 18:06:44  [INFO]   : 🚀 Start copying the images...
2025/06/10 18:06:44  [INFO]   : 📌 images to copy 2 
 ✓   (40s)  prometheus:latest ➡️  cache  
2 / 2 (49s) [============================================================================================================================================================================================================================================================================================================================================================] 100 %
 ✓   (49s)  nginx-ingress:latest ➡️  cache  
2025/06/10 18:07:34  [INFO]   : === Results ===
2025/06/10 18:07:34  [INFO]   :  ✓  2 / 2 helm images mirrored successfully
2025/06/10 18:07:34  [INFO]   : 📦 Preparing the tarball archive...
2025/06/10 18:07:35  [INFO]   : mirror time     : 51.12146711s
2025/06/10 18:07:35  [INFO]   : 👋 Goodbye, thank you for using oc-mirror
```

While with the previous code (main branch), it fails:

```
./bin/oc-mirror -c ./alex-tests/alex-isc/clid-372.yaml file://alex-tests/clid-372 --v2
2025/06/10 12:29:47  [INFO]   : 👋 Hello, welcome to oc-mirror
2025/06/10 12:29:47  [INFO]   : ⚙️  setting up the environment for you...
2025/06/10 12:29:47  [INFO]   : 🔀 workflow mode: mirrorToDisk 
2025/06/10 12:29:47  [INFO]   : 🕵  going to discover the necessary images...
2025/06/10 12:29:47  [INFO]   : 🔍 collecting release images...
2025/06/10 12:29:47  [INFO]   : 🔍 collecting operator images...
2025/06/10 12:29:47  [INFO]   : 🔍 collecting additional images...
2025/06/10 12:29:47  [INFO]   : 🔍 collecting helm images...
2025/06/10 12:29:47  [ERROR]  : debug unable to parse image correctly : tag and digest are empty
2025/06/10 12:29:47  [ERROR]  : [HelmImageCollector] debug unable to parse image correctly : tag and digest are empty
2025/06/10 12:29:47  [INFO]   : mirror time     : 4.49555ms
2025/06/10 12:29:47  [INFO]   : 👋 Goodbye, thank you for using oc-mirror
2025/06/10 12:29:47  [ERROR]  : [Executor] collection error: debug unable to parse image correctly : tag and digest are empty 
```


v2/internal/pkg/helm/local_stored_collector_test.go
```diff
@@ -539,6 +540,32 @@ func TestHelmImageCollector(t *testing.T) {
 			},
 			expectedError: nil,
 		},
+		{
+			caseName:     "local helm chart with images on deployment env vars- MirrorToDisk: should pass",
+			mirrorMode:   mirror.MirrorToDisk,
+			localStorage: testLocalStorageFQDN,
+			helmConfig: v2alpha1.Helm{
+				Local: []v2alpha1.Chart{
+					{Name: "test-mirror-helm", Path: filepath.Join(testChartsDataPath, "test-mirror-helm-0.3.0.tgz"), ImagePaths: []string{"{.spec.template.spec.containers[*].env[*].value}"}},
```
pgodowski, line 549: I am OK to include this helm chart provided earlier, as part of the test articacts

v2/internal/pkg/helm/local_stored_collector.go
```diff
@@ -447,6 +448,12 @@ func findImages(templateData []byte, paths ...string) (images []v2alpha1.Related
 		}
 
 		for _, result := range results {
+			_, err := image.ParseRef(result)
```
pgodowski, line 451: I am wondering whether this would introduce side effects of trying to mirror non-existing image references which are mentioned as value of the environment variable. Wondering whether perhaps we might be doing some additional checking of the matched jsonpath expressions, to see whether e.g. the env variable name matches some convention, before trying to extract image reference from the value?

I do understand that jsonpath library does not allow for regexp, but could we at the golang code check not only values under of the jsonpath, but also the actual keys matched and then do some additional filtering, e.g. include only matched paths for env variables with specific suffix like `_IMAGE`?

Daniel-Fan: I am wondering that if we could add additional "rules"/"filter" function after `results, err := parseJSONPath(data, j, path)`, to further filter out the image list based on ENV variable naming convention or any given rules?

For example, if the `results` of jsonpath is not a image list `["image1", "image2", "image3"]`, instead is a complex structure like following
```json
[
  { "name": "FOO_IMAGE", "value": "icr.io/foo:v1" },
  { "name": "BAR_TAG", "value": "icr.io/bar:v2" },
  { "name": "BAZ_IMAGE", "value": "icr.io/baz:v3" }
]
```
because user specifies `- "{.spec.template.spec.containers[*].env}"` in the `imagePaths` field.

We could interpret `results` again to obtain the exact image list we prefer, e.g.,
```golang
results, err := parseJSONPath(data, j, path)
if err != nil {
	return nil, err
}

# check if slice's item is a string image reference or not
# if not, doing further filter to get the image reference based on naming convention
if len(results) != 0 && results[0].(type) != string {
   filteredResults, err = getImageEnvValuesWithNamingConvention(results)
} else {
   filteredResults = results
}

for _, result := range results {
	_, err := image.ParseRef(filteredResults)
}
```

aguidirh: > I am wondering that if we could add additional "rules"/"filter" function after results, err := parseJSONPath(data, j, path), to further filter out the image list based on ENV variable naming convention or any given rules?

I am not sure if I understood the requests above. In line 451 introduced on this PR will skip everything that is not a container image, so from my perspective there is no need of ENV variable naming convention since the check is already happening.

> I do understand that jsonpath library does not allow for regexp, but could we at the golang code check not only values under of the jsonpath, but also the actual keys matched and then do some additional filtering, e.g. include only matched paths for env variables with specific suffix like _IMAGE?

The jsonpath does not allow suffix like _IMAGE because it [does not accept regular expressions](https://kubernetes.io/docs/reference/kubectl/jsonpath/#regular-expressions-in-jsonpath), but it accepts to add the exact container image name as below:

```
    local:
     - name: test-mirror-helm
       path: /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/test-mirror-helm-0.3.0.tgz
       imagePaths:
        - "{.spec.template.spec.containers[*].env[?(@.name == 'DUMMY_IMAGE1')].value}"
        - "{.spec.template.spec.containers[*].env[?(@.name == 'DUMMY_IMAGE2')].value}"
        - "{.spec.template.spec.containers[*].env[?(@.name == 'DUMMY_IMAGE3')].value}"
```

Adding suffixes like _IMAGE will make [L445](https://github.com/openshift/oc-mirror/pull/1178/files#diff-acdfd74c21cb313383b324f096c98a64d41c5c112b71c90660bf1e5a278694b2R445) fail, which won't allow even to retrieve the image names from the env var values.

aguidirh: @pgodowski were you able to test this PR to see if it works for your use case?

pgodowski: Tested and it works. I raised a concern that if the customer uses though this approach, we might gave extranous container image pulls. let's hypthotherically say that someone has env variable like:
`MY_WEBSITE`: `example.com/path/page` - this would end up with an attempt to do the image pull, since the path looks-llike container image reference.

aguidirh: Hi @pgodowski,

If you try the example you gave (`MY_WEBSITE: example.com/path/page`), you will see that the [L451](https://github.com/openshift/oc-mirror/pull/1178/files#diff-acdfd74c21cb313383b324f096c98a64d41c5c112b71c90660bf1e5a278694b2R451) `_, err := image.ParseRef(result)` will return an error like `example.com/path/page unable to parse image correctly : tag and digest are empty`.

This is because image.ParseRef will only parse valid container image references, otherwise it will return an error.

Daniel-Fan: Hi @aguidirh 

> I am not sure if I understood the requests above. In line 451 introduced on this PR will skip everything that is not a container image, so from my perspective there is no need of ENV variable naming convention since the check is already happening.

Thank you for the clarification!
I believe that the main concern might be how robust the `_, err := image.ParseRef(result)` is, to [identify if the `result`/`results` is an actual image that we would like to mirror](https://github.com/openshift/oc-mirror/blob/6211d4b8d192e683ae32d0b4d65cb9193908db00/v2/internal/pkg/image/image.go#L51).
- If it is robust enough, it would be sufficient to only parse image reference based on `result`/`results`.
- If we would like to have some additional "guards" besides parsing `result`/`result`, maybe we could consider adding some requirements, e.g., checking `name: MY_WEBSITE` when it is colocated with `value: example.com/path/page`
  - I know that it is complicated since jsonpath does not accept regular expression, so it has to be something implemented additionally in the code.

aguidirh: Hi @Daniel-Fan,

There are some [tests](https://github.com/openshift/oc-mirror/blob/6211d4b8d192e683ae32d0b4d65cb9193908db00/v2/internal/pkg/image/image_test.go#L32) for image.ParseRef to guarantee it only accepts valid container image paths. 

As I mentioned in my [previous](https://github.com/openshift/oc-mirror/pull/1178#discussion_r2140698770) comment the `value: example.com/path/page` would throw an error like `example.com/path/page unable to parse image correctly : tag and digest are empty` since it is not a valid container image path.

aguidirh: Feel free to play with image.ParseRef [tests](https://github.com/openshift/oc-mirror/blob/6211d4b8d192e683ae32d0b4d65cb9193908db00/v2/internal/pkg/image/image_test.go#L32), if you find scenarios where it fails please let me know, in this way we can make this func more robust in a separate PR.

pgodowski: ok, that's fine. lets have the initial implementation done and then we can refine subsequently

pgodowski (approved):

PR #1181, opened 2025-06-16
OCPBUGS-58186: v2/mirror: retry in case of ContextDeadlineExceeded

r4f4:
# Description

This change makes use of a custom error check function to extend the checks made by containers/common on whether to retry a copy operation or not.

Github / Jira issue: OCPBUGS-57362

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

## Expected Outcome
`--retry-times` settings is respected in case of 50x errors and `ContextDeadlineExceeded`

### DIFF
--- a/v2/internal/pkg/mirror/mirror.go
+++ b/v2/internal/pkg/mirror/mirror.go
@@ -2,13 +2,16 @@ package mirror
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"strings"
 
 	"github.com/containers/common/pkg/retry"
 	"github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/pkg/cli"
 	"github.com/containers/image/v5/signature"
@@ -198,9 +201,14 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions) 
 		co.ReportWriter = opts.Stdout
 	}
 
-	return retry.IfNecessary(ctx, func() error {
+	var retryOpts retry.Options
+	if opts.RetryOpts != nil {
+		retryOpts = *opts.RetryOpts
+	}
+	retryOpts.IsErrorRetryable = isErrorRetryable
 
-		//manifestBytes, err := copy.Image(ctx, policyContext, destRef, srcRef, &copy.Options{
+	//nolint:wrapcheck // context will be added by the calling function
+	return retry.IfNecessary(ctx, func() error {
 		manifestBytes, err := o.mc.CopyImage(ctx, policyContext, destRef, srcRef, co)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


v2/internal/pkg/mirror/mirror.go
```diff
@@ -215,7 +223,30 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions)
...
-	}, opts.RetryOpts)
+	}, &retryOpts)
+}
+
+// Custom implementation to extend `containers/common/pkg/retry.retry`
+func isErrorRetryable(err error) bool {
+	var httpError docker.UnexpectedHTTPStatusError
+	switch {
+	case err == nil:
+		return false
```
lmzuccarelli, line 234: @r4f4 - would it be worth while adding a warning message - indicating to the user that re-try is occurring.  ?

r4f4: I think warning is too much. What about just a debug message? Printing things while the spinners are working messes up the output.

r4f4: Another option is to log the opposite: print a message when we're _not_ retrying because of the specific error.

lmzuccarelli: @r4f4 - yes agreed about printing while the spinners are working,  so yes maybe the suggestion you made when were not retrying,  really my thinking was "retry set to x" , trying y of x , trying y+1 of x" etc

r4f4: The problem is that the retry mechanism happens inside `containers/common` via `retry.IfNecessary` so we don't really have that information unless we reimplement the retry on our side. Notice that `containers/common` already prints log information about retries if `logrus` is properly configured: https://github.com/containers/common/blob/main/pkg/retry/retry.go#L50

v2/internal/pkg/mirror/mirror.go
```diff
@@ -198,9 +201,14 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions)
 		co.ReportWriter = opts.Stdout
 	}
 
-	return retry.IfNecessary(ctx, func() error {
+	var retryOpts retry.RetryOptions
```
aguidirh:
```suggestion
	var retryOpts retry.Options
```

nit: `retry.RetryOptions` is deprecated on `containers/common` in favor of `retry.Options`

Holding this PR in case you wan to change it.

/hold

aguidirh (approved):
Thanks for fixing this issue, it will help a lot of customer since this is a common recurring issue.

Added only one nit suggestion.

v2/internal/pkg/mirror/mirror.go
r4f4: @lmzuccarelli so the problem here is that both `containers/*` and `dist/dist` use the standard `logrus` logger and we set that output to the `registry.log` file. If you look into that file in case of copy errors, you'll see:
```
❯ rg 'retrying' data/log-level/working-dir/logs/registry.log 
1097:time="2025-06-20T15:34:50.265837936+02:00" level=warning msg="Failed, retrying in 1s ... (1/2). Error: initializing source docker://quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:0800280100d6e3376a4c4afe98d724a032e7a05aff5b49dbaef0651c51d3f357: Get \"https://quay.io/v2/auth?account=<redacted>&scope=repository%3Aopenshift-release-dev%2Focp-v4.0-art-dev%3Apull&service=quay.io\": net/http: TLS handshake timeout"
```
None of those libs offer a way to use a custom logger instance instead of the standard logger; so besides reimplementing `retry.IfNecessary` the only alternative is to write a logrus Hook and have the hook write to different outputs depending on the fields set in the logger (e.g, `dist/dist` seems to set the `service=registry` field).

lmzuccarelli: @r4f4 - the juice is not worth the squeeze,  lets just leave it as is then. As you mentioned at least there is an entry in the registry.log

r4f4:
/retitle OCPBUGS-58186: v2/mirror: retry in case of ContextDeadlineExceeded

PR #1182, opened 2025-06-17
OCPBUGS-55374: fix to inform user of bad umask

lmzuccarelli:
# Description

Fix to inform user of bad umask

Github / Jira issue:  OPCBUGS-55374

## Type of change

# How Has This Been Tested?

Set umask 

```
umask 0077
```

## Expected Outcome

Execute oc mirror
Should inform client of bad umask setting (Error message)

### DIFF
--- a/v2/internal/pkg/cli/executor.go
+++ b/v2/internal/pkg/cli/executor.go
@@ -17,6 +17,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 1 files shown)


v2/internal/pkg/cli/executor.go
```diff
@@ -191,6 +192,13 @@ func NewMirrorCmd(log clog.PluggableLoggerInterface) *cobra.Command {
 			log.Info(emoji.WavingHandSign + " Hello, welcome to oc-mirror")
 			log.Info(emoji.Gear + "  setting up the environment for you...")
 
+			// OCPBUGS-55374 (check current umask)
+			currentUmask := syscall.Umask(0)
+			// 18 represents 22 in octal
+			if currentUmask != 18 {
```
r4f4, line 198: What about other umask values other than 022 that still allow oc-mirror to read its own files?

r4f4: The e2e tests failure is an example of that.

r4f4: Also don't you change the current umask to 0 by querying it? You have to restore the previous value.

lmzuccarelli: Other umask values (other than 0022 will not be considered ) i.e. from oc-mirror docs 

```
Prerequisites

You have installed the OpenShift CLI (oc). If you are mirroring image sets in a fully disconnected environment, ensure the following:

-You have installed the oc-mirror plugin on the host that has internet access.
-The host in the disconnected environment has access to the target mirror registry.
-You have set the umask parameter to 0022 on the operating system that uses oc-mirror.
-You have installed the correct binary for the RHEL version that you are using.

```

r4f4: Then this is not a bug and I don't think we need this check.

r4f4:
> @r4f4 - hmm yes e2e locally works , however running in PROW is another issue. I agree with you , I'll add a comment in OCPBUGS, saying that it is documented and close this PR

Maybe your system is configured with a default umask of `022` whereas in Prow it's some other value?
Even if documented, failing if umask is not set to `022` will cause `oc-mirror` to error out in systems/automations where the umask is more permissive than that, thus introducing a regression of sorts.

PR #1202, opened 2025-07-01
OCPBUGS-57370: v2/defaults Fix rate limiting issues by enabling exponential backoff retries

aguidirh:
# Description

This PR fixes rate limiting issues that occur during mirror operations when oc-mirror sends too many requests to container registries, resulting in "too many requests" errors.

The change sets the default `--retry-delay` to 0 seconds to force the containers/image library to use its built-in exponential backoff retry strategy instead of fixed 1-second delays. This provides smarter retry timing that adapts to rate limiting responses from container registries.

This change works in conjunction with:
- Increased `--retry-times` for more retry attempts
- Reduced `maxParallelLayerDownloads` and `maxParallelImageDownloads` to limit concurrent requests

Github / Jira issue: [OCPBUGS-57370](https://issues.redhat.com/browse/OCPBUGS-57370)

## Type of change

# How Has This Been Tested?

Since this is not an issue that is always happening, it is not easy to test it manually. 

It is possible to see less images are being copied during the worker phase because of the reduced `maxParallelImageDownloads`

## Expected Outcome

It is expected that customers are going to face the issue with `too many request` less often with this bug fix.

### DIFF
--- a/v2/internal/pkg/mirror/options.go
+++ b/v2/internal/pkg/mirror/options.go
@@ -273,7 +273,7 @@ func RetryFlags() (pflag.FlagSet, *retry.Options) {
 	opts := retry.Options{}
 	fs := pflag.FlagSet{}
 	fs.IntVar(&opts.MaxRetry, "retry-times", 5, "the number of times to possibly retry")
-	fs.DurationVar(&opts.Delay, "retry-delay", 0, "delay between 2 retries")
+	fs.DurationVar(&opts.Delay, "retry-delay", 0, "delay between 2 retries (default 0 uses exponential backoff, set value uses constant delay)")
 	return fs, &opts
 }
 
--- a/v2/internal/pkg/mirror/options.go
+++ b/v2/internal/pkg/mirror/options.go
@@ -272,8 +272,8 @@ func ImageSrcFlags(global *GlobalOptions, shared *SharedImageOptions, deprecated
 func RetryFlags() (pflag.FlagSet, *retry.Options) {
 	opts := retry.Options{}
 	fs := pflag.FlagSet{}
-	fs.IntVar(&opts.MaxRetry, "retry-times", 2, "the number of times to possibly retry")
-	fs.DurationVar(&opts.Delay, "retry-delay", time.Second, "delay between 2 retries")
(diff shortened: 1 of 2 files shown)


v2/internal/pkg/mirror/options.go
```diff
@@ -272,8 +272,8 @@ func ImageSrcFlags(global *GlobalOptions, shared *SharedImageOptions, deprecated
 func RetryFlags() (pflag.FlagSet, *retry.Options) {
 	opts := retry.Options{}
 	fs := pflag.FlagSet{}
-	fs.IntVar(&opts.MaxRetry, "retry-times", 2, "the number of times to possibly retry")
-	fs.DurationVar(&opts.Delay, "retry-delay", time.Second, "delay between 2 retries")
+	fs.IntVar(&opts.MaxRetry, "retry-times", 5, "the number of times to possibly retry")
+	fs.DurationVar(&opts.Delay, "retry-delay", 0, "delay between 2 retries")
```
lmzuccarelli: @aguidirh -  I'm not sure how setting this 2 zero does an exponential backoff

r4f4: Do we want to add to the help text about using an exp backoff if this value is zero but a constant delay otherwise?

aguidirh: Because if set to zero the containers/common will handle it internally as exp backoff.

See [here](https://github.com/containers/common/blob/c156ac84320ee1e3f6baa1a2e7d637b464525e5a/pkg/retry/retry.go#L47).

aguidirh: It is a good idea, I will change the message to explain that.

aguidirh: Hi @r4f4 

I changed the help text to the following:

```
--retry-delay duration           delay between 2 retries (default 0 uses exponential backoff, set value uses constant delay)
```

Let me know if you would like to change to something else.

lmzuccarelli: @aguidirh - message looks good to me

r4f4: LGTM

PR #1200, opened 2025-06-27
OCPBUGS-58193: v2/archive: error d2m when no tar files found

r4f4:
# Description

If we don't error out right away, the error message returned gives no clue about the actual reason for the failure:
```
2025/06/27 10:21:45  [ERROR]  : [Executor] collection error: collect catalog "registry.redhat.io/redhat/redhat-operator-index:v4.18": get manifest: error when creating a new image source: reading manifest v4.18 in localhost:55000/redhat/redhat-operator-index: manifest unknown
```

Github / Jira issue: OCPBUGS-58193

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code Improvements (Refactoring, Performance, CI upgrades, etc)
- [ ] Internal repo assets (diagrams / docs on github repo)
- [ ] This change requires a documentation update on openshift docs

# How Has This Been Tested?

```fish
❯ ~/prjs/github.com/openshift/oc-mirror/bin/oc-mirror --v2 --config catalog-isc.yaml --from file://data/catalog-filter-fail docker://localhost:5000/data --dest-tls-verify=false --log-level=debug
2025/06/27 14:31:30  [INFO]   : 👋 Hello, welcome to oc-mirror
2025/06/27 14:31:30  [INFO]   : ⚙️  setting up the environment for you...
2025/06/27 14:31:30  [INFO]   : 🔀 workflow mode: diskToMirror 
2025/06/27 14:31:30  [DEBUG]  : helm.New opts.SrcImage.TlsVerify false
Error: no tar archives matching "mirror_[0-9]{6}\\.tar" found in data/catalog-filter-fail
[...]
```

## Expected Outcome

Error message about tar files not found.

PR #1211, opened 2025-07-08
OCPBUGS-58467: Add range check for --parallel-images and --parallel-l…

lmzuccarelli:
…ayers flags

# Description

Add range checks

Github / Jira issue:  OCPBUGS-48467

## Type of change

# How Has This Been Tested?

Tested locally with flag --parallel-images 0 and 11 and --parallel-layers 0 and 11

## Expected Outcome
Should fail if out of range

### DIFF
--- a/v2/internal/pkg/cli/executor_test.go
+++ b/v2/internal/pkg/cli/executor_test.go
@@ -468,92 +468,13 @@ func TestExecutorValidate(t *testing.T) {
 		opts.Global.WorkingDir = "file://test"
 		assert.NoError(t, ex.Validate([]string{"docker://test"}))
 		log.AssertNotCalled(t, "Warn", mock.Anything)
-	})
-
-	t.Run("Testing Executor : validate should pass", func(t *testing.T) {
-		log := clog.New("trace")
-
-		global := &mirror.GlobalOptions{
-			SecurePolicy: false,
-		}
-
-		_, sharedOpts := mirror.SharedImageFlags()
-		_, deprecatedTLSVerifyOpt := mirror.DeprecatedTLSVerifyFlags()
-		_, srcOpts := mirror.ImageSrcFlags(global, sharedOpts, deprecatedTLSVerifyOpt, "src-", "screds")
-		_, destOpts := mirror.ImageDestFlags(global, sharedOpts, deprecatedTLSVerifyOpt, "dest-", "dcreds")
-		_, retryOpts := mirror.RetryFlags()
-
-		opts := &mirror.CopyOptions{
-			Global:              global,
-			DeprecatedTLSVerify: deprecatedTLSVerifyOpt,
-			SrcImage:            srcOpts,
-			DestImage:           destOpts,
-			RetryOpts:           retryOpts,
-			Dev:                 false,
-		}
-		opts.Global.ConfigPath = "test"
-		opts.Global.LogLevel = "info"
-		opts.ParallelLayerImages = 5
-		opts.ParallelImages = 5
-
-		ex := &ExecutorSchema{
-			Log:     log,
-			Opts:    opts,
-			LogsDir: "/tmp/",
-		}
-
-		err := ex.Validate([]string{"file://test"})
-		if err != nil {
-			t.Fatalf("should not fail")
-		}
-
-		// check for config path error
-		opts.Global.ConfigPath = ""
-		err = ex.Validate([]string{"file://test"})
-		assert.Equal(t, "use the --config flag it is mandatory", err.Error())
-
-		// check when using file protocol --from should not be used
-		opts.Global.ConfigPath = "test"
-		opts.Global.From = "test"
-		err = ex.Validate([]string{"file://test"})
-		assert.Equal(t, "when destination is file://, mirrorToDisk workflow is assumed, and the --from argument is not needed", err.Error())
-
-		// check when using --from protocol must be of type file://
-		opts.Global.ConfigPath = "test"
-		opts.Global.From = "test"
-		err = ex.Validate([]string{"docker://test"})
-		assert.Equal(t, "when --from is used, it must have file:// prefix", err.Error())
-
-		// check destination protocol
-		opts.Global.ConfigPath = "test"
-		opts.Global.From = ""
-		err = ex.Validate([]string{"test"})
-		assert.Equal(t, "destination must have either file:// (mirror to disk) or docker:// (diskToMirror) protocol prefixes", err.Error())
 
 		// check that since is a valid date
 		opts.Global.ConfigPath = "test"
 		opts.Global.From = ""
+		opts.Global.WorkingDir = ""
 		opts.Global.SinceString = "2024-01-01"
 		assert.NoError(t, ex.Validate([]string{"file://test"}))
-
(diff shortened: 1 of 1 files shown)


v2/internal/pkg/cli/executor.go
```diff
@@ -374,6 +374,13 @@ func (o ExecutorSchema) Validate(dest []string) error {
 			return fmt.Errorf("--since flag needs to be in format yyyy-MM-dd")
 		}
 	}
+	// OCPBUGS-58467
+	if o.Opts.ParallelImages > 10 || o.Opts.ParallelImages < 1 {
+		return fmt.Errorf("the flag parallel-images must be between the range 0 to 10")
+	}
```
MayXuQQ, line 380: range 1 to 10? 0 is not allowed

lmzuccarelli: Yes - also I figured 10 should be max as we have had complaints about oc-mirror v2 when default setting were 10 for the parallel-images and 8 for the parallel-layers

MayXuQQ: How about changing the message 0 to 1?

> the flag parallel-images must be between the range 0 to 10

MayXuQQ: Where is the code checking the parallel-layers value?

lmzuccarelli: @MayXuQQ - code for parallel-layers is just below :)

v2/internal/pkg/cli/executor.go
```diff
@@ -187,7 +187,7 @@ func NewMirrorCmd(log clog.PluggableLoggerInterface) *cobra.Command {
 		Example:       mirrorExamples,
 		Args:          cobra.MinimumNArgs(1),
 		SilenceErrors: true,
-		SilenceUsage:  false,
+		SilenceUsage:  true,
```
r4f4: Why are you undoing this change? I changed it because of https://github.com/openshift/oc-mirror/pull/1200#pullrequestreview-2973920153
See https://github.com/openshift/oc-mirror/pull/1200/commits/815cfded57b54ea0da7de650d9e8881d68bce9a8 for the reason.

v2/internal/pkg/cli/executor_test.go
```diff
@@ -504,7 +507,19 @@ func TestExecutorValidate(t *testing.T) {
...
+		// check ParallelImages
+		opts.ParallelImages = 11
+		err = ex.Validate([]string{"file://test"})
+		assert.Equal(t, "the flag parallel-images must be between the range 1 to 10", err.Error())
+
+		// check ParallelLayerImages
+		opts.ParallelImages = 5
+		opts.ParallelLayerImages = 11
+		err = ex.Validate([]string{"file://test"})
+		assert.Equal(t, "the flag parallel-layers must be between the range 1 to 10", err.Error())
```
r4f4: Semantically these don't belong to this test case. They should be in their own `"validate should fail"`. Also instead of `assert.Equal` let's use `assert.Error` which prints a more clear message when the test fails.

v2/internal/pkg/cli/executor.go
lmzuccarelli: It shows the help layout on the console before the error. Horrible user experience

r4f4: How is that horrible? If you use the oc-mirror flags in the wrong way (or the wrong flags) the help output will display the correct flags and their description.

v2/internal/pkg/cli/executor_test.go
```diff
@@ -564,6 +567,48 @@ func TestExecutorValidate(t *testing.T) {
...
+		// check ParallelImages
+		opts.ParallelImages = 11
+		err := ex.Validate([]string{"file://test"})
+		assert.Error(t, err)
+
+		// check ParallelLayerImages
+		opts.ParallelImages = 5
+		opts.ParallelLayerImages = 11
+		err = ex.Validate([]string{"file://test"})
+		assert.Error(t, err)
```
r4f4: Let's add tests for `0` and negative values as well so we test both conditions in the validation check.

v2/internal/pkg/cli/executor_test.go
```diff
@@ -564,6 +567,48 @@ func TestExecutorValidate(t *testing.T) {
 		opts.Global.WorkingDir = "" // reset
 		assert.Equal(t, "when destination is docker://, either --from (assumes disk to mirror workflow) or --workspace (assumes mirror to mirror workflow) need to be provided", ex.Validate([]string{"docker://test"}).Error())
 	})
+
+	t.Run("Testing Executor : validate should fail", func(t *testing.T) {
```
r4f4: Since you're creating a "should fail" test case, let's use the opportunity to refactor the previous "should pass" testcase and move all the checks for failures into this testcase?

v2/internal/pkg/cli/executor_test.go
```diff
@@ -499,70 +514,92 @@ func TestExecutorValidate(t *testing.T) {
...
 		}
 
+		// check ParallelImages
+		opts.ParallelImages = 11
 		err := ex.Validate([]string{"file://test"})
-		if err != nil {
-			t.Fatalf("should not fail")
-		}
+		expectedErr := "the flag parallel-images must be between the range 1 to 10"
+		assert.EqualErrorf(t, err, expectedErr, "Error should be: %s, got: %w", expectedErr, err)
```
r4f4: `EqualErrorf` doesn't understand the `%w` verb it's specific to `fmt.Errorf`:
```
--- FAIL: TestExecutorValidate (0.00s)
    --- FAIL: TestExecutorValidate/Testing_Executor_:_validate_should_fail (0.00s)
        executor_test.go:521: 
            	Error Trace:	~/prjs/github.com/openshift/oc-mirror/v2/internal/pkg/cli/executor_test.go:521
            	Error:      	Error message not equal:
            	           	expected: "the flag parallel-images must be between the range 1 to 10 banana"
            	           	actual  : "the flag parallel-images must be between the range 1 to 10"
            	Test:       	TestExecutorValidate/Testing_Executor_:_validate_should_fail
            	Messages:   	Error should be: the flag parallel-images must be between the range 1 to 10 banana, got: %!w(*errors.errorString=&{the flag parallel-images must be between the range 1 to 10})
```
Also if the assert message is just going to say the errors don't match, we don't need `EqualErrorf`, `EqualError` suffices:
```suggestion
		assert.EqualError(t, err,  "the flag parallel-images must be between the range 1 to 10")
```
```
--- FAIL: TestExecutorValidate (0.00s)
    --- FAIL: TestExecutorValidate/Testing_Executor_:_validate_should_fail (0.00s)
        executor_test.go:521: 
            	Error Trace:	~/prjs/github.com/openshift/oc-mirror/v2/internal/pkg/cli/executor_test.go:521
            	Error:      	Error message not equal:
            	           	expected: "the flag parallel-images must be between the range 1 to 10 banana"
            	           	actual  : "the flag parallel-images must be between the range 1 to 10"
            	Test:       	TestExecutorValidate/Testing_Executor_:_validate_should_fail
```

r4f4 (commented):
/lgtm

PR #1220, opened 2025-07-22
OCPBUGS-59646: cmd: unpack v2 to $HOME instead of /tmp

r4f4:
# Description

It seems it's quite common for users to run oc-mirror in a system with the /tmp dir mounted `noexec` for STIG compliance. In that case, oc-mirror v2 would silently fail because we were not properly printing the error output and the binary could not be executed.

Even though it's possible to workaround the issue by temporarily overriding the default tmp dir with the `TMPDIR` envvar, let's change the target directory for unpacking v2 to `$HOME/.oc-mirror` so the binary works in STIG-compliant systems out-of-the-box.

Github / Jira issue: OCPBUGS-59646

## Type of change

# How Has This Been Tested?

TBA

## Expected Outcome
oc-mirror v2 works when `/tmp` is mounted `noexec`.


aguidirh (commented):
Thanks for this PR @r4f4.

It looks good to me.

I have only one question, it seems that `tmpDir, err := os.MkdirTemp("", "oc-mirror-")` was always creating a random folder under temp dir which allow running several oc-mirror processes in parallel with no concurrent conflict. 

With the new approach this is not going to conflict for concurrent usage ? For example, the first oc-mirror process extract the binary under the user home dir and a second oc-mirror process does the same. 

The first exits and removes the binary ([L54](https://github.com/openshift/oc-mirror/pull/1220/commits/d63343c99a521f67710410293d3670c4762d28ad#diff-7507486b75db598af6bb47d14b737760bcdee3628e8ef962363db5c63c695aaeR54)) while the second process is still running.

If this is a valid point, would this suggestion below be a good mix of both approachs? 

```
 tmpDir, err := os.MkdirTemp("", "oc-mirror-")
   if err != nil {
       // Fallback to home directory
       homeDir, err := os.UserHomeDir()
       // ... rest of implementation
   }
``` 
If this is not a valid point, I'm more than happy to approve this PR since this is my only concern.

PR #1231, opened 2025-07-31
NO-JIRA: updates dependabot config

aguidirh:
# Description

Updates the dependabot config to the following:
* Upgrade v1 dependencies only in case of CVEs
* Upgrade v2 dependencies when new versions are available and also in case of CVEs

## Type of change

# How Has This Been Tested?

It was tested on [my forked](https://github.com/aguidirh/oc-mirror/pulls) version of oc-mirror.

## Expected Outcome
v1 dependencies will be updated only in case of CVES and v2 dependencies will be updated in case of CVEs and new versions available.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,14 +1,24 @@
 version: 2
 updates:
+  # Configuration for the oc-mirror v1 go mod
   - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


r4f4 (commented):
/lgtm

PR #1256, opened 2025-08-29
OCPBUGS-60999: fix issue when using full true in m2m

aguidirh:
# Description

Before this fix during m2m when using full: true with no packages specified, the dispatch func was trying to retrieve a rebuilt catalog image from cache. This behavior was wrong since full: true with no packages specified does not rebuild the catalog.

This PR fixes this behavior, it gets the original catalog image in case it is a full catalog (`full: true` with no packages specified in the ImageSetConfiguration).

Github / Jira issue: [OCPBUGS-60999](https://issues.redhat.com/browse/OCPBUGS-60999)

## Type of change

# How Has This Been Tested?

With the following ImageSetConfiguration: 

```
kind: ImageSetConfiguration
mirror:
  operators:
  - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.18
    targetCatalog: registry.test/operatorsredhat/redhat/catalog
    targetTag: v4.18_20250829
    full: true
```

Run `m2m`. 

## Expected Outcome
It should pass with the code from this PR and fail with the code from the main branch.

### DIFF
--- a/v2/internal/pkg/api/v2alpha1/type_internal.go
+++ b/v2/internal/pkg/api/v2alpha1/type_internal.go
@@ -195,6 +195,8 @@ type RelatedImage struct {
 	//Used to identify if a related image is from an operator catalog on disk (oci:// on ImageSetConfiguration)
 	//TODO remove me when the migration from oc-mirror v1 to v2 ends
 	OriginFromOperatorCatalogOnDisk bool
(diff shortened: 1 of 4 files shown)


r4f4 (commented):
LGTM but I haven't tested this myself.

/lgtm
/hold

PR #1333, opened 2025-12-29
CLID-513: operator catalogs pinned by digest in ISC

adolfo-ab:
# Description

Add a feature to pin operator catalogs by digest, even if the original ImageSetConfig operator catalogs are referenced by tag.

When running the m2d or m2m workflows, a pinned ISC and the corresponding DISC are created in the working directory.

Github / Jira issue: 
N/A

# How Has This Been Tested?
- Adding unit tests
- Relevant scenarios:
  - m2d using an isc with a catalog by tag, then verify that the pinned isc and disc are created
  - m2d by tag, then d2m using the newly created isc pinned by digest, then delete using the pinned disc
  - m2d using an older oc-mirror version without these changes, delete cache and working dir (everything except tar) then d2m using the code of this PR
  - m2m

## Expected Outcome
Please describe the outcome expected from the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,12 @@ The `delete` sub-command has its own set of flags:
 ### Other Features
 * [Enclave support](v2/docs/features/enclave_support.md)
 
+#### Catalog Pinning
+
+When running mirror-to-disk or mirror-to-mirror workflows, oc-mirror automatically generates pinned configuration files where operator catalogs are referenced by SHA256 digests instead of tags. This ensures reproducible mirroring operations since digest references always point to the exact same image content, while tag-based references can change over time.
+
+Two files are generated in the working directory: `isc_pinned_{timestamp}.yaml` contains a pinned ImageSetConfiguration with catalogs referenced by digest, and `disc_pinned_{timestamp}.yaml` contains a corresponding DeleteImageSetConfiguration. The pinned ISC can be used for reproducible mirrors of the exact same content, while the pinned DISC can be used later with the delete sub-command to remove precisely what was mirrored.
+
 ## Testing
 It is possible to run in this module unit tests.
 
--- a/internal/pkg/api/v2alpha1/type_config.go
+++ b/internal/pkg/api/v2alpha1/type_config.go
@@ -48,14 +48,14 @@ type DeleteImageSetConfigurationSpec struct {
 // Mirror defines the configuration for content types within the imageset.
 type Mirror struct {
 	// Platform defines the configuration for OpenShift and OKD platform types.
-	Platform Platform `json:"platform,omitempty"`
+	Platform Platform `json:"platform,omitempty,omitzero"`
 	// Operators defines the configuration for Operator content types.
 	Operators []Operator `json:"operators,omitempty"`
 	// AdditionalImages defines the configuration for a list
 	// of individual image content types.
 	AdditionalImages []Image `json:"additionalImages,omitempty"`
 	// Helm define the configuration for Helm content types.
-	Helm Helm `json:"helm,omitempty"`
+	Helm Helm `json:"helm,omitempty,omitzero"`
 	// BlockedImages define a list of images that will be blocked
 	// from the mirroring process if they exist in other content
 	// types in the configuration.
@@ -68,14 +68,14 @@ type Mirror struct {
 // Delete defines the configuration for content types within the imageset.
 type Delete struct {
 	// Platform defines the configuration for OpenShift and OKD platform types.
-	Platform Platform `json:"platform,omitempty"`
+	Platform Platform `json:"platform,omitempty,omitzero"`
 	// Operators defines the configuration for Operator content types.
 	Operators []Operator `json:"operators,omitempty"`
 	// AdditionalImages defines the configuration for a list
 	// of individual image content types.
 	AdditionalImages []Image `json:"additionalImages,omitempty"`
 	// Helm define the configuration for Helm content types.
-	Helm Helm `json:"helm,omitempty"`
+	Helm Helm `json:"helm,omitempty,omitzero"`
 	// Samples defines the configuration for Sample content types.
 	// This is currently not implemented.
 	Samples []SampleImages `json:"samples,omitempty"`
--- a/internal/pkg/cli/executor.go
+++ b/internal/pkg/cli/executor.go
@@ -888,6 +888,8 @@ func (o *ExecutorSchema) RunMirrorToDisk(cmd *cobra.Command, args []string) erro
 		return batchError
 	}
 
+	o.createConfigsWithPinnedCatalogs(collectorSchema)
+
 	o.Log.Info(emoji.Package + " Preparing the tarball archive...")
 	return o.MirrorArchiver.BuildArchive(cmd.Context(), copiedSchema.AllImages)
 }
@@ -933,6 +935,8 @@ func (o *ExecutorSchema) RunMirrorToMirror(cmd *cobra.Command, args []string) er
 	// call the batch worker
 	// NOTE: we will check for batch errors at the end
 	copiedSchema, batchError := o.Batch.Worker(cmd.Context(), collectorSchema, *o.Opts)
+
+	o.createConfigsWithPinnedCatalogs(collectorSchema)
 
 	// create IDMS/ITMS
 	forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0
@@ -1336,4 +1340,21 @@ func removeDuplicatedImages(allRelatedImages []v2alpha1.CopyImageSchema, mode st
 			return a.Source == b.Source && a.Destination == b.Destination && a.Origin == b.Origin
 		}
 	})
+}
+
+// createConfigsWithPinnedCatalogs generates and writes pinned ISC and DISC configurations.
+func (o *ExecutorSchema) createConfigsWithPinnedCatalogs(collectorSchema v2alpha1.CollectorSchema) {
+	o.Log.Info("Generating pinned configurations...")
+	iscPath, discPath, err := config.PinAndWriteISCAndDSC(
+		o.Config,
+		collectorSchema.CatalogToFBCMap,
+		o.Opts.Global.WorkingDir,
+		o.Log,
+	)
+	if err != nil {
+		o.Log.Warn("Failed to generate pinned configs: %v", err)
+	} else {
+		o.Log.Info("Pinned ISC written to: %s", iscPath)
+		o.Log.Info("Pinned DISC written to: %s", discPath)
+	}
 }
--- a/internal/pkg/config/pin_catalogs.go
+++ b/internal/pkg/config/pin_catalogs.go
@@ -1,0 +1,212 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/oc-mirror/v2/internal/pkg/api/v2alpha1"
+	"github.com/openshift/oc-mirror/v2/internal/pkg/image"
+	clog "github.com/openshift/oc-mirror/v2/internal/pkg/log"
+)
+
+// PinAndWriteISCAndDSC pins catalogs and writes both ISC and DISC files.
+// Returns paths to both written files (ISC path, DISC path).
+func PinAndWriteISCAndDSC(
+	cfg v2alpha1.ImageSetConfiguration,
+	catalogToFBCMap map[string]v2alpha1.CatalogFilterResult,
+	workingDir string,
+	log clog.PluggableLoggerInterface,
+) (string, string, error) {
+	// Pin catalog digests
+	pinnedISC, err := pinCatalogDigests(cfg, catalogToFBCMap, log)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to pin catalog digests: %w", err)
+	}
+
+	// Write pinned ISC
+	iscPath, err := writePinnedISC(pinnedISC, workingDir)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to write pinned ISC: %w", err)
+	}
+
+	// Write pinned DISC
+	discPath, err := createDISCFromISC(pinnedISC, workingDir)
+	if err != nil {
+		return iscPath, "", fmt.Errorf("failed to create pinned DISC: %w", err)
+	}
+
+	return iscPath, discPath, nil
+}
+
+// pinCatalogDigests creates a copy of ImageSetConfiguration with catalog references
+// pinned to SHA256 digests instead of tags.
+//
+// For each operator catalog in cfg.Mirror.Operators:
+// - Parses the catalog reference
+// - Looks up the digest in catalogToFBCMap
+// - Replaces the catalog field with digest format: {registry}/{path}@sha256:{digest}
+//
+// Returns a new ImageSetConfiguration with pinned catalogs and any errors encountered.
+func pinCatalogDigests(
+	cfg v2alpha1.ImageSetConfiguration,
+	catalogToFBCMap map[string]v2alpha1.CatalogFilterResult,
+	log clog.PluggableLoggerInterface,
+) (v2alpha1.ImageSetConfiguration, error) {
+	if len(cfg.Mirror.Operators) == 0 {
+		return cfg, nil
+	}
+
+	// Create copy to avoid mutating original
+	pinnedCfg := copyISC(cfg)
+
+	// Iterate through operator catalogs by index to modify in place
+	for i := range pinnedCfg.Mirror.Operators {
+		op := &pinnedCfg.Mirror.Operators[i]
+
+		pinnedRef, err := pinSingleCatalogDigest(op.Catalog, catalogToFBCMap, log)
+		if err != nil {
+			return v2alpha1.ImageSetConfiguration{}, err
+		}
+
+		if pinnedRef != "" {
+			op.Catalog = pinnedRef
+		}
+	}
+
+	return pinnedCfg, nil
+}
+
+// pinSingleCatalogDigest pins a single catalog reference to its SHA256 digest.
+//
+// Returns:
+// - pinnedRef: The pinned catalog reference (empty string if catalog should be skipped)
+// - error: Fatal error if parsing fails
+//
+// The function handles the following cases:
+// - Already digest-pinned: returns empty string (skip)
+// - Not found in catalogToFBCMap: returns empty string (skip)
+// - Empty digest: returns empty string (skip)
+// - Parse error: returns error
+// - Success: returns pinned reference
+func pinSingleCatalogDigest(
+	catalog string,
+	catalogToFBCMap map[string]v2alpha1.CatalogFilterResult,
+	log clog.PluggableLoggerInterface,
+) (string, error) {
+	// Parse catalog reference
+	imgSpec, err := image.ParseRef(catalog)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse catalog %s: %w", catalog, err)
+	}
+
+	// Skip if the image is already pinned by digest
+	if imgSpec.IsImageByDigest() {
+		log.Debug("Catalog %s is already pinned by digest, skipping", catalog)
+		return "", nil
+	}
+
+	// Look up digest in map
+	filterResult, ok := catalogToFBCMap[imgSpec.ReferenceWithTransport]
+	if !ok {
+		// Log warning but continue (non-fatal)
+		log.Warn("Catalog %s not found in CatalogToFBCMap, skipping pin", catalog)
+		return "", nil
+	}
+
+	// Check for empty digest
+	if filterResult.Digest == "" {
+		log.Warn("Empty digest for catalog %s, skipping pin", catalog)
+		return "", nil
+	}
+
+	// Build pinned reference: {registry}/{path}@sha256:{digest}
+	pinnedRef := image.WithDigest(imgSpec.Name, filterResult.Digest)
+
+	// Add transport prefix if non-docker (docker:// is default and can be omitted)
+	if imgSpec.Transport != "" && imgSpec.Transport != "docker://" {
+		pinnedRef = imgSpec.Transport + pinnedRef
+	}
+
+	log.Debug("Pinning catalog %s to %s", catalog, pinnedRef)
+	return pinnedRef, nil
+}
+
+// writeConfigToFile writes a config object (ISC or DISC) to a YAML file with timestamp naming.
+// Returns the absolute path to the written file.
+func writeConfigToFile(obj interface{}, configType, workingDir string) (string, error) {
+	filename := fmt.Sprintf("%s_pinned_%s.yaml", configType, time.Now().UTC().Format(time.RFC3339))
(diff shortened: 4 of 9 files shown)


README.md
```diff
@@ -241,6 +241,12 @@ The `delete` sub-command has its own set of flags:
 ### Other Features
 * [Enclave support](v2/docs/features/enclave_support.md)
 
+#### Catalog Pinning
+
+When running mirror-to-disk or mirror-to-mirror workflows, oc-mirror automatically generates pinned configuration files where operator catalogs are referenced by SHA256 digests instead of tags. This ensures reproducible mirroring operations since digest references always point to the exact same image content, while tag-based references can change over time.
+
+Two files are generated in the working directory: `isc_pinned_{timestamp}.yaml` contains a pinned ImageSetConfiguration with catalogs referenced by digest, and `disc_pinned_{timestamp}.yaml` contains a corresponding DeleteImageSetConfiguration. The pinned ISC can be used for reproducible mirrors of the exact same content, while the pinned DISC can be used later with the delete sub-command to remove precisely what was mirrored.
```
r4f4, line 248: I don't think we need to generate the `DeleteImageSetConfiguration` since it can be easily obtained from the `ImagetSetConfiguration` by replacing the `kind` and the top level `mirror` -> `delete`. But I'm not against generating it.

internal/pkg/cli/executor.go
```diff
@@ -757,7 +756,7 @@ func (o *ExecutorSchema) isLocalStoragePortBound() bool {
 // the correct local storage directory
 func (o *ExecutorSchema) setupLocalStorageDir() error {
 	o.LocalStorageDisk = filepath.Join(o.Opts.Global.CacheDir, cacheRelativePath)
-	err := os.MkdirAll(o.LocalStorageDisk, 0755)
+	err := os.MkdirAll(o.LocalStorageDisk, 0o755)
```
r4f4: Can you move these type of changes to their own commit?

README.md
adolfo-ab: @aguidirh wdyt?

aguidirh: My initial thought was generating the `DeleteImageSetConfiguration` to avoid customers creating it wrongly (even though we know it is almost the same as the ImageSetConfiguration) and also to make easier the creation of delete automations on pipelines (they would need to only reuse the `DeleteImageSetConfiguration` included in the tar.

I would suggest to generate it but only if we don't see more disadvantages than advantages of doing it.

internal/pkg/cli/executor.go
adolfo-ab: Done

internal/pkg/cli/executor.go
```diff
@@ -888,6 +888,21 @@ func (o *ExecutorSchema) RunMirrorToDisk(cmd *cobra.Command, args []string) erro
...
+		collectorSchema.CatalogToFBCMap,
+		o.Opts.Global.WorkingDir,
+		o.Log,
+	)
+	if err != nil {
+		o.Log.Warn("Failed to generate pinned configs: %v", err)
+	} else {
+		o.Log.Info("Pinned ISC written to: %s", iscPath)
+		o.Log.Info("Pinned DISC written to: %s", discPath)
+	}
```
aguidirh: Since this code is the same for m2d and m2m, it could be extracted to a common func to avoid duplication.

internal/pkg/operator/common.go
```diff
@@ -282,6 +282,18 @@ func (o OperatorCollector) prepareM2DCopyBatch(images map[string][]v2alpha1.Rela
...
+				if img.Type == v2alpha1.TypeOperatorCatalog && len(img.CatalogDigest) > 0 && !imgSpec.IsImageByDigestOnly() {
+					// Create digest-based tag: sha256-{digest}
+					digestTag := "sha256-" + img.CatalogDigest
+					// Build destination with digest tag (replace existing tag)
+					destParts := strings.Split(dest, ":")
+					if len(destParts) >= 2 {
+						digestDest := strings.Join(destParts[:len(destParts)-1], ":") + ":" + digestTag
+						result = append(result, v2alpha1.CopyImageSchema{Source: src, Destination: digestDest, Origin: imgSpec.ReferenceWithTransport, Type: img.Type, RebuiltTag: img.RebuiltTag})
+					}
+				}
```
aguidirh: Not sure if it is needed to store the catalog with digest-based tag. The original catalog is already stored today if I remember correctly and it could be retrieved by digest only.

internal/pkg/config/pin_catalogs.go
```diff
@@ -0,0 +1,188 @@
...
+func PinCatalogDigests(
+	cfg v2alpha1.ImageSetConfiguration,
+	catalogToFBCMap map[string]v2alpha1.CatalogFilterResult,
+	log clog.PluggableLoggerInterface,
+) (v2alpha1.ImageSetConfiguration, error) {
+	// Create copy to avoid mutating original
+	pinnedCfg := copyISC(cfg)
+
+	// Iterate through operator catalogs
+	for i := range pinnedCfg.Mirror.Operators {
```
aguidirh, line 68: nit suggestion: for _, op := range pinnedCfg.Mirror.Operators

In this way there is no need to handle the index directly in line 49.

internal/pkg/config/pin_catalogs.go
```diff
@@ -0,0 +1,188 @@
...
+	return pinnedCfg, nil
+}
+
+// WritePinnedISC writes an ImageSetConfiguration to a YAML file with timestamp naming.
+//
+// The file is written to: {workingDir}/isc_pinned_{timestamp}.yaml
+// e.g., isc_pinned_2025-12-31T11:37:17Z.yaml
+//
+// Returns the absolute path to the written file.
+func WritePinnedISC(
```
aguidirh: Since the code of WritePinnedISC is very similar with CreateDISCFromISC, it would be possible to reuse common code to avoid code duplication.

internal/pkg/config/pin_catalogs.go
```diff
@@ -0,0 +1,188 @@
...
+	if err := os.WriteFile(filePath, yamlData, 0o600); err != nil {
+		return "", fmt.Errorf("failed to write pinned DISC to %s: %w", filePath, err)
+	}
+
+	return filePath, nil
+}
+
+// PinAndWriteConfigs pins catalogs and writes both ISC and DISC files.
+// Returns paths to both written files (ISC path, DISC path).
+func PinAndWriteConfigs(
```
aguidirh: This is the only func used outside of config pkg, so it is correct to expose it. The other funcs PinCatalogDigests, WritePinnedISC and CreateDISCFromISC are used only internally in the config pkg so they could be lower-case which makes the funcs not exposed externally.

Another suggestion is about exposing PinAndWriteConfigs as an interface, this could allow us changing the implementation easily in future and also mocking tests for this func outside of the config pkg.

internal/pkg/config/pin_catalogs.go
adolfo-ab: I tried, but `for _, op := range slice` creates a copy of each element, so modifying op.Catalog only changes the copy and not the original element.

For reference: https://www.calhoun.io/does-range-copy-the-slice-in-go/

internal/pkg/cli/executor.go
adolfo-ab: Done

internal/pkg/config/pin_catalogs.go
adolfo-ab: Done

internal/pkg/config/pin_catalogs.go
adolfo-ab: Partially done, want to discuss about the interface a bit more. Maybe do that in a follow-up PR?

aguidirh: @r4f4 the interface suggestion, what do you think about it?

internal/pkg/config/pin_catalogs.go
```diff
@@ -0,0 +1,181 @@
...
+func writeConfigToFile(obj interface{}, configType, workingDir string) (string, error) {
+	filename := fmt.Sprintf("%s_pinned_%s.yaml", configType, time.Now().UTC().Format(time.RFC3339))
+	filePath := filepath.Join(workingDir, filename)
+
+	yamlData, err := yaml.Marshal(obj)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal %s to YAML: %w", strings.ToUpper(configType), err)
+	}
+
+	if err := os.WriteFile(filePath, yamlData, 0o600); err != nil {
```
aguidirh: When writing the file should we allow others to read? Probably it is not mandatory that the user who wrote the file is the same who is going to read it in another phase.

internal/pkg/operator/filtered_collector.go
```diff
@@ -129,11 +129,23 @@ func createFolders(paths []string) error {
...
+	// Use catalog name + digest to ensure:
+	// - Same catalog with same filter → same rebuiltTag (both v4.18 and the corresponding digest produce same hash)
+	// - Different catalog versions with same filter → different rebuiltTag (v4.18 and v4.19 produce different hashes)
+	if c.Catalog != "" && catalogDigest != "" {
+		imgSpec, err := image.ParseRef(c.Catalog)
+		if err == nil {
+			// Normalize to: name@sha256:digest
+			c.Catalog = imgSpec.Name + "@sha256:" + catalogDigest
+		}
+	}
```
aguidirh: @r4f4 I was wondering if changing the digestOfFilter could break something, but I was not able to find a scenario where it would go wrong. Do you think of any?

aguidirh:
The scenario from my previous comment is going to return this:

```
aguidi@customer:~/go/src/github.com/aguidirh/oc-mirror$ ./bin/oc-mirror -c ./alex-tests/alex-isc/isc.yaml --from file://alex-tests/pin-test/ docker://localhost:5000 --v2 --dest-tls-verify=false
2026/01/16 09:49:34  [INFO]   : 👋 Hello, welcome to oc-mirror
2026/01/16 09:49:34  [INFO]   : ⚙️  setting up the environment for you...
2026/01/16 09:49:34  [INFO]   : ⚙️  environment version: v0.2.0-alpha.1-508-g3f8031f
2026/01/16 09:49:34  [INFO]   : 🔀 workflow mode: diskToMirror 
2026/01/16 09:49:34  [INFO]   : Verified we can authenticate against registry "localhost:5000"
2026/01/16 09:49:34  [INFO]   : 📦 Extracting mirror archive(s)...
alex-tests/pin-test/mirror_000001.tar (7.0 GiB / 7.0 GiB) [=========================================================================================================================================================================================] 25s
2026/01/16 09:50:00  [INFO]   : 🕵  going to discover the necessary images...
2026/01/16 09:50:00  [INFO]   : 🔍 collecting release images...
2026/01/16 09:50:00  [INFO]   : 🔍 collecting operator images...
 ✓   (8s) Collecting catalog registry.redhat.io/redhat/redhat-operator-index:v4.20 
2026/01/16 09:50:08  [INFO]   : 🔍 collecting additional images...
2026/01/16 09:50:08  [INFO]   : 🔍 collecting helm images...
2026/01/16 09:50:08  [INFO]   : 🚀 Start copying the images...
2026/01/16 09:50:08  [INFO]   : 📌 images to copy 17 
 ✓   (2s)  aws-load-balancer-controller-rhel9@sha256:03106a145ba31ae648092a48016d28d3b41dca737f76efdc82eae4628e4fa905 ➡️  localhost:5000/albo/  
 ✓   (3s)  aws-load-balancer-rhel8-operator@sha256:16e9ffed36107527a37713ac5bd34a7bc20f042269a81077429fb5884914c4d0 ➡️  localhost:5000/albo/  
 ✓   (5s)  ose-kube-rbac-proxy@sha256:ba9ff4c933739f1774bc8277d636053c5306863221a8c7b7b9ddc4470eb7feff ➡️  localhost:5000/openshift4/  
 ✓   (2s)  aws-load-balancer-rhel8-operator@sha256:6e15df39a3d25755a79888f739e1af8b052c3fba57bf8f6b3434ef58fd4288cf ➡️  localhost:5000/albo/  
 ✓   (4s)  aws-load-balancer-controller-rhel8@sha256:f0b47b8200f85b557497a4f03d7b186cf415c4137e9cc0361633b67891cf2a24 ➡️  localhost:5000/albo/  
 ✓   (2s)  aws-load-balancer-rhel8-operator@sha256:53f14e88f05641545b675ff4edbcff6909d752659c02a1babf7491734f347153 ➡️  localhost:5000/albo/  
 ✓   (2s)  aws-load-balancer-controller-rhel8@sha256:73db97d20ee9415e9a8941cc6a36ed81d901d760d9949e95e10c90866ef8bf64 ➡️  localhost:5000/albo/  
 ✓   (2s)  aws-load-balancer-controller-rhel8@sha256:2e0b9332a44d8d9c23e19c7accab0813a651f39210257820db508cac28876595 ➡️  localhost:5000/albo/  
 ✓   (13s)  ose-kube-rbac-proxy@sha256:f11f71448986aa17abec9caadb568a6cc34ef1a7898e6dc20bc6a512830ba476 ➡️  localhost:5000/openshift4/  
 ✓   (1s)  aws-load-balancer-rhel9-operator@sha256:1ce50e777efaef0e8df8d9488c1d065cbfa87494b8f191ac79c5ccd4cd37034f ➡️  localhost:5000/albo/  
 ✓   (0s)  aws-load-balancer-operator-bundle@sha256:dbad5680cdb7bbe998b569533d1135b9df8a37786ed64893373152a7f16b3e12 ➡️  localhost:5000/albo/  
 ✓   (11s)  ose-kube-rbac-proxy@sha256:9f9f8f69ab98c1f25cc301572339e0cc5c5b76f2e29ed26e10ac91ad20e0090c ➡️  localhost:5000/openshift4/  
 ✗   ()  redhat-operator-index:v4.20 ➡️  localhost:5000/redhat/  
 ✓   (0s)  aws-load-balancer-operator-bundle@sha256:01f2ca529d2486f113bcefc9fedce6a6fd07bcb48aaf534394b5b04c353f8853 ➡️  localhost:5000/albo/  
 ✓   (0s)  aws-load-balancer-operator-bundle@sha256:59ff4966f4f653e49d42d97bfed6351d3730128db74f666ee4ecacff11790baf ➡️  localhost:5000/albo/  
 ✓   (0s)  aws-load-balancer-operator-bundle@sha256:e371c45e259eaf54d79b1bfc6c47b1093d87965a8d6076205d4860047146ed43 ➡️  localhost:5000/albo/  
17 / 17 (21s) [===================================================================================================================================================================================================================================] 100 %
 ✓   (11s)  ose-kube-rbac-proxy@sha256:4a2324acaea757bae3b01b1aec59f49f4dd79bd1868e69d3418d57c210a6dfd9 ➡️  localhost:5000/openshift4/  
2026/01/16 09:50:30  [INFO]   : === Results ===
2026/01/16 09:50:30  [INFO]   :  ✗  16 / 17 operator images mirrored: Some operator images failed to be mirrored - please check the logs
2026/01/16 09:50:30  [ERROR]  : [Worker] error mirroring image docker://registry.redhat.io/redhat/redhat-operator-index:v4.20 error: initializing source docker://localhost:55000/redhat/redhat-operator-index:dfe4d68572653245c42a4fda9610a8be: reading manifest dfe4d68572653245c42a4fda9610a8be in localhost:55000/redhat/redhat-operator-index: manifest unknown
2026/01/16 09:50:30  [INFO]   : 📄 Generating IDMS file...
2026/01/16 09:50:30  [INFO]   : alex-tests/pin-test/working-dir/cluster-resources/idms-oc-mirror.yaml file created
2026/01/16 09:50:30  [INFO]   : 📄 No images by tag were mirrored. Skipping ITMS generation.
2026/01/16 09:50:30  [INFO]   : mirror time     : 55.901258813s
2026/01/16 09:50:30  [INFO]   : 👋 Goodbye, thank you for using oc-mirror
2026/01/16 09:50:30  [ERROR]  : [Executor] [Worker] some errors occurred during the mirroring.
         Please review alex-tests/pin-test/working-dir/logs/mirroring_errors_20260116_094934.txt for a list of mirroring errors.
         You may consider:
         * removing images or operators that cause the error from the image set config, and retrying
         * keeping the image set config (images are mandatory for you), and retrying
         * mirroring the failing images manually, if retries also fail. 
```

I believe we could fix the issue above calling `digestOfFilter` in this way:

```
var digest string
	if imgSpec.IsImageByDigestOnly() {
		digest = catalogDigest
	}

	filterDigest, err := digestOfFilter(op, digest)
	if err != nil {
		return v2alpha1.CatalogFilterResult{}, err
	}
```

internal/pkg/config/pin_catalogs.go
adolfo-ab: done

internal/pkg/operator/filtered_collector.go
```diff
@@ -129,18 +129,65 @@ func createFolders(paths []string) error {
...
+
+	legacyDigestFile := filepath.Join(filteredCatalogsDir, legacyDigest, "digest")
+	if _, err := os.Stat(legacyDigestFile); err == nil {
+		// Migrate legacy folder to new location
+		legacyDir := filepath.Join(filteredCatalogsDir, legacyDigest)
+		newDir := filepath.Join(filteredCatalogsDir, filterDigest)
+		if err := os.Rename(legacyDir, newDir); err != nil {
+			return "", fmt.Errorf("migrate legacy filter digest folder: %w", err)
+		}
```
aguidirh: I'm having troubles in understanding why it is needed to migrate the legacy folder to a new location, could you please tell me what I'm missing?

adolfo-ab: This is just a nice to have, in order to migrate rebuilt catalogs that still use the old sha.

adolfo-ab: Reverted this change since it was causing some issues with the cache, will address in a follow-up PR.

internal/pkg/operator/filtered_collector.go
```diff
@@ -435,7 +489,7 @@ func TagRebuiltCatalogByDigestOnly(collectorSchema *v2alpha1.CollectorSchema, lo
 			continue
 		}
 		dest := strings.Split(img.Destination, imgSpec.Algorithm)
-		if len(dest) == 0 {
+		if len(dest) >= 1 {
```
aguidirh: `if len(dest) == 0` was to avoid accessing an unavailable index when doing dest[0]

The current `if len(dest) >= 1` means there is a dest[0] so it should reach to line 496 but with the current implementation it will be skipped.

Am I missing anything?

adolfo-ab: Let me explain. The `if len(dest) == 0` condition was not being triggered because when we use `strings.Split`, the result always has at least length 1 (the whole string without splitting), and that was causing the `targetTag` bug you reported in a previous review. That being said, my previous code was also wrong because it should've been `		if len(dest) <= 1`. Now it's fixed.

internal/pkg/config/pin_catalogs.go
```diff
@@ -0,0 +1,182 @@
...
+		// Build pinned reference: {registry}/{path}@sha256:{digest}
+		pinnedRef := fmt.Sprintf("%s@sha256:%s", imgSpec.Name, filterResult.Digest)
+
+		// Add transport prefix if non-docker (docker:// is default and can be omitted)
+		if imgSpec.Transport != "" && imgSpec.Transport != "docker://" {
+			pinnedRef = imgSpec.Transport + pinnedRef
+		}
+
+		log.Debug("Pinning catalog %s to %s", op.Catalog, pinnedRef)
+		op.Catalog = pinnedRef
```
r4f4: I think this could be its own function so that the loop reads:
```go
for i := range pinnedCfg.Mirror.Operators {
    if err := pinCatalogDigest(&pinnedCfg.Mirror.Operators[i], catalogToFBCMap, log); err != nil {
        return nil, err
    }
}
```
That way we can unit test the pinning of a single catalog.

internal/pkg/operator/filtered_collector.go
```diff
@@ -129,18 +138,58 @@ func createFolders(paths []string) error {
...
+//   - When empty: uses the catalog reference as-is.
+func digestOfFilter(catalog v2alpha1.Operator, catalogDigest string) (string, error) {
 	c := catalog
 	c.TargetCatalog = ""
 	c.TargetTag = ""
 	c.TargetCatalogSourceTemplate = ""
+	if c.Catalog != "" && catalogDigest != "" {
+		imgSpec, err := image.ParseRef(c.Catalog)
+		if err == nil {
+			c.Catalog = imgSpec.Name + "@sha256:" + catalogDigest
```
r4f4: Nit: We are doing this operation a lot. We should create a function for it to avoid eventual typos. For ex:
```go
func WithDigest(name, digest string) string {
    return fmt.Sprintf("%s@sha256:%s", name, digest)
}
```

internal/pkg/operator/filtered_collector.go
r4f4: I don't know about "wrong" but currently it is not supported to have the same catalog twice in the ISC with different filterings. This change would allow different versions of the same catalog to be filtered and mirrored at the same time:
```yaml
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: registry.redhat.io/redhat/redhat-operator-index@sha256:deadbeef
      packages:
        - name: operator1
    - catalog: registry.redhat.io/redhat/redhat-operator-index@sha256:foobar
      packages:
        - name: operator2
```
where `@sha256:deadbeef` is the digest of the `v4.20` catalog from day `N` and `@sha256:foobar` is the digest of `v4.20` on day `N+1`.

adolfo-ab: An option could be to delete the "legacy" digest of filtered catalogs and migrate them to the new format, but we would have to change it in the cache as well, I think.

@r4f4 @aguidirh What do you think?

internal/pkg/operator/filtered_collector.go
adolfo-ab: Done

internal/pkg/config/pin_catalogs.go
adolfo-ab: Done

internal/pkg/config/pin_catalogs.go
r4f4: For now I think it's fine. We can refactor into an interface later.

internal/pkg/operator/filtered_collector.go
r4f4: I think deleting the "legacy" digest might be too disruptive and we'd need to test it very well before doing such a change. I think we can leave that for after we revamp our test suite.

internal/pkg/api/v2alpha1/type_config.go
```diff
@@ -68,14 +68,14 @@ type Mirror struct {
 // Delete defines the configuration for content types within the imageset.
 type Delete struct {
 	// Platform defines the configuration for OpenShift and OKD platform types.
-	Platform Platform `json:"platform,omitempty"`
+	Platform Platform `json:"platform,omitzero"`
```
r4f4: Why did you replace here but kept both for the `Mirror` fields?

adolfo-ab: Missed it, fixed

nidangavali (approved):

PR #1339, opened 2026-01-16
OCPBUGS-69745: fix duplicated channel bug in filtered catalogs

r4f4:
# Description

When we were extracting the catalog configs from a filtered oci catalog image, we were using a custom implementation of OCI layers extraction. That implementation did not take into account deleted files (via the whiteout directive [1]), causing `oc-mirror` to see the original catalogs configs from the unfiltered catalog image as well as the filtered configs.

In enclave scenarios, where this filtered catalog was used as a base for further filtering, we ended up with a situation where operators would contain duplicated configs when the original configuration file was not named `catalog.json`. These duplicated configs, in turn, would cause the declcfg OLM library to throw an error about duplicated channels.

Since we already use the `go-containerregistry` library to build the filtered catalog image, this change proposes the usage of `mutate.Extract` from that library to do the OCI layers extraction while taking into account deleted files/directories instead of using or own (flawed) implementation.

This changeset also includes the following changes:
 * Refactoring for functions reuse
 * Rename the whiteout file from `/.wh.configs` to `.wh.configs` to workaround a bug in `go-containerregistry` not taking into account absolute paths when comparing them.
 * Updates the custom implementation `ExtractOCILayers` to use go-containerregistry `mutate.Extract`.
 * Update the dependency so we get the fix for replaced directories in oci layers [2]

Thanks to @sherine-k for helping out with the bug investigation and suggestion of the proper fix.

[1] https://specs.opencontainers.org/image-spec/layer/#whiteouts
[2] https://github.com/google/go-containerregistry/pull/2191

Github / Jira issue: OCPBUGS-69745

## Type of change

# How Has This Been Tested?

* m2m to registry `localhost:5000` with the following ISC:
```yaml
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.18
      packages:
        - channels:
            - minVersion: 1.9.0
              name: stable
          name: netobserv-operator
```

* m2m from `localhost:5000` to `localhost:5001` with `full:true` for the catalog:
```yaml
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: localhost:5000/redhat/redhat-operator-index:v4.18
      full: true
```

* m2m from `localhost:5000` to `localhost:5001` with a filtered catalog config:
```yaml
kind: ImageSetConfiguration
mirror:
  operators:
    - catalog: localhost:5000/redhat/redhat-operator-index:v4.18
      packages:
        - channels:
            - minVersion: 1.9.0
              name: stable
          name: netobserv-operator
```

* m2m to `localhost:5000` with the following config:
```yaml
kind: ImageSetConfiguration
mirror:
  platform:
    channels:
      - name: stable-4.16
        minVersion: 4.16.1
        maxVersion: 4.16.1
    graph: false
```

## Expected Outcome

* for the catalog m2m to `localhost:5000`:
```bash
2026/01/16 16:27:51  [INFO]   : 🔂 rebuilding catalogs
2026/01/16 16:27:52  [INFO]   : 🚀 Start copying the images...
2026/01/16 16:27:52  [INFO]   : 📌 images to copy 38 
2026/01/16 16:37:58  [INFO]   : === Results ===
2026/01/16 16:37:58  [INFO]   :  ✓  38 / 38 operator images mirrored successfully
```

* for the m2m to `localhost:5001` with `full:true`: no errors about duplicate channels and only 38 images instead of copying the full unfiltered catalog
```bash
2026/01/16 17:00:29  [DEBUG]  : [OperatorImageCollector] setting copy option o.Opts.MultiArch=all when collecting operator images
2026/01/16 17:00:29  [DEBUG]  : [OperatorImageCollector] copying operator image localhost:5000/redhat/redhat-operator-index:v4.18
2026/01/16 17:00:29  [DEBUG]  : Filtering catalog "localhost:5000/redhat/redhat-operator-index:v4.18"
2026/01/16 17:00:29  [DEBUG]  : Catalog has not been filtered previously
2026/01/16 17:00:29  [DEBUG]  : Ensuring catalog is in OCI format
2026/01/16 17:00:31  [DEBUG]  : Extracting OCI catalog layers
2026/01/16 17:00:31  [DEBUG]  : found catalog image "sha256:2a2f67814623ada0317fddf9bc0495a5c67d0f02ef9b72d621dde2afa9f1f455"
2026/01/16 17:00:31  [DEBUG]  : [OperatorImageCollector] label "/configs"
2026/01/16 17:00:31  [DEBUG]  : extracting OCI layers to "data/double2/working-dir/operator-catalogs/redhat-operator-index/f3720c531ecb09ffc7d877185ba2d767dfd222c520f91a81eca916795c34bf14/cat
alog-config/configs"
2026/01/16 17:00:37  [DEBUG]  : Found 6 related images for catalog "localhost:5000/redhat/redhat-operator-index:v4.18"
2026/01/16 17:00:37  [DEBUG]  : [OperatorImageCollector] related images length 7 
2026/01/16 17:00:37  [DEBUG]  : [OperatorImageCollector] images to copy (before duplicates) 37 
2026/01/16 17:00:37  [DEBUG]  : [CollectAll] total operator images to copy 38 
[...]
2026/01/16 17:00:37  [INFO]   : 🔂 rebuilding catalogs
2026/01/16 17:00:37  [INFO]   : 🚀 Start copying the images...
2026/01/16 17:00:37  [INFO]   : 📌 images to copy 38 
2026/01/16 17:03:34  [INFO]   : === Results ===
2026/01/16 17:03:34  [INFO]   :  ✓  38 / 38 operator images mirrored successfully
```

* for the other m2m to `localhost:5001`: no errors about duplicate channels
```bash
2026/01/16 17:00:37  [INFO]   : 🔂 rebuilding catalogs
2026/01/16 17:00:37  [INFO]   : 🚀 Start copying the images...
2026/01/16 17:00:37  [INFO]   : 📌 images to copy 38 
2026/01/16 17:03:34  [INFO]   : === Results ===
2026/01/16 17:03:34  [INFO]   :  ✓  38 / 38 operator images mirrored successfully
```

* for the release m2m:
```bash
2026/01/16 22:02:40  [INFO]   : 🔂 rebuilding catalogs
2026/01/16 22:02:40  [INFO]   : 🚀 Start copying the images...
2026/01/16 22:02:40  [INFO]   : 📌 images to copy 191 
```

* for the m2d:
```bash
2026/01/23 01:19:00  [INFO]   : 🔂 rebuilding catalogs
2026/01/23 01:19:01  [INFO]   : 🚀 Start copying the images...
2026/01/23 01:19:01  [INFO]   : 📌 images to copy 37 
2026/01/23 01:21:17  [INFO]   : === Results ===
2026/01/23 01:21:17  [INFO]   :  ✓  37 / 37 operator images mirrored successfully
```


r4f4:
/hold
for fixing the unit tests I missed.

internal/pkg/archive/utils/tar.go
```diff
@@ -0,0 +1,58 @@
...
+	// treatment because of the way Golang handles it after `Clean`
+	absV, err := filepath.Abs(v)
+	if err != nil {
+		return "", fmt.Errorf("get absolute path for %q: %w", v, err)
+	}
+	absDir, err := filepath.Abs(dir)
+	if err != nil {
+		return "", fmt.Errorf("get absolute path for %q: %w", dir, err)
+	}
+	if strings.HasPrefix(absV, absDir+string(os.PathSeparator)) {
```
adolfo-ab, line 28: Maybe we could use [filepath.Rel](https://pkg.go.dev/path/filepath#Rel) here. But we don't need to do this now necessarily, since you're only relocating this function in this PR.

internal/pkg/archive/utils/tar.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	f, err := os.OpenFile(filePath, os.O_CREATE|os.O_RDWR, perm)
+	if err != nil {
+		return fmt.Errorf("unable to create file %s: %w", filePath, err)
+	}
+	defer f.Close()
+
+	// copy contents in chunks to avoid gosec:G110 decompression bomb
+	// https://stackoverflow.com/questions/67327323/g110-potential-dos-vulnerability-via-decompression-bomb-gosec
```
adolfo-ab, line 49: In the stackoverflow thread someone mentions that while this snippets prevents the gosec alert, it wouldn't prevent the actual attack necessarily. Maybe (in this or a different PR) we could try using [io.LimitReader](https://pkg.go.dev/io#LimitReader)?

internal/pkg/common/tar.go
```diff
@@ -0,0 +1,58 @@
...
+
+// WriteFile copies bytes from reader to a file.
+func WriteFile(filePath string, reader io.Reader, perm os.FileMode) error {
+	// make sure all the parent directories exist
+	descriptorParent := filepath.Dir(filePath)
+	if err := os.MkdirAll(descriptorParent, 0o755); err != nil {
+		return fmt.Errorf("unable to create parent directory for %s: %w", filePath, err)
+	}
+
+	f, err := os.OpenFile(filePath, os.O_CREATE|os.O_RDWR, perm)
```
adolfo-ab: What would happen here if the file already existed?

internal/pkg/operator/common.go
```diff
@@ -346,79 +353,82 @@ func (o OperatorCollector) extractOCIConfigLayers(catalog string, imgSpec image.
...
-			return "", fmt.Errorf("manifest %s: %w", catalog, err)
-		}
+	// We must use mutate.Extract here because it properly handles the whiteout
+	// configuration introduced during catalog filtering. Otherwise we could be
+	// extracting a layer with the unfiltered catalog configs that's supposed
+	// to be hidden.
+	imgReader := mutate.Extract(archfulImg)
+	defer imgReader.Close()
+
+	if err := manifest.Untar(imgReader, configsDir, label); err != nil {
```
adolfo-ab: Do we need to cleanup if Untar fails partway?

internal/pkg/operator/common.go
adolfo-ab, line 341: These 2 are unused now right?
```suggestion
func (o OperatorCollector) extractOCIConfigLayers(imageIndexDir string) (string, error) { //nolint:cyclop // TODO: this needs further refactoring
```

internal/pkg/operator/common.go
```diff
@@ -346,79 +353,82 @@ func (o OperatorCollector) extractOCIConfigLayers(catalog string, imgSpec image.
...
 	}
 
-	// read the config digest to get the detailed manifest
-	// looking for the label to search for a specific folder
-	configDigest, err := digest.Parse(oci.Config.Digest)
+	return toPath, nil
+}
+
+// getCatalogConfigImage recurses through OCI image indexes to find an archful catalog image.
+func (o OperatorCollector) getCatalogConfigImage(index v1.ImageIndex) (v1.Image, error) { // nolint:ireturn // this is the type from the go-containerregistry lib.
```
adolfo-ab: Would it make sense to add a max depth for recursion?

internal/pkg/operator/common.go
```diff
@@ -346,79 +353,82 @@ func (o OperatorCollector) extractOCIConfigLayers(catalog string, imgSpec image.
...
-		return "", err
+	for _, desc := range indexManifest.Manifests {
+		switch {
+		case desc.MediaType.IsIndex():
+			ociIndex, err := index.ImageIndex(desc.Digest)
+			if err != nil {
+				o.Log.Debug("failed to get image index %q: %v", desc.Digest.String(), err)
+				continue
+			}
+			return o.getCatalogConfigImage(ociIndex)
```
adolfo-ab: I think this would cause an early return if first index is empty or returns an error, no? Maybe try something like 

```
img, err := o.getCatalogConfigImage(ociIndex)
    if err == nil {
        return img, nil
    }
```

internal/pkg/release/local_stored_collector.go
```diff
@@ -227,6 +210,35 @@ func (o *LocalStorageCollector) ensureReleaseInOCIFormat(ctx context.Context, re
...
+		return fmt.Errorf("no manifests found for release image")
+	}
+	o.Log.Debug(collectorPrefix+"image manifest digest %s", indexManifest.Manifests[0].Digest.String())
+
+	img, err := ociIndex.Image(indexManifest.Manifests[0].Digest)
+	if err != nil {
+		return fmt.Errorf("failed to get release image %q: %w", indexManifest.Manifests[0].Digest.String(), err)
+	}
+
+	imgStream := mutate.Extract(img)
```
adolfo-ab: Should we `defer imgReader.Close()` here?

internal/pkg/operator/common.go
r4f4: In this case I think we want an early return as all the images in a catalog index should have the same config. I'm not sure in which situation we'd fail to get the first image from the manifestlist but would succeed in the others.

internal/pkg/operator/common.go
r4f4: We didn't do it before, so we can do it as a follow-up.

internal/pkg/common/tar.go
r4f4: According to the `man 2 open` page:

> The file offset is set to the beginning of the file (see lseek(2)).

internal/pkg/archive/utils/tar.go
r4f4: To use the `LimitReader` we need to know the size of the data we're writing. So it will require changing the signature of `WriteFile`, which might be better to do in a follow-up PR.

r4f4: Well, `CopyN` uses `LimitReader` internally: https://cs.opensource.google/go/go/+/refs/tags/go1.25.6:src/io/io.go;l=357-364

internal/pkg/release/local_stored_collector.go
r4f4: Done

internal/pkg/operator/common.go
r4f4: Done

internal/pkg/common/tar.go
r4f4: Added `O_TRUNC`

internal/pkg/archive/utils/tar.go
r4f4: We are using the file size now, as reported in the tar header, instead of just reading until EOF.

internal/pkg/archive/utils/tar.go
r4f4: Yeah, let's do in a follow-up.

internal/pkg/common/tar.go
```diff
@@ -0,0 +1,106 @@
+package common
```
aguidirh: nit: rename the package name for something more specific so it is easier to understand the goal of the package.

Source: [Packages names to avoid](https://go.dev/blog/package-names#bad-package-names)

internal/pkg/archive/utils/tar.go
```diff
@@ -0,0 +1,106 @@
...
+
+	f, err := os.OpenFile(filePath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, perm)
+	if err != nil {
+		return fmt.Errorf("unable to create file %s: %w", filePath, err)
+	}
+	defer f.Close()
+
+	// copy contents in chunks to avoid gosec:G110 decompression bomb.
+	// https://stackoverflow.com/questions/67327323/g110-potential-dos-vulnerability-via-decompression-bomb-gosec
+	const maxChunkSize = 2048
```
aguidirh, line 50: Thanks for adding this safeguard.

aguidirh: The value definition of the maxChunkSize came from where? Is it possible to have chunkSize bigger than 2048 without being a attempt of decompression bomb attack?

internal/pkg/manifest/oci-manifest_test.go
```diff
@@ -210,86 +192,9 @@ func TestExtractOCILayers(t *testing.T) {
 			err := os.Mkdir(destDir, 0o600)
 			assert.NoError(t, err)
 
-			oci := &v2alpha1.OCISchema{}
-			err = manifest.ExtractOCILayers(filepath.Join(common.TestFolder, "test-untar", "blobs", "sha256"), destDir, "release-manifests", oci)
+			err = manifest.ExtractOCILayers(&fake.FakeImage{}, destDir, "release-manifests")
 			assert.Error(t, err)
 			assert.Regexp(t, "extract directory: .*", err)
 		})
-		t.Run("when OCI layer digests are malformed", func(t *testing.T) {
```
aguidirh, line 218: Were these tests removed because they are already handled internally in the go-containerregistry?

internal/pkg/archive/utils/tar.go
r4f4: We define that value. I'm not sure what's the right value to pick.

internal/pkg/manifest/oci-manifest_test.go
r4f4: These tests only made sense in the context of our own implementation. If they are not handled by `go-containerregistry`, then the tests belong in that repo, not here.

internal/pkg/common/tar.go
r4f4: We already had a `common` package before this PR and I don't want to have an internal pkg named `tar` since the actual tar implementation is `archive/tar`. These are just helper functions to avoid duplication in our code for some tar-related tasks.
Do you have a better suggestion for a name?

aguidirh: Even if I don't like the word `util`, it could be `archiveutil` or `tarutil` since the funcs there are all related to handling tar files.

r4f4: Done.

nidangavali:
Hi @aguidirh FYI,I am working on this PR currently.

PR #1394, opened 2026-04-24
OCPBUGS-90096: Make dry-run also generate cluster-resources

javipolo:
# Description

Create cluster-resource manifests while running with `--dry-run` flag

Github / Jira issue: 

https://redhat.atlassian.net/browse/RFE-4107

## Type of change

# How Has This Been Tested?

Run `oc-mirror` with `--dry-run` and manifests should apear in working-dir/cluster-resources
It was tested manually

## Expected Outcome
Manifests should be present in `cluster-resources` directory

## Summary by CodeRabbit

* **New Features**
  * Dry-run now writes image mapping output immediately, always logs its path, and creates a missing-image report only when missing images are detected. Cluster resource generation is skipped when mirroring to disk; otherwise IDMS/ITMS, catalog manifests and signature ConfigMap are produced. UpdateService is generated when graph/release images are resolvable.

* **Tests**
  * Refactored/expanded dry-run tests into table-driven cases validating mapping output, missing-image handling, conditional cluster-resources, catalog manifests, and UpdateService content.

### DIFF
--- a/internal/pkg/cli/dryrun.go
+++ b/internal/pkg/cli/dryrun.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 
@@ -26,13 +27,64 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
 	mappingTxtFilePath := filepath.Join(outDir, mappingFile)
 	mappingTxtFile, err := os.Create(mappingTxtFilePath)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create mapping file: %w", err)
 	}
 	defer mappingTxtFile.Close()
-	imagesAvailable := map[string]bool{}
-	nbMissingImgs := 0
 	var buff bytes.Buffer
 	var missingImgsBuff bytes.Buffer
+	nbMissingImgs, nbAvailableImgs := o.processImages(ctx, allImages, &buff, &missingImgsBuff)
+
+	_, err = mappingTxtFile.Write(buff.Bytes())
+	if err != nil {
+		return fmt.Errorf("failed to write mapping file: %w", err)
+	}
+	if err := o.writeMissingImagesFile(outDir, &missingImgsBuff, nbMissingImgs, len(allImages)); err != nil {
+		return err
+	}
+
+	if nbAvailableImgs > 0 {
+		o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", nbAvailableImgs)
+	}
+	o.Log.Info(emoji.PageFacingUp+" list of all images for mirroring in : %s", mappingTxtFilePath)
+
+	// Generate cluster resources in dry-run mode (skip for mirror-to-disk since target registry is unknown)
+	if !o.Opts.IsMirrorToDisk() {
+		if err := o.generateClusterResources(ctx, allImages); err != nil {
+			// In dry-run mode, log cluster resource generation errors as warnings instead of failing
+			o.Log.Warn("Cluster resources generation failed (dry-run mode): %v", err)
+		}
+	}
+
+	return nil
+}
+
+// writeMissingImagesFile creates the missing.txt file if there are missing images
+func (o *ExecutorSchema) writeMissingImagesFile(outDir string, missingImgsBuff *bytes.Buffer, nbMissingImgs, totalImgs int) error {
+	if nbMissingImgs == 0 {
+		return nil
+	}
+
+	missingImgsFilePath := filepath.Join(outDir, missingImgsFile)
+	missingImgsTxtFile, err := os.Create(missingImgsFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to create missing images file: %w", err)
+	}
+	defer missingImgsTxtFile.Close()
+
+	_, err = missingImgsTxtFile.Write(missingImgsBuff.Bytes())
+	if err != nil {
+		return fmt.Errorf("failed to write missing images file: %w", err)
+	}
+
+	o.Log.Warn(emoji.Warning+"  %d/%d images necessary for mirroring are not available in the cache.", nbMissingImgs, totalImgs)
+	o.Log.Warn("List of missing images in : %s.\nplease re-run the mirror to disk process", missingImgsFilePath)
+	return nil
+}
+
+// processImages processes all images and returns the count of missing and available images
+func (o *ExecutorSchema) processImages(ctx context.Context, allImages []v2alpha1.CopyImageSchema, buff, missingImgsBuff *bytes.Buffer) (int, int) {
+	nbMissingImgs := 0
+	nbAvailableImgs := 0
 	for _, img := range allImages {
 		buff.WriteString(img.Source + "=" + img.Destination + "\n")
 		if o.Opts.IsMirrorToDisk() {
@@ -43,33 +95,10 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
 			if err != nil || !exists {
 				missingImgsBuff.WriteString(img.Source + "=" + img.Destination + "\n")
 				nbMissingImgs++
+			} else {
+				nbAvailableImgs++
 			}
 		}
 	}
-
-	_, err = mappingTxtFile.Write(buff.Bytes())
-	if err != nil {
-		return err
-	}
-	if nbMissingImgs > 0 {
-		// creating file for storing list of cached images
-		missingImgsFilePath := filepath.Join(outDir, missingImgsFile)
-		missingImgsTxtFile, err := os.Create(missingImgsFilePath)
-		if err != nil {
-			return err
-		}
-		defer missingImgsTxtFile.Close()
-		_, err = missingImgsTxtFile.Write(missingImgsBuff.Bytes())
-		if err != nil {
-			return err
-		}
-		o.Log.Warn(emoji.Warning+"  %d/%d images necessary for mirroring are not available in the cache.", nbMissingImgs, len(allImages))
-		o.Log.Warn("List of missing images in : %s.\nplease re-run the mirror to disk process", missingImgsFilePath)
-	}
-
-	if len(imagesAvailable) > 0 {
-		o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", len(imagesAvailable))
-	}
-	o.Log.Info(emoji.PageFacingUp+" list of all images for mirroring in : %s", mappingTxtFilePath)
-	return nil
+	return nbMissingImgs, nbAvailableImgs
 }
--- a/internal/pkg/cli/dryrun_test.go
+++ b/internal/pkg/cli/dryrun_test.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
@@ -10,15 +11,15 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/openshift/oc-mirror/v2/internal/pkg/api/v2alpha1"
+	"github.com/openshift/oc-mirror/v2/internal/pkg/clusterresources"
 	"github.com/openshift/oc-mirror/v2/internal/pkg/config"
 	"github.com/openshift/oc-mirror/v2/internal/pkg/consts"
 	clog "github.com/openshift/oc-mirror/v2/internal/pkg/log"
 	"github.com/openshift/oc-mirror/v2/internal/pkg/mirror"
 )
 
-// TestExecutorRunPrepare
-func TestDryRun(t *testing.T) {
-	imgs := []v2alpha1.CopyImageSchema{
+func getTestImages() []v2alpha1.CopyImageSchema {
+	return []v2alpha1.CopyImageSchema{
 		{Source: consts.DockerProtocol + "registry/name/namespace/sometestimage-a@sha256:f30638f60452062aba36a26ee6c036feead2f03b28f2c47f2b0a991e41baebea", Destination: "oci:test"},
 		{Source: consts.DockerProtocol + "registry/name/namespace/sometestimage-b@sha256:f30638f60452062aba36a26ee6c036feead2f03b28f2c47f2b0a991e41baebea", Destination: "oci:test"},
 		{Source: consts.DockerProtocol + "registry/name/namespace/sometestimage-c@sha256:f30638f60452062aba36a26ee6c036feead2f03b28f2c47f2b0a991e41baebea", Destination: "oci:test"},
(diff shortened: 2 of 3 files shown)


aguidirh:
@javipolo could you please have a look on the codeRabbit suggestions and also the linter job that is failing?

javipolo:
@coderabbitai review

javipolo:
> @javipolo could you please have a look on the codeRabbit suggestions and also the linter job that is failing?

@aguidirh done :)

internal/pkg/cli/dryrun.go
```diff
@@ -71,5 +73,82 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
 	return nil
 }
+
+// generateClusterResourcesForDryRun generates all cluster resources that would be created during mirroring
+func (o *ExecutorSchema) generateClusterResourcesForDryRun(allImages []v2alpha1.CopyImageSchema) error {
+	// Skip cluster resource generation if ClusterResources is not initialized (e.g., in tests)
+	if o.ClusterResources == nil {
+		o.Log.Debug("ClusterResources not initialized, skipping cluster resource generation")
+		return nil
+	}
```
aguidirh: It is a best practice to avoid adding production code to satisfy tests purposes. The production code needs to be tested but it shouldn't know about tests (meaning it shouldn't have code related to tests).

internal/pkg/cli/dryrun.go
```diff
@@ -71,5 +73,82 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
+
+	o.Log.Info(emoji.PageFacingUp + " Generating cluster resources...")
+
+	o.generateBasicClusterResources(allImages)
+	o.generateUpdateServiceIfNeeded(allImages)
```
aguidirh: Currently there is no definition about which cluster resource is basic or complex, so I would keep all the calls inside of the `generateClusterResources`

internal/pkg/cli/dryrun.go
```diff
@@ -71,5 +73,82 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
+	// Generate cluster resources in dry-run mode
+	if err := o.generateClusterResourcesForDryRun(allImages); err != nil {
+		return err
+	}
+
 	return nil
 }
+
+// generateClusterResourcesForDryRun generates all cluster resources that would be created during mirroring
+func (o *ExecutorSchema) generateClusterResourcesForDryRun(allImages []v2alpha1.CopyImageSchema) error {
```
aguidirh: If there is no conflict with other funcs, it could be called `generateClusterResources`

internal/pkg/cli/dryrun.go
```diff
@@ -71,5 +73,82 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
 		o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", len(imagesAvailable))
 	}
 	o.Log.Info(emoji.PageFacingUp+" list of all images for mirroring in : %s", mappingTxtFilePath)
+
+	// Generate cluster resources in dry-run mode
+	if err := o.generateClusterResourcesForDryRun(allImages); err != nil {
```
aguidirh: During `m2d` it is not possible to know the target registry, so cluster resources generation shouldn't be generated for `m2d`.

internal/pkg/cli/dryrun.go
```diff
@@ -71,5 +73,82 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
+	// generate signature config map
+	if err := o.ClusterResources.GenerateSignatureConfigMap(allImages); err != nil {
+		// as this is not a seriously fatal error we just log the error
+		o.Log.Warn("Failed to generate signature ConfigMap: %v", err)
+	}
+}
+
+// generateUpdateServiceIfNeeded generates UpdateService if release images are present
+func (o *ExecutorSchema) generateUpdateServiceIfNeeded(allImages []v2alpha1.CopyImageSchema) {
+	releaseImage := o.findFirstReleaseImage(allImages)
```
aguidirh: Try to use the same approach as in `m2m` or `d2m`

```
releaseImage, err := o.Release.ReleaseImage(cmd.Context())
```

internal/pkg/cli/dryrun.go
```diff
@@ -26,13 +27,119 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
+	if err != nil {
+		return fmt.Errorf("failed to write mapping file: %w", err)
+	}
+	if err := o.writeMissingImagesFile(outDir, &missingImgsBuff, nbMissingImgs, len(allImages)); err != nil {
+		return err
+	}
+
+	if len(imagesAvailable) > 0 {
+		o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", len(imagesAvailable))
+	}
```
aguidirh: This is never going to be true, so we can safely remove it. I know it does not come from this PR, but since we are changing the code we can remove it.

internal/pkg/cli/dryrun.go
```diff
@@ -26,13 +27,119 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create mapping file: %w", err)
 	}
 	defer mappingTxtFile.Close()
 	imagesAvailable := map[string]bool{}
 	nbMissingImgs := 0
 	var buff bytes.Buffer
 	var missingImgsBuff bytes.Buffer
+	nbMissingImgs = o.processImages(ctx, allImages, &buff, &missingImgsBuff)
```
aguidirh: nit suggestion: 
```suggestion
	var buff bytes.Buffer
	var missingImgsBuff bytes.Buffer
	nbMissingImgs := o.processImages(ctx, allImages, &buff, &missingImgsBuff)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +142,190 @@ func TestDryRun(t *testing.T) {
...
 
-		global.WorkingDir = testFolder
+	opts := &mirror.CopyOptions{
+		Global:              global,
+		DeprecatedTLSVerify: deprecatedTLSVerifyOpt,
+		SrcImage:            srcOpts,
+		DestImage:           destOpts,
+		RetryOpts:           retryOpts,
+		IsDryRun:            true,
+		Mode:                mirror.MirrorToMirror,
```
aguidirh: Can we add a scenario to check if m2d is not going to generate cluster resources? It could be using a [table driven test](https://go.dev/wiki/TableDrivenTests).

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +142,190 @@ func TestDryRun(t *testing.T) {
...
+	// Note: these may have various names depending on implementation
+	catalogFiles, err := filepath.Glob(filepath.Join(clusterResourcesPath, "*catalog*.yaml"))
+	if err == nil && len(catalogFiles) > 0 {
+		t.Logf("Found catalog files: %v", catalogFiles)
+	}
 
-		for _, img := range imgs {
-			assert.Contains(t, mapping, img.Source+"="+img.Destination)
+	// Verify UpdateService manifest exists for OCP release images
+	updateServiceFiles, err := filepath.Glob(filepath.Join(clusterResourcesPath, "*updateservice*.yaml"))
```
aguidirh:
```suggestion
	updateServiceFiles, err := filepath.Glob(filepath.Join(clusterResourcesPath, "*updateservice*.yaml"))
    require.NotEmpty(t, updateServiceFiles, "UpdateService manifest should be generated when platform.graph=true")

```

aguidirh (changes requested):
Thanks for the changes, it is in a much better shape now!

I've added few comments.

internal/pkg/cli/dryrun.go
adolfo-ab: I think the `imagesAvailable` variable is initialized but never updated, so maybe we can remove it altogether

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,110 @@ func TestDryRun(t *testing.T) {
...
+		t.Fatalf("failed to read mapping file: %v", err)
+	}
+	mapping := string(mappingBytes)
+
+	for _, img := range imgs {
+		assert.Contains(t, mapping, img.Source+"="+img.Destination)
+	}
+
+	missingImgsPath := filepath.Join(testFolder, dryRunOutDir, missingImgsFile)
+	assert.FileExists(t, mappingPath)
```
adolfo-ab:
```suggestion
	assert.FileExists(t, missingImgsPath)
```

adolfo-ab (commented):
Thanks for the PR @javipolo ! 

The only thing that concerns me is that now we have some code duplication for the creation of cluster resources, which is also done in `executor.go` for m2m and d2m. Maybe we can encapsulate the code to generate the cluster resources into a function (your PR is already doing this part) and reuse them in the three places.

Maybe we can do this in a follow-up PR, for the sake of not blocking this PR much longer. Wdyt @aguidirh @r4f4 @dorzel ?

javipolo:
> Thanks for the changes, it is in a much better shape now!
> 
> I've added few comments.

Thanks!

I addressed all the comments. Please take another look when you can!

javipolo:
> Thanks for the PR @javipolo !
> 
> The only thing that concerns me is that now we have some code duplication for the creation of cluster resources, which is also done in `executor.go` for m2m and d2m. Maybe we can encapsulate the code to generate the cluster resources into a function (your PR is already doing this part) and reuse them in the three places.
> 
> Maybe we can do this in a follow-up PR, for the sake of not blocking this PR much longer. Wdyt @aguidirh @r4f4 @dorzel ?

Thanks a lot for the review!
I tried to remove the code duplication in a separate commit. Please see if you like how it's done :)

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
 	// NOTE: we will check for batch errors at the end
 	copiedSchema, batchError := o.Batch.Worker(cmd.Context(), collectorSchema, *o.Opts)
 
-	// create IDMS/ITMS
-	forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0
-	if err := o.ClusterResources.IDMS_ITMSGenerator(copiedSchema.AllImages, forceRepositoryScope); err != nil {
+	// generate cluster resources using centralized function
+	if err := o.generateClusterResources(cmd.Context(), copiedSchema.AllImages, false); err != nil {
```
aguidirh: I'm trying to understand why `forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0` is not needed here anymore and instead false is always being passed.

Could you please explain it to me?

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
-		if err != nil {
+	if err := o.generateUpdateServiceResource(ctx, warnOnError); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// generateIDMSResources generates IDMS/ITMS resources
+func (o *ExecutorSchema) generateIDMSResources(images []v2alpha1.CopyImageSchema, warnOnError bool) error {
```
aguidirh: It seems this `warnOnError` is redundant here, since the `forceRepositoryScope` is the one who is really going to pass the argument needed.

aguidirh: This func could be named `generateImageMirrorSet` which is generic and common for both IDMS and ITMS.

internal/pkg/cli/dryrun.go
```diff
@@ -26,13 +27,59 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
...
+	}
+	if err := o.writeMissingImagesFile(outDir, &missingImgsBuff, nbMissingImgs, len(allImages)); err != nil {
+		return err
+	}
+
+	o.Log.Info(emoji.PageFacingUp+" list of all images for mirroring in : %s", mappingTxtFilePath)
+
+	// Generate cluster resources in dry-run mode (skip for mirror-to-disk since target registry is unknown)
+	if !o.Opts.IsMirrorToDisk() {
+		if err := o.generateClusterResources(ctx, allImages, true); err != nil {
```
aguidirh: If instead of returning the error here, you only warn as you did in handleClusterResourceError, all the warnCode implemented in the executor.go can be removed, reducing in this way the complexity.

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
 
-	// create catalog source
-	if err := o.ClusterResources.CatalogSourceGenerator(copiedSchema.AllImages); err != nil {
+	return batchError
+}
+
+// generateClusterResources generates all cluster resources (IDMS/ITMS, CatalogSource, etc.)
+// This function centralizes cluster resource generation to avoid code duplication across workflows
+func (o *ExecutorSchema) generateClusterResources(ctx context.Context, images []v2alpha1.CopyImageSchema, warnOnError bool) error {
+	if o.ClusterResources == nil {
```
aguidirh, line 1018: This is good guardrails against panic with nil reference since we don't check it in the complete().

Thanks for adding it :)

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+func (o *ExecutorSchema) handleClusterResourceError(message string, err error, warnOnError bool) error {
+	if warnOnError {
+		o.Log.Warn("%s: %v", message, err)
+		return nil
+	}
+	return fmt.Errorf("%s: %w", message, err)
+}
+
+// generateUpdateServiceForClusterResources generates UpdateService resource
+func (o *ExecutorSchema) generateUpdateServiceForClusterResources(ctx context.Context, warnOnError bool) error {
```
aguidirh: It seems all the code here could be placed on `generateUpdateServiceResource`

internal/pkg/cli/dryrun.go
```diff
@@ -26,13 +27,59 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
 	mappingTxtFilePath := filepath.Join(outDir, mappingFile)
 	mappingTxtFile, err := os.Create(mappingTxtFilePath)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create mapping file: %w", err)
```
aguidirh, line 30: Nice wrapping, thanks for adding it.

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,169 +30,382 @@ func TestDryRun(t *testing.T) {
...
+	} else {
+		assert.NoDirExists(t, clusterResourcesPath, "cluster-resources directory should NOT exist for %s mode", tt.mode)
+	}
+}
 
-		for _, img := range imgs {
-			assert.Contains(t, mapping, img.Source+"="+img.Destination)
-		}
+// verifyClusterResourcesExist validates that expected cluster resource files exist
+func verifyClusterResourcesExist(t *testing.T, clusterResourcesPath string, enableGraphGeneration bool) {
```
aguidirh: How about CatalogSource, ClusterCatalog and SignatureConfigMap, should them also be verified here?

internal/pkg/cli/executor.go
```diff
@@ -956,41 +956,11 @@ func (o *ExecutorSchema) RunMirrorToMirror(cmd *cobra.Command, args []string) er
 
 	o.createConfigsWithPinnedCatalogs(collectorSchema)
 
-	// create IDMS/ITMS
-	forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0
-	if err := o.ClusterResources.IDMS_ITMSGenerator(copiedSchema.AllImages, forceRepositoryScope); err != nil {
+	// generate cluster resources using centralized function
```
aguidirh: This comment can be removed.

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
 	// NOTE: we will check for batch errors at the end
 	copiedSchema, batchError := o.Batch.Worker(cmd.Context(), collectorSchema, *o.Opts)
 
-	// create IDMS/ITMS
-	forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0
-	if err := o.ClusterResources.IDMS_ITMSGenerator(copiedSchema.AllImages, forceRepositoryScope); err != nil {
+	// generate cluster resources using centralized function
```
aguidirh: This comment can be removed

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+	if err := o.generateClusterResources(cmd.Context(), copiedSchema.AllImages, false); err != nil {
 		return err
 	}
 
-	// create catalog source
-	if err := o.ClusterResources.CatalogSourceGenerator(copiedSchema.AllImages); err != nil {
+	return batchError
+}
+
+// generateClusterResources generates all cluster resources (IDMS/ITMS, CatalogSource, etc.)
```
aguidirh:
```suggestion
// generateClusterResources generates the following cluster resources: IDMS/ITMS, CatalogSource, ClusterCatalog, UpdateService and SignatureConfigMap.
```

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+// This function centralizes cluster resource generation to avoid code duplication across workflows
```
aguidirh: This comment can be removed

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+// handleClusterResourceError handles errors from cluster resource generation based on the warnOnError flag
+func (o *ExecutorSchema) handleClusterResourceError(message string, err error, warnOnError bool) error {
+	if warnOnError {
+		o.Log.Warn("%s: %v", message, err)
+		return nil
+	}
+	return fmt.Errorf("%s: %w", message, err)
+}
+
+// generateUpdateServiceForClusterResources generates UpdateService resource
```
aguidirh: This comment is redundant

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+}
+
+// handleClusterResourceError handles errors from cluster resource generation based on the warnOnError flag
+func (o *ExecutorSchema) handleClusterResourceError(message string, err error, warnOnError bool) error {
+	if warnOnError {
+		o.Log.Warn("%s: %v", message, err)
+		return nil
+	}
+	return fmt.Errorf("%s: %w", message, err)
+}
```
aguidirh: Probably this func is not needed if error is handled as warning in dry-run.go

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
-			return err
+	}
+
+	if err := o.ClusterResources.ClusterCatalogGenerator(images); err != nil {
+		return o.handleClusterResourceError("Failed to generate ClusterCatalog", err, warnOnError)
+	}
+	return nil
+}
+
+// generateUpdateServiceResource generates UpdateService resource if graph is enabled
```
aguidirh: Redundant comment.

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
+// generateIDMSResources generates IDMS/ITMS resources
+func (o *ExecutorSchema) generateIDMSResources(images []v2alpha1.CopyImageSchema, warnOnError bool) error {
+	forceRepositoryScope := o.Opts.Global.MaxNestedPaths > 0
+	if err := o.ClusterResources.IDMS_ITMSGenerator(images, forceRepositoryScope); err != nil {
+		return o.handleClusterResourceError("Failed to generate IDMS/ITMS", err, warnOnError)
+	}
+	return nil
+}
+
+// generateCatalogResources generates CatalogSource and ClusterCatalog resources
```
aguidirh: Redundant comment.

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
-	if err := o.ClusterResources.ClusterCatalogGenerator(copiedSchema.AllImages); err != nil {
+	if err := o.generateCatalogResources(images, warnOnError); err != nil {
 		return err
 	}
 
-	// generate signature config map
-	if err := o.ClusterResources.GenerateSignatureConfigMap(copiedSchema.AllImages); err != nil {
-		// as this is not a seriously fatal error we just log the error
-		o.Log.Warn("%s", err)
+	// generate signature config map - always warn, as this is not a seriously fatal error
```
aguidirh: Redundant comment.

internal/pkg/cli/executor.go
```diff
@@ -1036,43 +1006,111 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
 	}
 
-	// create catalog source
-	if err := o.ClusterResources.CatalogSourceGenerator(copiedSchema.AllImages); err != nil {
+	return batchError
+}
+
+// generateClusterResources generates all cluster resources (IDMS/ITMS, CatalogSource, etc.)
+// This function centralizes cluster resource generation to avoid code duplication across workflows
+func (o *ExecutorSchema) generateClusterResources(ctx context.Context, images []v2alpha1.CopyImageSchema, warnOnError bool) error {
```
aguidirh: Thanks for making it reusable, it's much better now! 🎉

aguidirh (changes requested):

internal/pkg/cli/executor.go
javipolo: Sure!

forceRepositoryScope is now calculated inside the function `generateIDMSResources`, this way we only have it in one place and not in dry-run, m2m and d2m code

the `false` we're passing to the `generateClusterResources` function controls the `warnOnError` variable

it is false for m2m and d2m, we want to fail if cluster resources fail to be generated there
it is true for dry-run, since we might want the whole dry-run process to pass, even if warning about it

If you dont like that behaviour, we can just make it fail always on cluster resource files generation failure... :shrug:

internal/pkg/cli/executor.go
javipolo: > It seems this `warnOnError` is redundant here, since the `forceRepositoryScope` is the one who is really going to pass the argument needed.

As explained in the previous comment, that is not true. `warnOnError` has nothing to do with `forceRepositoryScope`
During an error generating the ITMS, it will make the code fail or just warn

javipolo: > This func could be named `generateImageMirrorSet` which is generic and common for both IDMS and ITMS.

Agreed. Just changed it!

internal/pkg/cli/dryrun.go
javipolo: We can decide to always warn or always fail, that will remove complexity indeed. I'm happy to update the code to reflect any of the options

Should I just warn always in generateClusterResources?

internal/pkg/cli/executor.go
javipolo: IIRC I created an extra function because linting was failing due to cyclomatic complexity ....

internal/pkg/cli/dryrun_test.go
javipolo: I added some code that validates also catalog files and signatureconfigmaps files

internal/pkg/cli/executor.go
aguidirh: Probably returning an error always instead of a warn would remove few ifs and it would reduce the cyclomatic complexity, make it possible to have everything under only one `generateUpdateServiceResource`

internal/pkg/cli/dryrun.go
aguidirh: Let's keep the previous behavior otherwise we can break customers. I believe the previous behavior was always to return an error, we can continue doing that and in the dry-run code we handle these errors as warnings, in this way the dry-run also keeps working as before.

internal/pkg/cli/executor.go
aguidirh: I think we could handle if the error should fail or only warn on the caller, in this way we could handle cluster resources generation differently only for dry-run, instead of failing we only warn when called by dry-run. But without the need of passing a argument like `warnOnError`.

internal/pkg/cli/dryrun.go
```diff
@@ -46,8 +46,9 @@ func (o *ExecutorSchema) DryRun(ctx context.Context, allImages []v2alpha1.CopyIm
 
 	// Generate cluster resources in dry-run mode (skip for mirror-to-disk since target registry is unknown)
 	if !o.Opts.IsMirrorToDisk() {
-		if err := o.generateClusterResources(ctx, allImages, true); err != nil {
-			return err
+		if err := o.generateClusterResources(ctx, allImages); err != nil {
+			// In dry-run mode, log cluster resource generation errors as warnings instead of failing
+			o.Log.Warn("Cluster resources generation failed (dry-run mode): %v", err)
```
aguidirh, line 54: Thanks, it is simpler now to maintain it 🎉

internal/pkg/cli/executor.go
```diff
@@ -1014,94 +1014,68 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
...
 
-func (o *ExecutorSchema) generateUpdateServiceForClusterResources(ctx context.Context, warnOnError bool) error {
 	graphImage, err := o.Release.GraphImage()
 	if err != nil {
-		if warnOnError {
-			o.Log.Warn("Failed to get graph image reference for UpdateService: %v", err)
-			return nil
-		}
-		return err
+		return fmt.Errorf("failed to get graph image reference for UpdateService: %w", err)
```
aguidirh: Could you please have a look if the `o.Release.GraphImage()` isn't already returning context about the error? I would like to avoid echoing similar error messages across all the call stack. It would be better to have only one (exactly where the error happened), wrap it with context if needed a return it up to the call stack.

This comment is valid also for all the other fmt.Errorf wrappers created on executor.go.

aguidirh (changes requested):

internal/pkg/cli/dryrun.go
javipolo: Indeed it is! :D

internal/pkg/cli/executor.go
javipolo: You're quite right ... I removed a few error wrappings that are redundant and just add noise instead of clarifying the error

aguidirh:
@javipolo please squash the commits and I will approve it.

Don't worry about `/test v1-e2e` for now, it is a flaky test that we are fixing right now.

aguidirh (approved):

aguidirh:
I will let LGTM to @adolfo-ab.

/assign @adolfo-ab

adolfo-ab (approved):
/lgtm

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+	expectSignatures               bool
+	description                    string
+}, imgs []v2alpha1.CopyImageSchema) {
 	log := clog.New("trace")
+	testFolder := t.TempDir()
+
+	ex, testImages, err := setupDryRunTest(t, testFolder, tt, imgs, log)
+	if err != nil {
+		t.Fatalf("setup error: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+
+	err = ex.DryRun(context.TODO(), testImages)
+	if err != nil {
+		t.Fatalf("DryRun should not fail: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+
+	verifyDryRunResults(t, testFolder, tt, testImages)
+	t.Logf("Test '%s' completed successfully: %s", tt.name, tt.description)
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+
+// verifyMappingFile validates the mapping.txt file contents
+func verifyMappingFile(t *testing.T, testFolder string, testImages []v2alpha1.CopyImageSchema) {
+	mappingPath := filepath.Join(testFolder, dryRunOutDir, "mapping.txt")
+	assert.FileExists(t, mappingPath, "mapping.txt should always be created")
+
+	mappingBytes, err := os.ReadFile(mappingPath)
+	if err != nil {
+		t.Fatalf("failed to read mapping file: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+
+// verifyUpdateServiceFile validates the UpdateService file when graph generation is enabled
+func verifyUpdateServiceFile(t *testing.T, clusterResourcesPath string) {
+	updateServiceFile := filepath.Join(clusterResourcesPath, "updateService.yaml")
+	assert.FileExists(t, updateServiceFile, "UpdateService manifest should be generated when platform.graph=true")
+
+	updateServiceContent, err := os.ReadFile(updateServiceFile)
+	if err != nil {
+		t.Fatalf("failed to read UpdateService file %s: %v", updateServiceFile, err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -29,10 +30,373 @@ func TestDryRun(t *testing.T) {
...
+}
+
+// verifyMissingImagesFileExists validates that the missing images file contains expected content
+func verifyMissingImagesFileExists(t *testing.T, missingImgsPath string, testImages []v2alpha1.CopyImageSchema) {
+	assert.FileExists(t, missingImgsPath, "missing.txt should exist when there are missing images")
 
+	missingBytes, err := os.ReadFile(missingImgsPath)
+	if err != nil {
+		t.Fatalf("failed to read missing images file: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
-			RetryOpts:           retryOpts,
-			IsDryRun:            true,
-			Mode:                mirror.MirrorToDisk,
-			Dev:                 false,
-			LocalStorageFQDN:    regCfg.HTTP.Addr,
+	if expectSignatures {
+		// Strict validation: files MUST exist
+		assert.FileExists(t, signatureYamlFile, "signature-configmap.yaml MUST exist when signatures are expected")
+		assert.FileExists(t, signatureJsonFile, "signature-configmap.json MUST exist when signatures are expected")
+		t.Logf("✅ Successfully validated expected signature ConfigMap files")
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
+	} else {
+		// Conditional validation: if files exist, both should exist
+		if yamlExists || jsonExists {
+			assert.FileExists(t, signatureYamlFile, "signature-configmap.yaml should exist when signatures are present")
+			assert.FileExists(t, signatureJsonFile, "signature-configmap.json should exist when signatures are present")
+			t.Logf("✅ Found unexpected signature ConfigMap files (but validated they are correct)")
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
+		} else {
+			t.Logf("ℹ️  No signature ConfigMap files found (expected when no signatures directory exists)")
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
+	if expectSignatures {
+		// Strict validation: files MUST exist
+		assert.FileExists(t, signatureYamlFile, "signature-configmap.yaml MUST exist when signatures are expected")
+		assert.FileExists(t, signatureJsonFile, "signature-configmap.json MUST exist when signatures are expected")
+		t.Logf("✅ Successfully validated expected signature ConfigMap files")
+	} else {
+		// Conditional validation: if files exist, both should exist
+		if yamlExists || jsonExists {
+			assert.FileExists(t, signatureYamlFile, "signature-configmap.yaml should exist when signatures are present")
+			assert.FileExists(t, signatureJsonFile, "signature-configmap.json should exist when signatures are present")
```
r4f4:
```suggestion
		if yamlExists {
			assert.FileExists(t, signatureYamlFile, "signature-configmap.yaml should exist when signatures are present")
		}
	    if jsonExists {
			assert.FileExists(t, signatureJsonFile, "signature-configmap.json should exist when signatures are present")
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
-			log.Error("imagesetconfig %v ", err)
+	}
+}
+
+// verifyCatalogSourceFiles validates CatalogSource files based on expectations
+func verifyCatalogSourceFiles(t *testing.T, clusterResourcesPath string, expectCatalogs bool) {
+	catalogSourceFiles, err := filepath.Glob(filepath.Join(clusterResourcesPath, "cs-*.yaml"))
+	if err != nil {
+		t.Fatalf("failed to search for CatalogSource files: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
+
+	if expectCatalogs {
+		// Strict validation: files MUST exist
+		assert.True(t, len(catalogSourceFiles) > 0, "CatalogSource files MUST be generated when catalogs are expected")
+		t.Logf("✅ Successfully validated %d expected CatalogSource files: %v", len(catalogSourceFiles), catalogSourceFiles)
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
-			t.Fatalf("should not fail")
-		}
-		mappingPath := filepath.Join(testFolder, dryRunOutDir, mappingFile)
-		assert.FileExists(t, mappingPath)
+// verifyClusterCatalogFiles validates ClusterCatalog files based on expectations
+func verifyClusterCatalogFiles(t *testing.T, clusterResourcesPath string, expectCatalogs bool) {
+	clusterCatalogFiles, err := filepath.Glob(filepath.Join(clusterResourcesPath, "cc-*.yaml"))
+	if err != nil {
+		t.Fatalf("failed to search for ClusterCatalog files: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
 
-		mappingBytes, err := os.ReadFile(mappingPath)
-		if err != nil {
-			t.Fatalf("failed to read mapping file: %v", err)
+	if expectCatalogs {
+		// Strict validation: files MUST exist
+		assert.True(t, len(clusterCatalogFiles) > 0, "ClusterCatalog files MUST be generated when catalogs are expected")
+		t.Logf("✅ Successfully validated %d expected ClusterCatalog files: %v", len(clusterCatalogFiles), clusterCatalogFiles)
```
r4f4:
```suggestion
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
 
-		missingImgsPath := filepath.Join(testFolder, dryRunOutDir, missingImgsFile)
-		assert.FileExists(t, mappingPath)
+// setupSignaturesDirectory creates a test signatures directory with sample signature files
+func setupSignaturesDirectory(t *testing.T, testFolder string) {
+	signaturesDir := filepath.Join(testFolder, "signatures")
+	err := os.MkdirAll(signaturesDir, 0755)
+	if err != nil {
+		t.Fatalf("failed to create signatures directory: %v", err)
+	}
```
r4f4:
```suggestion
	assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
 
-		for _, img := range imgs {
-			assert.Contains(t, missing, img.Source+"="+img.Destination)
+	for _, sig := range sampleSignatures {
+		sigFile := filepath.Join(signaturesDir, sig.name)
+		err := os.WriteFile(sigFile, []byte(sig.content), 0600)
+		if err != nil {
+			t.Fatalf("failed to create signature file %s: %v", sigFile, err)
 		}
```
r4f4:
```suggestion
		assert.NoError(t, err)
```

internal/pkg/cli/dryrun_test.go
```diff
@@ -41,157 +405,232 @@ func TestDryRun(t *testing.T) {
...
-	})
+		t.Logf("Created test signature file: %s", sigFile)
```
r4f4:
```suggestion

```

r4f4:
Was the message
```
if len(imagesAvailable) > 0 {
      o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", len(imagesAvailable))
  }
```
removed on purpose?

javipolo:
> Was the message
> 
> ```
> if len(imagesAvailable) > 0 {
>       o.Log.Info("all %d images required for mirroring are available in local cache. You may proceed with mirroring from disk to disconnected registry", len(imagesAvailable))
>   }
> ```
> 
> removed on purpose?

nope. I guess with the refactor it just went away ... I added it back

Please have a look

I also included all your suggestions to the test file

nidangavali (approved):

PR #1411, opened 2026-05-19
CLID-602: Test for enclave scenario

nidangavali:
# Description

Add an integration test for the enclave mirroring scenario. This test validates the two-hop disconnected workflow where operator images are first mirrored from a remote catalog to an intermediate registry, then forwarded to a separate enclave registry using a `registries.conf` redirect. The test verifies that both registries end up with the same set of repositories.

Github / Jira issue: [CLID-602](https://redhat.atlassian.net/browse/CLID-602)

## Type of change

# How Has This Been Tested?

## Expected Outcome
The test starts two local registries, mirrors operators into the first (intermediate), then uses registries.conf to redirect and mirror into the second (enclave). Both registries should contain the same repositories, and the test should pass.

## Summary by CodeRabbit

* **Tests**
  * Added an integration test covering mirror-to-mirror operator mirroring using a `registries.conf` redirect.
  * Added integration test helper utilities to validate mirrored catalog contents, log command diagnostics, and compare repository sets.
* **Bug Fixes**
  * Improved test registry startup to support automatic port selection when no port is specified.

### DIFF
--- a/tests/integration/enclave_test.go
+++ b/tests/integration/enclave_test.go
@@ -1,0 +1,101 @@
+package integration_test
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/oc-mirror/tests/integration/pkg/ocmirror"
+	"github.com/openshift/oc-mirror/tests/integration/pkg/registry"
+)
+
+// A registries.conf redirect makes the enclave-side oc-mirror pull from the
+// intermediate registry instead of the original remote source.
+var _ = Describe("enclave", func() {
+	var workDir string
+
+	BeforeEach(func() {
+		workDir = setupWorkDir()
+	})
+
+	AfterEach(func() {
+		cleanupWorkDir(workDir)
+	})
+
+	Describe("mirror-to-mirror with registries.conf redirect", func() {
+		iscFile := filepath.Join("operators", "isc-operator-version-range.yaml")
+
+		// Workflow:
+		//  1. Mirror operators from the remote catalog into an intermediate registry.
+		//  2. Configure registries.conf to redirect quay.io pulls to the intermediate registry.
+		//  3. Mirror from the intermediate registry into a separate enclave registry.
+		//  4. Assert that the enclave catalog contains the correctly filtered bundles.
+		It("should forward operators from intermediate to enclave registry via registries.conf", func() {
+			iscPath := filepath.Join(iscDir, iscFile)
+			GinkgoWriter.Printf("ISC path: %s\n", iscPath)
+			GinkgoWriter.Printf("workDir: %s\n", workDir)
+
+			By("starting a second registry for the enclave destination")
+			enclaveRegistry, err := registry.Start(ctx, registryConfig, 0)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(enclaveRegistry.WaitReady(ctx, 30*time.Second)).To(Succeed())
+			defer enclaveRegistry.Stop()
+			GinkgoWriter.Printf("intermediate registry: %s\n", testRegistry.Endpoint())
+			GinkgoWriter.Printf("enclave registry:      %s\n", enclaveRegistry.Endpoint())
+
+			By("mirroring from the remote catalog to the intermediate registry")
+			result, err := runner.MirrorToMirror(ctx, iscPath, workDir, testRegistry.Endpoint(),
+				"--remove-signatures=true", "--dest-tls-verify=false")
+			logOcMirrorResult("step-1 mirrorToMirror", result)
+			expectOcMirrorCommandSuccess(result, err)
+
+			By("verifying operator images exist in the intermediate registry")
+			logRegistryRepositories("intermediate", *testRegistry)
+			expectSuccessfulMirrorInRegistry(iscPath, *testRegistry)
+
+			By("mirroring from the intermediate registry to the enclave registry using registries.conf")
+			enclaveWorkDir := setupWorkDir()
+			defer cleanupWorkDir(enclaveWorkDir)
+
+			registriesConf := writeRegistriesConf(enclaveWorkDir, testRegistry.Endpoint())
+			GinkgoWriter.Printf("registries.conf: %s\n", registriesConf)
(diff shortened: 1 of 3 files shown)


adolfo-ab (approved):
/lgtm

adolfo-ab:
@aguidirh can you take a look at this one, since you are more familiar with how enclaves work?

tests/integration/enclave_test.go
```diff
@@ -0,0 +1,121 @@
...
+			"bar": {"bar.v1.0.0"},
+		}
+
+		It("should mirror operators to an intermediate registry and then forward to a second registry", func() {
+			iscSourcePath := filepath.Join(iscDir, iscSource)
+
+			By("starting a second registry for the enclave destination")
+			secondRegistry, err := registry.Start(ctx, registryConfig, 5001)
+			Expect(err).NotTo(HaveOccurred(), "Failed to start second registry")
+			err = secondRegistry.WaitReady(ctx, 30*time.Second)
```
aguidirh: I'm wondering if the registry is taking 30 seconds to come up. 

Could we have another way of checking it in some kind of loop? In this way we improve the test performance.

tests/integration/enclave_test.go
```diff
@@ -0,0 +1,121 @@
...
+			By("verifying operator images are mirrored in the first registry")
+			logRegistryRepositories("intermediate", *testRegistry)
+			expectSuccessfulMirrorInRegistry(iscSourcePath, *testRegistry)
+
+			By("verifying the intermediate catalog contains only filtered operator bundles")
+			expectCatalogBundlesMatchISC(ctx, *testRegistry, iscSourcePath, expectedBundles)
+			expectPackagesNotInMirroredCatalog(ctx, *testRegistry, iscSourcePath, "baz")
+
+			By("verifying IDMS from the first mirror contains expected content")
+			expectCorrectIDMS(workDir, iscSourcePath)
```
aguidirh: I'm not sure if we should test IDMS here since the idea is not to test cluster resources generation.

tests/integration/enclave_test.go
```diff
@@ -0,0 +1,121 @@
...
+			logOcMirrorResult("step-1 mirrorToMirror", result)
+			expectOcMirrorCommandSuccess(result, err)
+
+			By("verifying operator images are mirrored in the first registry")
+			logRegistryRepositories("intermediate", *testRegistry)
+			expectSuccessfulMirrorInRegistry(iscSourcePath, *testRegistry)
+
+			By("verifying the intermediate catalog contains only filtered operator bundles")
+			expectCatalogBundlesMatchISC(ctx, *testRegistry, iscSourcePath, expectedBundles)
+			expectPackagesNotInMirroredCatalog(ctx, *testRegistry, iscSourcePath, "baz")
```
aguidirh: Wondering if we should be more focused when testing one feature, for example, when testing enclave support, should we also test if the catalog was filtered or if cluster resources where generated correctly? 

cc: @adolfo-ab what do you think about it?

tests/integration/testdata/imagesetconfigs/enclave/isc-enclave-full.yaml
```diff
@@ -0,0 +1,9 @@
+# ImageSetConfig for enclave test - Step 2: mirror from first registry to second registry
+# Uses full:true to mirror all content from the intermediate registry's catalog.
+# The catalog reference is replaced at runtime with the intermediate registry endpoint.
+kind: ImageSetConfiguration
+apiVersion: mirror.openshift.io/v2alpha1
+mirror:
+  operators:
+    - catalog: __INTERMEDIATE_CATALOG__
```
aguidirh: There is one missing piece here, when we are doing enclave support the catalog will be like the original one, something like `- catalog: quay.io/oc-mirror/oc-mirror-dev:test-catalog-latest`, using `full: true` is not the way of testing enclave support. Instead we need to place a registries.conf file on the expected directory and add the mirror map, which will tell the cluster, if you find `quay.io/oc-mirror/oc-mirror-dev:test-catalog-latest` get it from `CUSTOMER_HOST/oc-mirror/oc-mirror-dev:test-catalog-latest`

This is also true for other types like releases, helm and additional images.

tests/integration/enclave_test.go
```diff
@@ -0,0 +1,121 @@
...
+		expectedBundles := map[string][]string{
+			"foo": {"foo.v0.2.0", "foo.v0.3.0", "foo.v0.3.1"},
+			"bar": {"bar.v1.0.0"},
+		}
+
+		It("should mirror operators to an intermediate registry and then forward to a second registry", func() {
+			iscSourcePath := filepath.Join(iscDir, iscSource)
+
+			By("starting a second registry for the enclave destination")
+			secondRegistry, err := registry.Start(ctx, registryConfig, 5001)
```
aguidirh: Consider using a dynamic port allocation (OS assigned) to avoid conflicts

aguidirh (commented):
Thanks for this PR, I added few comments.

Please let me know if you have question about them.

Also, please fill the PR description following the template on .github directory, it is important for future references. If possible include context to the commit message.

tests/integration/enclave_test.go
adolfo-ab: We're already doing this, the 30s is just the timeout.

tests/integration/enclave_test.go
adolfo-ab: I think we should only test what's necessary and specific to each test. So if the enclave scenarios do not have any particularity wrt to catalog filtering, and we're already checking catalog filtering in other tests, then we don't need to verify that here. Same for cluster resources.

adolfo-ab (approved):
/lgtm

tests/integration/helpers_test.go
```diff
@@ -774,6 +777,75 @@ func idmsSourceCoversRef(sources []string, ref string) bool {
 	return false
 }
 
+// expectPackagesNotInMirroredCatalog verifies that the given operator packages
+// were not included in the mirrored catalog image stored in the registry.
+func expectPackagesNotInMirroredCatalog(ctx context.Context, reg registry.Registry, iscPath string, packages ...string) {
```
aguidirh: This func is not used so probably it can be safely removed.

aguidirh (changes requested):
Only one comment from my side before approving it.

aguidirh (approved):

PR #1415, opened 2026-05-20
CLID-625, CLID-626: Tests for --parallel-images and --parallel-layers flags

adolfo-ab:
# Description
Adds tests for --parallel-images and --parallel-layers flags.
Replaces manual tests OCP-83188 and OCP-83189.

Github / Jira issue: 
CLID-625, [CLID-626](https://redhat.atlassian.net/browse/CLID-626)

# How Has This Been Tested?
By running the tests

## Summary by CodeRabbit

* **Tests**
  * Added an integration test suite validating parallel image and layer copying behavior and concurrency limits, plus supporting test images and configurations for parallel-copy scenarios.
* **Documentation**
  * Added READMEs describing the additional multi-layer test images, how to build/publish them, and the image reference used by tests.
* **Bug Fixes**
  * Adjusted mirror retry behavior to better recover from a broader range of server error responses.

### DIFF
--- a/tests/integration/parallel_copy_test.go
+++ b/tests/integration/parallel_copy_test.go
@@ -1,0 +1,197 @@
+package integration_test
+
+import (
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/oc-mirror/tests/integration/pkg/ocmirror"
+)
+
+var _ = Describe("parallel copy", func() {
+	var workDir string
+
+	BeforeEach(func() {
+		workDir = setupWorkDir()
+	})
+
+	AfterEach(func() {
+		cleanupWorkDir(workDir)
+	})
+
+	Describe("parallel-layers", func() {
+		Describe("multilayers image mirrorToDisk", func() {
+			iscFile := filepath.Join("parallel_copy", "isc-multilayers.yaml")
+
+			It("should copy at most 10 layers in parallel with --parallel-layers 10", func() {
+				By("running mirrorToDisk with --parallel-layers 10")
+				result, err := runner.MirrorToDisk(ctx, filepath.Join(iscDir, iscFile), workDir,
+					"--parallel-images", "1",
+					"--parallel-layers", "10",
+					"--log-level", "debug",
+				)
+				expectOcMirrorCommandSuccess(result, err)
+
+				By("verifying no more than 10 layers were copied concurrently")
+				expectParallelLayersLimit(result, 10)
+			})
+		})
+
+		Describe("operator mirrorToDisk", func() {
+			iscFile := filepath.Join("operators", "isc-operator-pinned-version.yaml")
+
+			It("should copy at most 5 layers in parallel using the default parallel-layers value", func() {
+				By("running mirrorToDisk with default parallel-layers")
+				result, err := runner.MirrorToDisk(ctx, filepath.Join(iscDir, iscFile), workDir,
+					"--parallel-images", "1",
+					"--log-level", "debug",
+				)
+				expectOcMirrorCommandSuccess(result, err)
+
+				By("verifying no more than 5 layers were copied concurrently")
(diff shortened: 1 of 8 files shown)


tests/integration/parallel_copy_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+	Describe("parallel-layers", func() {
+		Describe("multilayers image mirrorToDisk", func() {
+			iscFile := filepath.Join("parallel_copy", "isc-multilayers.yaml")
+
+			It("should copy at most 10 layers in parallel with --parallel-layers 10", func() {
+				By("running mirrorToDisk with --parallel-layers 10")
+				result, err := runner.MirrorToDisk(ctx, filepath.Join(iscDir, iscFile), workDir,
+					"--parallel-images", "1",
+					"--parallel-layers", "10",
```
aguidirh: Would it be beneficial to add the defaults as constants ?

tests/integration/parallel_copy_test.go
```diff
@@ -0,0 +1,197 @@
...
+		if inProgress > max {
+			max = inProgress
+		}
+	}
+
+	if inProgress > max {
+		max = inProgress
+	}
+	return max
+}
```
aguidirh: I'm wondering if there is a less fragile approach we could rely on. 

Is it difficult to count the connections created to the registry or doing some kind of profiling to see how many go routines were triggered? 

Only throwing ideas here, if it is too difficult maybe we should ignore them.

aguidirh (commented):
Thanks for the PR. I added few comments.

Needs to check why the integration test is failing, is it because of some kind of race condition when images share the same blob on few layers?

tests/integration/parallel_copy_test.go
adolfo-ab: I'm not sure if you mean adding them in the product code, or in the test code. I think it would make sense to add them to the product code only.

tests/integration/parallel_copy_test.go
adolfo-ab: I agree, checking the output is very fragile. I've tried a couple ways to solve this by counting the amount of concurrent request to the registry but I think that will lead to flakiness too, as it depends on the sizes of the images and layers.

IMO checking this at the integration level is wrong, at this level we only care about the behavior, and not about what's happening internally. For now I've simplified the scenarios and only left the most basic checks to make sure that the flags work. Maybe later down the line we can discuss if it's worth to add unit tests to check concurrency, but honestly I think we'll be only checking go std at that point.

I think the right way to do this would be to do some performance benchmarks, where we run the same workflow with different number of parallel layers and images and see how performance is affected. Maybe we can create an initiative for this.

tests/integration/parallel_copy_test.go
aguidirh: Yes, it could be exported constants from the product code.

tests/integration/parallel_copy_test.go
aguidirh: > I think the right way to do this would be to do some performance benchmarks, where we run the same workflow with different number of parallel layers and images and see how performance is affected. Maybe we can create an initiative for this.

I totally agree with testing the performance in some kind of benchmark, please add one epic to our project so we don't forget about it.

tests/integration/parallel_copy_test.go
adolfo-ab: The problem is that `internal` and the integration tests live in different modules, so I can't just import them. We would need to import the whole oc-mirror module including its dependency tree, or put the constants in a common place. For now I don't think any of those options are worth it. Maybe in the future if we end up needing a lot of oc-mirror constants we can do that. Wdyt? @aguidirh

aguidirh: Yeap, I agree, I don't think that the juice will be worth the squeeze.

aguidirh (approved):

r4f4 (commented):
/lgtm

aguidirh:
/assign @nidangavali

For learning purposes only.

aguidirh:
/verified bypass

Since there is no production code being changed, there is no need to verify.

PR #656, opened 2023-06-12
CFE-834: Vendor to the lastest oc

jeffdyoung:
# Description

Vendor to the latest oc to pick up mirroring for manifest listed release payloads.

## Type of change
Update oc vendor to capture this change: https://github.com/openshift/oc/pull/1423

# Passed local tests

- [x] make test-unit
- [x] make test-e2e

**Test Configuration**:
* Hardware: x86_64
* OS: Fedora 37

PR #1435, opened 2026-06-08
CLID-290: feat: Add sparse manifest platform filtering for multi-arch images

aguidirh:
# Description

This PR implements sparse manifest platform filtering for multi-architecture images in oc-mirror. Users can now specify OS/Architecture pairs to filter which platforms are mirrored, significantly reducing disk usage and bandwidth requirements.

The implementation brings the platform filtering capability from containers/image (containers/container-libs PR #656) to oc-mirror. It uses sparse manifests where the manifest list references all platforms but only the specified platforms have their actual layer blobs downloaded.

Platform filtering works across all image types:
- Additional images
- Operator catalogs
- Helm charts
- Release images (with backward compatibility for deprecated `architectures` field)

Github / Jira issue: [CLID-290](https://redhat.atlassian.net/browse/CLID-290)

## Type of change

# How Has This Been Tested?

# Sparse Manifest Platform Filtering — Manual Test Plan

## What the feature does

When `platform.platforms` is set in the ISC, oc-mirror copies only the requested
platform blobs from multi-arch manifest lists. The manifest list index in the target
registry is preserved intact, but only the specified architecture blobs are present.
This requires the target registry to have blob-presence validation disabled
(`validation.manifests.indexes.platforms: none`).

---

### 1. Registries

**Registry for filtered ISC run (`isc-platforms.yaml`)** — `localhost:5000`:
```bash
podman run -dt \
  -p 5000:5000 \
  -v /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/local-registry/registry-config.yml:/etc/distribution/config.yml:z \
  --name local-registry-with-sparse-manifest \
  docker.io/distribution/distribution:3
```

**Registry for baseline ISC run (`isc.yaml`)** — `localhost:6000`:
```bash
podman run -dt \
  -p 6000:5000 \
  -v /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/local-registry/registry-config.yml:/etc/distribution/config.yml:z \
  --name local-registry-without-sparse-manifest \
  docker.io/distribution/distribution:3
```

Both registries use the same config (which includes `platforms: none`). The difference
is which ISC is mirrored to each one.

### 2. ISC files used

| File | Purpose |
|---|---|
| `alex-tests/alex-isc/isc-platforms.yaml` | All image types with `platforms` field — primary test ISC |
| `alex-tests/alex-isc/isc.yaml` | All image types without `platforms` — backward compat + comparison baseline |

### 3. Verification script

```bash
chmod +x scripts/verify-platform-filter.sh
```

---

## Test Suite

---

### TEST 1 — New Platforms field: all image types in one run (M2M)

**ISC:** `alex-tests/alex-isc/isc-platforms.yaml`

```bash
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc-platforms.yaml \
  docker://localhost:5000 \
  --workspace file://alex-tests/clid-290-platforms/working-dir \
  --dest-tls-verify=false
```

**Expected log:** Warning printed before copying:
```
WARN: Platform filtering with the Platforms field requires a registry with manifest
      list blob check disabled (e.g. validation.manifests.indexes.platforms: none).
```

**Verify — report mode:**
```bash
./scripts/verify-platform-filter.sh localhost:5000
```

**What to look for in the report output:**
| Image | Expected present | Expected missing |
|---|---|---|
| `openshift/release-images` (top-level release payload) | amd64, arm64, s390x | ppc64le |
| `openshift/release:4.18.1-multi-*` (component images) | amd64, arm64, s390x | ppc64le |
| `rh_ee_aguidi/multi-platform-container` | amd64, arm64 | s390x, ppc64le |
| `rh_ee_aguidi/empty-image` | all (no filter) | nothing |
| `stefanprodan/podinfo` | amd64 only | arm64, others |
| `projectsigstore/cosigned` | amd64, arm64 | others |
| Operator images from v4.18 catalog | amd64 | arm64+ |
| Operator images from v4.19 catalog | amd64, arm64 | others |
| Operator images from v4.20 catalog | all (no filter) | nothing |
| Operator images from v4.21 catalog | amd64, arm64, s390x, ppc64le | none |

Both `openshift/release-images` (top-level manifest list from Cincinnati) and the
`openshift/release` component images (extracted from image-references) must show only
the configured platforms. The verify script checks both.

> **Why report mode only?** Validate mode applies a global allowed list, but this ISC
> uses per-image platform restrictions that can't be expressed as a single global set.

---

### TEST 2 — Backward compat: old behavior unchanged (M2M)

**ISC:** `alex-tests/alex-isc/isc.yaml` (no `platforms` field anywhere)

```bash
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc.yaml \
  docker://localhost:6000 \
  --workspace file://alex-tests/clid-290/working-dir \
  --dest-tls-verify=false
```

**Expected:**
- No sparse manifest warning in logs
- **Release images**: appear as `single-manifest (no index)` — Cincinnati called with
  `?arch=amd64`, only single-arch amd64 payload
- **Non-release manifest-list images**: all platform blobs present (`[N present, 0 missing]`)

**Verify:**
```bash
./scripts/verify-platform-filter.sh localhost:6000
```

---

### Cleanup between M2M and M2D runs

Before running M2D tests, free up disk space:

```bash
rm -rf alex-tests/clid-290-platforms/working-dir
rm -rf alex-tests/clid-290/working-dir

podman rm -f local-registry-with-sparse-manifest local-registry-without-sparse-manifest

podman run -dt \
  -p 5000:5000 \
  -v /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/local-registry/registry-config.yml:/etc/distribution/config.yml:z \
  --name local-registry-with-sparse-manifest \
  docker.io/distribution/distribution:3

podman run -dt \
  -p 6000:5000 \
  -v /home/aguidi/go/src/github.com/aguidirh/oc-mirror/alex-tests/local-registry/registry-config.yml:/etc/distribution/config.yml:z \
  --name local-registry-without-sparse-manifest \
  docker.io/distribution/distribution:3
```

---

### TEST 3 — New Platforms field: M2D then D2M

**Step 1 — M2D:**
```bash
rm -rf ~/.oc-mirror
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc-platforms.yaml \
  file://alex-tests/clid-290-platforms/m2d-output
```

**Step 2 — Prepare for D2M (simulate disconnected environment):**
```bash
rm -rf ~/.oc-mirror
mkdir -p alex-tests/clid-290-platforms/d2m-input
mv alex-tests/clid-290-platforms/m2d-output/mirror_000001.tar \
   alex-tests/clid-290-platforms/d2m-input/
rm -rf alex-tests/clid-290-platforms/m2d-output/working-dir
```

**Step 3 — D2M:**
```bash
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc-platforms.yaml \
  --from file://alex-tests/clid-290-platforms/d2m-input \
  docker://localhost:5000 \
  --dest-tls-verify=false
```

**Verify — same checks as TEST 1:**
```bash
./scripts/verify-platform-filter.sh localhost:5000
```

**Expected:** Identical results to TEST 1. Platform filtering preserved through archive.
Check logs to confirm no external HTTP calls.

---

### TEST 4 — Backward compat: M2D then D2M with isc.yaml

**Step 1 — M2D:**
```bash
rm -rf ~/.oc-mirror
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc.yaml \
  file://alex-tests/clid-290/m2d-output
```

**Step 2 — Prepare:**
```bash
rm -rf ~/.oc-mirror
mkdir -p alex-tests/clid-290/d2m-input
mv alex-tests/clid-290/m2d-output/mirror_000001.tar \
   alex-tests/clid-290/d2m-input/
rm -rf alex-tests/clid-290/m2d-output/working-dir
```

**Step 3 — D2M:**
```bash
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc.yaml \
  --from file://alex-tests/clid-290/d2m-input \
  docker://localhost:6000 \
  --dest-tls-verify=false
```

**Verify:**
```bash
./scripts/verify-platform-filter.sh localhost:6000
```

**What to look for:**
- Release images: `single-manifest (no index)` — single-arch amd64
- Non-release manifest-list images: `[N present, 0 missing]` — all platforms present

---

### TEST 5 — Validation error: both Platforms and Architectures set

Create `/tmp/invalid-isc.yaml`:
```yaml
kind: ImageSetConfiguration
mirror:
  platform:
    platforms:
      - os: linux
        architecture: amd64
    architectures:
      - arm64
    channels:
      - name: stable-4.18
        minVersion: 4.18.1
        maxVersion: 4.18.1
```

```bash
./bin/oc-mirror --v2 --config /tmp/invalid-isc.yaml \
  docker://localhost:5000 \
  --workspace file:///tmp/test-invalid \
  --dest-tls-verify=false
```

**Expected:** oc-mirror exits immediately with:
```
invalid configuration: platform.platforms and platform.architectures cannot be used together
```

---

### TEST 6 — Deprecated Architectures field with `multi` value

Create `/tmp/isc-archs-multi.yaml`:
```yaml
kind: ImageSetConfiguration
mirror:
  platform:
    architectures:
      - multi
      - amd64
  additionalImages:
    - name: quay.io/rh_ee_aguidi/multi-platform-container:latest
```

```bash
./bin/oc-mirror --v2 --config /tmp/isc-archs-multi.yaml \
  docker://localhost:6000 \
  --workspace file://alex-tests/tmp-archs-multi/working-dir \
  --dest-tls-verify=false
```

**Expected:**
- Deprecation warning: `platform.architectures is deprecated; use platform.platforms instead`
- Command exits successfully (no crash from `multi` in architectures)
- `multi-platform-container` repo exists in `:6000`

**Verify:**
```bash
./scripts/verify-platform-filter.sh localhost:6000
```

---

### TEST 7 — Negative: target registry without sparse manifest support

Start a plain registry (no `platforms: none`):
```bash
podman run -dt \
  -p 7000:5000 \
  --name local-registry-strict \
  docker.io/distribution/distribution:3
```

```bash
./bin/oc-mirror --v2 \
  --config alex-tests/alex-isc/isc-platforms.yaml \
  docker://localhost:7000 \
  --workspace file://alex-tests/tmp-strict/working-dir \
  --dest-tls-verify=false
```

**Expected:**
- Sparse manifest warning printed at start
- Mirror fails with registry error about missing blobs or manifest validation

---

## Summary Table

| # | Workflow | ISC | Registry | Pass condition |
|---|---|---|---|---|
| TEST 1 | M2M | isc-platforms.yaml | :5000 | Only declared arch blobs present per image |
| TEST 2 | M2M | isc.yaml | :6000 | Release: single-arch amd64; others: all blobs |
| TEST 3 | M2D+D2M | isc-platforms.yaml | :5000 | Same as TEST 1, no external HTTP calls |
| TEST 4 | M2D+D2M | isc.yaml | :6000 | Same as TEST 2, no external HTTP calls |
| TEST 5 | — | invalid (both fields) | — | Error before mirroring starts |
| TEST 6 | M2M | archs:[multi,amd64] | :6000 | Success, deprecation warning, no crash |
| TEST 7 | M2M | isc-platforms.yaml | :7000 (strict) | Fails with registry error |

## Known Limitations

1. **Release component images may remain single-arch.** The top-level release image is
   a manifest list and is correctly filtered. Component images from `image-references`
   may still be single-arch depending on the OCP version.

2. **D2M can only copy what M2D stored.** Requesting a platform in D2M that wasn't
   stored in M2D results in "no instances found" fallback — the whole image is copied.

3. **Single-arch operator/helm images.** Many operator bundle images are published
   as single-arch (amd64 only). Platform filtering is a no-op for these; they are
   always copied whole via the fallback retry.

# Script to verify archs on registries:

```
#!/usr/bin/env bash
# verify-platform-filter.sh
#
# Verify which platform blobs are actually stored in a registry for manifest-list images.
# Sparse manifest registries store the full index but may omit blobs for unselected platforms.
# This script distinguishes PRESENT (blob reachable) from MISSING (listed in index but no blob).
#
# Usage:
#   # Report all manifest-list images and their blob presence:
#   ./scripts/verify-platform-filter.sh localhost:6000
#
#   # Validate that no image has blobs outside the expected set (fail on unexpected extras only;
#   # single-arch images with fewer platforms than expected are fine):
#   ./scripts/verify-platform-filter.sh localhost:6000 linux/amd64 linux/arm64
#
#   # Side-by-side comparison of two registries (useful to spot the filtering effect):
#   ./scripts/verify-platform-filter.sh --compare localhost:5000 localhost:6000
#
#   # Side-by-side comparison AND validate the second registry against expected platforms:
#   ./scripts/verify-platform-filter.sh --compare localhost:5000 localhost:6000 linux/amd64 linux/arm64

set -euo pipefail

# ── colours ───────────────────────────────────────────────────────────────────
RED='\033[0;31m'; GREEN='\033[0;32m'; YELLOW='\033[1;33m'
CYAN='\033[0;36m'; BOLD='\033[1m'; RESET='\033[0m'

# ── helpers ───────────────────────────────────────────────────────────────────
die()  { echo -e "${RED}ERROR: $*${RESET}" >&2; exit 1; }
info() { echo -e "${CYAN}$*${RESET}"; }
ok()   { echo -e "    ${GREEN}✔ $*${RESET}"; }
fail() { echo -e "    ${RED}✘ $*${RESET}"; }
warn() { echo -e "    ${YELLOW}~ $*${RESET}"; }

require() { command -v "$1" &>/dev/null || die "$1 is required but not installed."; }
require curl
require python3

# ── registry helpers ──────────────────────────────────────────────────────────

# List all repositories in a registry.
list_repos() {
    curl -sf "http://${1}/v2/_catalog" \
        | python3 -c "import json,sys; [print(r) for r in json.load(sys.stdin).get('repositories',[])]"
}

# List non-signature tags for an image (skip *.sig tags, which are Cosign signatures).
list_tags() {
    local reg=$1 image=$2
    curl -sf "http://${reg}/v2/${image}/tags/list" \
        | python3 -c "
import json,sys
for t in (json.load(sys.stdin).get('tags') or []):
    if not t.endswith('.sig'):
        print(t)
"
}

# Fetch a manifest, accepting both manifest-list and single-manifest media types.
fetch_manifest() {
    local reg=$1 image=$2 ref=$3
    curl -sf \
        -H "Accept: application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.oci.image.index.v1+json,application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json" \
        "http://${reg}/v2/${image}/manifests/${ref}"
}

# Return 0 if the manifest/blob for a digest is reachable (HTTP 200).
digest_present() {
    local reg=$1 image=$2 digest=$3
    local http_code
    http_code=$(curl -so /dev/null -w "%{http_code}" \
        -H "Accept: application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json" \
        "http://${reg}/v2/${image}/manifests/${digest}")
    [[ "$http_code" == "200" ]]
}

# Extract tab-separated "digest  os  arch  variant" lines from a manifest list JSON.
manifest_entries() {
    python3 -c "
import json,sys
m=json.load(sys.stdin)
for e in m.get('manifests',[]):
    p=e.get('platform',{})
    print(e['digest'] + '\t' + p.get('os','?') + '\t' + p.get('architecture','?') + '\t' + p.get('variant',''))
"
}

# ── check one image ───────────────────────────────────────────────────────────
#
# Prints a block like:
#   image:tag
#     ✔ linux/amd64
#     ✘ linux/s390x (blob missing)
#
# Sets globals: IS_INDEX, PRESENT_COUNT, MISSING_COUNT
# Returns 1 if any expected platform is missing or an unexpected one is present.
check_image() {
    local reg=$1 image=$2 tag=$3
    shift 3
    local -a expected_platforms=("$@")

    IS_INDEX=0 PRESENT_COUNT=0 MISSING_COUNT=0

    local manifest
    manifest=$(fetch_manifest "$reg" "$image" "$tag" 2>/dev/null) || return 0

    local has_manifests
    has_manifests=$(echo "$manifest" | python3 -c "import json,sys; print('yes' if 'manifests' in json.load(sys.stdin) else 'no')" 2>/dev/null)
    [[ "$has_manifests" == "yes" ]] || return 0   # single-manifest — nothing to check

    IS_INDEX=1

    # Collect per-platform results before printing so the header can come first.
    local -a lines=()
    local -a present_platforms=()

    while IFS=$'\t' read -r digest os arch variant; do
        local plat="${os}/${arch}"
        [[ -n "$variant" ]] && plat="${plat}/${variant}"

        if digest_present "$reg" "$image" "$digest"; then
            lines+=("ok:${plat}")
            present_platforms+=("$plat")
            (( PRESENT_COUNT++ )) || true
        else
            lines+=("fail:${plat}")
            (( MISSING_COUNT++ )) || true
        fi
    done < <(echo "$manifest" | manifest_entries)

    # Print header with summary counts.
    echo -e "  ${BOLD}${image}:${tag}${RESET}  [${PRESENT_COUNT} present, ${MISSING_COUNT} missing]"

    # Print per-platform lines.
    for line in "${lines[@]}"; do
        local status="${line%%:*}" plat="${line#*:}"
        if [[ "$status" == "ok" ]]; then
            ok "$plat"
        else
            fail "$plat (blob missing)"
        fi
    done

    # Validate against expected platforms if provided.
    # Only fails on UNEXPECTED platforms (present but not in the allowed set).
    # An image having fewer platforms than expected is fine — it may be single-arch.
    if (( ${#expected_platforms[@]} > 0 )); then
        local img_fail=0

        for got in "${present_platforms[@]}"; do
            local expected_flag=0
            for exp in "${expected_platforms[@]}"; do
                # Match exact or prefix (e.g. linux/arm64 matches linux/arm64/v8).
                [[ "$got" == "$exp" || "$got" == "${exp}/"* ]] && expected_flag=1 && break
            done
            (( expected_flag )) || { fail "UNEXPECTED platform present: ${got}"; img_fail=1; }
        done

        return $img_fail
    fi
}

# ── report mode ───────────────────────────────────────────────────────────────

report_registry() {
    local reg=$1
    shift
    local -a expected_platforms=("$@")
    local total_images=0 total_index=0 total_missing=0 total_fail=0

    info "\n=== Registry: ${reg} ==="
    if (( ${#expected_platforms[@]} > 0 )); then
        echo -e "  Expected platforms: ${BOLD}${expected_platforms[*]}${RESET}"
    fi
    echo ""

    while read -r repo; do
        while read -r tag; do
            (( total_images++ )) || true

            if ! check_image "$reg" "$repo" "$tag" "${expected_platforms[@]}"; then
                (( total_fail++ )) || true
            fi

            if (( IS_INDEX )); then
                (( total_index++ )) || true
                (( total_missing += MISSING_COUNT )) || true
            fi
        done < <(list_tags "$reg" "$repo")
    done < <(list_repos "$reg")

    echo ""
    if (( ${#expected_platforms[@]} > 0 )); then
        if (( total_fail == 0 )); then
            echo -e "${GREEN}${BOLD}PASS — all ${total_index} manifest list(s) match expected platforms${RESET}"
        else
            echo -e "${RED}${BOLD}FAIL — ${total_fail} image(s) have platform mismatches${RESET}"
            return 1
        fi
    else
        echo "Scanned ${total_images} image(s). ${total_index} are manifest lists, ${total_missing} missing blobs total."
    fi
}

# ── compare mode ─────────────────────────────────────────────────────────────

compare_registries() {
    local reg1=$1 reg2=$2
    shift 2
    local -a expected_platforms=("$@")

    info "\n=== Comparing ${reg1}  ↔  ${reg2} ==="
    if (( ${#expected_platforms[@]} > 0 )); then
        echo -e "  Validating ${reg2} against: ${BOLD}${expected_platforms[*]}${RESET}"
    fi
    echo ""

    local total=0 diff_count=0 fail_count=0

    # Helper: return newline-separated list of platforms whose blobs are present.
    get_present_platforms() {
        local reg=$1 image=$2 tag=$3
        local manifest
        manifest=$(fetch_manifest "$reg" "$image" "$tag" 2>/dev/null) || return 0
        echo "$manifest" | python3 -c "
import json,sys,urllib.request,urllib.error
m=json.load(sys.stdin)
if 'manifests' not in m: exit(0)
reg='${1}'; im='${2}'
for e in m.get('manifests',[]):
    p=e.get('platform',{})
    plat=p.get('os','?')+'/'+p.get('architecture','?')
    v=p.get('variant','')
    if v: plat+='/'+v
    try:
        req=urllib.request.Request(
            f'http://{reg}/v2/{im}/manifests/'+e['digest'],
            headers={'Accept':'application/vnd.docker.distribution.manifest.v2+json,application/vnd.oci.image.manifest.v1+json'})
        urllib.request.urlopen(req)
        print(plat)
    except urllib.error.HTTPError:
        pass
"
    }

    while read -r repo; do
        while read -r tag; do
            # Check if it's a manifest list in reg1.
            local manifest1
            manifest1=$(fetch_manifest "$reg1" "$repo" "$tag" 2>/dev/null) || continue
            local is_index
            is_index=$(echo "$manifest1" | python3 -c "import json,sys; print('yes' if 'manifests' in json.load(sys.stdin) else 'no')" 2>/dev/null)
            [[ "$is_index" == "yes" ]] || continue

            (( total++ )) || true
            echo -e "  ${BOLD}${repo}:${tag}${RESET}"

            local -a p1 p2
            mapfile -t p1 < <(get_present_platforms "$reg1" "$repo" "$tag")
            mapfile -t p2 < <(get_present_platforms "$reg2" "$repo" "$tag")

            # Union of all platforms across both registries.
            local -A all_plats=()
            for p in "${p1[@]}"; do all_plats["$p"]=1; done
            for p in "${p2[@]}"; do all_plats["$p"]=1; done

            local differs=0
            for plat in $(echo "${!all_plats[@]}" | tr ' ' '\n' | sort); do
                local in1=0 in2=0
                for p in "${p1[@]}"; do [[ "$p" == "$plat" ]] && in1=1; done
                for p in "${p2[@]}"; do [[ "$p" == "$plat" ]] && in2=1; done

                local r1_mark r2_mark
                (( in1 )) && r1_mark="${GREEN}✔${RESET}" || r1_mark="${YELLOW}—${RESET}"
                (( in2 )) && r2_mark="${GREEN}✔${RESET}" || r2_mark="${YELLOW}—${RESET}"

                echo -e "    ${plat}  (${reg1}: ${r1_mark}  ${reg2}: ${r2_mark})"
                (( in1 != in2 )) && differs=1
            done
            (( differs )) && (( diff_count++ )) || true

            # Validate reg2 against expected platforms if provided.
            # Only fails on UNEXPECTED platforms (present but not in the allowed set).
            if (( ${#expected_platforms[@]} > 0 )); then
                local img_fail=0
                for got in "${p2[@]}"; do
                    local expected_flag=0
                    for exp in "${expected_platforms[@]}"; do
                        [[ "$got" == "$exp" || "$got" == "${exp}/"* ]] && expected_flag=1 && break
                    done
                    (( expected_flag )) || { fail "${reg2} has UNEXPECTED: ${got}"; img_fail=1; }
                done
                (( img_fail )) && (( fail_count++ )) || true
            fi

            echo ""
        done < <(list_tags "$reg1" "$repo")
    done < <(list_repos "$reg1")

    echo "Scanned ${total} manifest list(s). ${diff_count} differ between registries."
    if (( ${#expected_platforms[@]} > 0 )); then
        if (( fail_count > 0 )); then
            echo -e "${RED}${BOLD}FAIL — ${fail_count} image(s) in ${reg2} don't match expected platforms${RESET}"
            return 1
        else
            echo -e "${GREEN}${BOLD}PASS — ${reg2} matches expected platforms on all manifest lists${RESET}"
        fi
    fi
}

# ── main ──────────────────────────────────────────────────────────────────────

usage() {
    grep '^# ' "$0" | sed 's/^# //' | grep -v '^$'
    exit 1
}

if [[ $# -eq 0 ]]; then usage; fi

if [[ "$1" == "--compare" ]]; then
    shift
    [[ $# -lt 2 ]] && die "--compare requires two registry arguments"
    reg1=$1; reg2=$2; shift 2
    compare_registries "$reg1" "$reg2" "$@"
else
    reg=$1; shift
    report_registry "$reg" "$@"
fi
```

## Summary by CodeRabbit

* **New Features**
  * Added explicit platform selection using `OS/architecture` filters such as `linux/amd64` and `linux/arm64`.
  * Added platform filtering support for releases, operators, Helm charts, and additional images.
  * Added sparse multi-architecture mirroring, transferring only selected platform manifests and blobs.
  * Added support for platform-aware archive creation and restoration.
* **Bug Fixes**
  * Preserved compatibility with the deprecated architecture configuration and added validation preventing conflicting settings.
* **Tests**
  * Added comprehensive integration coverage for sparse manifests and multi-architecture release mirroring.


internal/pkg/release/local_stored_collector.go
```diff
@@ -50,6 +50,31 @@ func (o LocalStorageCollector) destinationRegistry() string {
...
+// Returns nil if no platforms are specified (meaning mirror all platforms).
+func (o LocalStorageCollector) getPlatformFilters() *[]string {
+	// Use new Platforms field if available
+	if len(o.Config.Mirror.Platform.Platforms) > 0 {
+		return v2alpha1.ConvertPlatformsToStringSlice(o.Config.Mirror.Platform.Platforms)
+	}
+
+	// Fall back to deprecated Architectures field (assume linux)
+	//nolint:staticcheck // SA1019: Architectures is deprecated but we maintain backward compatibility
+	if len(o.Config.Mirror.Platform.Architectures) > 0 {
```
adolfo-ab: Since `Architectures` is deprecated, maybe we can add a log line mentioning it here, and asking to consider using `Platform` instead

internal/pkg/api/v2alpha1/type_platform.go
```diff
@@ -60,3 +60,33 @@ func (pt PlatformType) validate() error {
 	}
 	return nil
 }
+
+// InstancePlatformFilter defines OS and Architecture for filtering
+// multi-architecture manifest lists. This allows selecting specific
+// platforms (e.g., linux/amd64, linux/arm64) when mirroring images.
+type InstancePlatformFilter struct {
```
adolfo-ab, line 69: What happens if a user only sets the OS or the Architecture? Should we check that we always have both? Otherwise we might get `linux/` or `/amd64`, no?

adolfo-ab (changes requested):
Can you check if you need to also update `internal/pkg/config/defaults.go` and `cincinnati.go`?

Also, maybe you could add one or two integration tests, so that this feature is already covered.

internal/pkg/release/local_stored_collector.go
aguidirh: Fixed at [8c26134](https://github.com/aguidirh/oc-mirror/commit/8c26134611ac13e6ea100eec50611c8530b89487)

internal/pkg/api/v2alpha1/type_platform.go
aguidirh: Fixed at [8c26134](https://github.com/aguidirh/oc-mirror/commit/8c26134611ac13e6ea100eec50611c8530b89487)

internal/pkg/release/local_stored_collector.go
aguidirh: Please mark it as resolved in case there is no further suggestions for this thread.

internal/pkg/api/v2alpha1/type_platform.go
aguidirh: Please mark it as resolved in case there is no further suggestions for this thread.

aguidirh:
Next round: 

- Confirm the feature is working by running some manual tests.
- Implement an integration test to cover the sparse manifest feature on all types on both m2m and m2d/d2m workflows.

aguidirh:
/hold

I will unhold it as soon as I send the integration tests commit.

adolfo-ab (commented):
Currently, if I understand correctly, if a user writes a platform that doesn't exist in the ISC, we always continue the mirroring process. 

In the case that the image is a manifest list but doesn't contain the architecture defined in the ISC, we retry ignoring the platform filter. In other cases (typo in the platform, or the image being single-arch) we just ignore the filter without warning. 

IMHO, if the user defines a platform in the ISC, we can assume that he wants that specific architecture and nothing else, and if that platform doesn't exist for that image for whatever reason, we should just fail for that image, print a warning, and continue with the next image in the ISC, instead of retrying without filter or ignoring the filter altogether.

Wdyt @aguidirh ?

internal/pkg/cli/executor.go
```diff
@@ -1171,45 +1177,51 @@ func (o *ExecutorSchema) CollectAll(ctx context.Context) (v2alpha1.CollectorSche
...
-		oImgs = excludeImages(oImgs, o.Config.Mirror.BlockedImages)
+		oImgs := excludeImages(operatorImgs.AllImages, o.Config.Mirror.BlockedImages)
 		oImgs = removeDuplicatedImages(oImgs, o.Opts.Function)
 		collectorSchema.TotalOperatorImages = len(oImgs)
 		o.Log.Debug(collecAllPrefix+"total operator images to %s %d ", o.Opts.Function, collectorSchema.TotalOperatorImages)
 		allRelatedImages = append(allRelatedImages, oImgs...)
 		collectorSchema.CopyImageSchemaMap = operatorImgs.CopyImageSchemaMap
 		collectorSchema.CatalogToFBCMap = operatorImgs.CatalogToFBCMap
+
+		// TODO: if images are shared between types, maps.Copy will override platform filters for
```
adolfo-ab: Maybe we can perform this merge in `removeDuplicatedImages`, which is called once per collector in `CollectAll`. That way we can have `Platforms` in `CopyImageSchema`, populated by each collector and we don't need the `PlatformFilter` map in `CollectorSchema`. I think this would simplify the PR, although it's similar to your first approach so maybe I'm missing something here.

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
...
+func determinePlatformSelection(opts *CopyOptions) (copy.ImageListSelection, []copy.InstancePlatformFilter, error) {
+	// Priority 1: Use InstancePlatforms if provided (set by batch worker from ImageSetConfig)
+	if len(opts.InstancePlatforms) > 0 {
+		platformsStr := strings.Join(opts.InstancePlatforms, ",")
+		return parseMultiArch(platformsStr)
+	}
+
+	// Priority 2: Fall back to MultiArch flag (for CLI direct usage or default "all")
+	if len(opts.MultiArch) > 0 && opts.All {
+		return copy.CopySystemImage, nil, fmt.Errorf("cannot use --all and --multi-arch flags together")
```
adolfo-ab: These error messages make it seem like `oc-mirror` had `--all` and `--multi-arch` flags, which might confuse the user, since these are skopeo CLI flags.

internal/pkg/release/local_stored_collector.go
adolfo-ab: It won't let me resolve this one, but feel free to do it yourself in you can or just move on as if it was

aguidirh:
@coderabbitai help

aguidirh:
@coderabbitai full review

internal/pkg/release/local_stored_collector.go
```diff
@@ -50,6 +50,28 @@ func (o LocalStorageCollector) destinationRegistry() string {
...
+		platformStrs := make([]string, len(o.Config.Mirror.Platform.Architectures))
+		for i, arch := range o.Config.Mirror.Platform.Architectures {
+			platformStrs[i] = "linux/" + arch
+		}
+		return &platformStrs
+	}
+
+	// No platforms specified - mirror all
+	return nil
+}
```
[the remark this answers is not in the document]
aguidirh: @coderabbitai is this still valid even after the latest changes?

internal/pkg/release/local_stored_collector.go
```diff
@@ -50,9 +50,21 @@ func (o LocalStorageCollector) destinationRegistry() string {
...
+
+func (o *LocalStorageCollector) ReleaseImageCollector(ctx context.Context) (v2alpha1.CollectorSchema, error) {
 	// we just care for 1 platform release, in order to read release images
 	o.Opts.MultiArch = "system"
+	if len(o.Config.Mirror.Platform.Platforms) > 0 {
+		// New Platforms field: download all arch payloads from the multi-arch release so that
+		// ensureReleaseInOCIFormat can extract image-references with manifest-list component digests.
+		o.Opts.MultiArch = "all"
+	}
```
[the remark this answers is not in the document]
aguidirh: @coderabbitai please close it since we're only keeping previous beavior untouched.

internal/pkg/batch/concurrent_chan_worker.go
```diff
@@ -144,8 +145,22 @@ func (o *ChannelConcurrentBatch) Worker(ctx context.Context, collectorSchema v2a
...
+								options.InstancePlatforms = platforms
+							}
+
 							err = o.Mirror.Run(timeoutCtx, img.Source, img.Destination, mirror.Mode(opts.Function), &options) //nolint:contextcheck
 
+							// TODO
+							// TBD not sure if I like this approach below
+							// Single-arch images don't support platform filtering; retry without it.
+							if err != nil && options.InstancePlatforms != nil && strings.Contains(err.Error(), "no instances found for platform") {
+								o.Log.Debug("Platform filtering not supported for %s, retrying without filter", img.Source)
```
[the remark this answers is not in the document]
aguidirh: @coderabbitai it was covered by the last commit, can you resolve it?

internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -188,3 +202,36 @@ func (o *ImageBlobGatherer) imageSignatureBlobs(ctx context.Context, in internal
...
+
+// platformAllowed reports whether platform is in allowedPlatforms.
+func platformAllowed(platform string, allowedPlatforms []string) bool {
+	for _, allowed := range allowedPlatforms {
+		if platform == allowed {
+			return true
+		}
+	}
+	return false
+}
```
r4f4:
```suggestion
```

internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -93,13 +97,23 @@ func (o *ImageBlobGatherer) multiArchBlobs(ctx context.Context, in internalImage
...
 	digests := manifestList.Instances()
 	for _, digest := range digests {
-		blobs.Insert(digest.String())
 		singleIn := in
 
 		singleIn.manifestBytes, singleIn.mimeType, err = o.ocmirrormanifest.ImageManifest(ctx, in.sourceCtx, in.imgRef, &digest)
 		if err != nil {
+			if strings.Contains(err.Error(), "manifest unknown") && len(in.allowedPlatforms) > 0 {
+				platform := platformForDigest(in.manifestBytes, digest.String())
+				if platform != "" && !platformAllowed(platform, in.allowedPlatforms) {
```
r4f4:
```suggestion
				if platform != "" && !slices.Contains(in.allowedPlatforms, platform) {
```

internal/pkg/release/cincinnati.go
```diff
@@ -285,6 +205,125 @@ func (o *CincinnatiSchema) GetReleaseReferenceImages(ctx context.Context) ([]v2a
...
+	case v2alpha1.TypeOKD:
+		return o.NewOKDClient()
+	default:
+		return fmt.Errorf("invalid platform type %v", ch.Type)
+	}
+}
+
+// warnEUSChannel emits a one-time warning when an EUS channel spans two or more minor versions.
+func (o *CincinnatiSchema) warnEUSChannel(ch v2alpha1.ReleaseChannel, flagReport *bool) {
+	if len(ch.MaxVersion) == 0 || len(ch.MinVersion) == 0 || *flagReport {
```
r4f4: `flagReport` pointer could be `nil` (even if it's not in the current code), so it's safer to check if it's a valid pointer before dereferencing it. Maybe this function can receive just a `bool` instead of a `*bool`?

internal/pkg/release/cincinnati.go
```diff
@@ -285,6 +205,125 @@ func (o *CincinnatiSchema) GetReleaseReferenceImages(ctx context.Context) ([]v2a
...
+	}
+	min, err := semver.Parse(ch.MinVersion)
+	if err != nil {
+		return
+	}
+	if strings.Contains(ch.Name, "eus") && (max.Minor-min.Minor) >= 2 {
+		o.Log.Warn("Extended Update Support (EUS) channel detected with minor version range >= 2\n" +
+			"\t\t\t\tPlease refer to the web console https://access.redhat.com/labs/ocpupgradegraph/update_path\n" +
+			"\t\t\t\tTo correctly determine the upgrade path for EUS releases")
+		*flagReport = true
```
r4f4: Instead of setting a pointer, why not return a `bool` to signify the warning was printed?

internal/pkg/release/cincinnati.go
r4f4: Another option is to just call `warEUSChannel` if `flagReport != nil && !*flagReport`

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
...
+		imageListSelection = copy.CopyAllImages
+	}
+
+	return imageListSelection, instancePlatforms, nil
+}
+
+// parseMultiArch parses the multi-arch option and returns the image list selection
+// and optional platform filters for specific OS/Architecture combinations.
+// Supports legacy options ('system', 'all', 'index-only') and new platform
+// specifications (e.g., 'linux/amd64,linux/arm64').
```
r4f4: I think this makes the variable too loaded? Can we keep the old string as-is for legacy behavior and use a new one for platforms?

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
...
+			platform = strings.TrimSpace(platform)
+			parts := strings.Split(platform, "/")
+			if len(parts) != 2 {
+				return copy.CopySystemImage, nil, fmt.Errorf("unknown multi-arch option %q. Choose one of the supported options: 'system', 'all', 'index-only', or a comma-separated platform list like 'linux/amd64,linux/arm64'", multiArch)
+			}
+			platforms = append(platforms, copy.InstancePlatformFilter{
+				OS:           parts[0],
+				Architecture: parts[1],
+			})
+		}
```
r4f4: This should be function we can reuse whenever we want to parse platform strings. That way, if we need to update the parsing to include other fields (e.g. `variant`), we can do it only once.

internal/pkg/release/cincinnati.go
```diff
@@ -159,109 +159,29 @@ func (o *CincinnatiSchema) GetReleaseReferenceImages(ctx context.Context) ([]v2a
...
-				continue
-			}
-			allImages = append(allImages, newDownloads...)
+		imgs, channelErrs := o.collectChannelImages(ctx, filterCopy.Channels, &flagReport)
+		allImages = append(allImages, imgs...)
+		errs = append(errs, channelErrs...)
+	} else {
+		// Deprecated Architectures field: original per-arch loop.
+		// Each arch results in a separate Cincinnati call and a separate release payload.
+		o.Log.Warn("platform.architectures is deprecated; use platform.platforms instead for sparse manifest platform filtering")
```
r4f4: This warning is going to be annoying for customers who didn't set any architectures/platforms in their ISC. They will see this message telling them the fields is deprecated but it was never set. It would be better if we could show this warning only when `architectures` was explicitly set.

internal/pkg/api/v2alpha1/type_platform.go
```diff
@@ -60,3 +61,50 @@ func (pt PlatformType) validate() error {
...
+func ConvertPlatformsToStringSlice(platforms []InstancePlatformFilter) []string {
+	if len(platforms) == 0 {
+		return nil
+	}
+
+	platformStrs := make([]string, 0, len(platforms))
+	for _, p := range platforms {
+		if p.Validate() != nil {
+			continue
+		}
```
r4f4: Silently dropping invalid platforms could lead to behaviors that are tricky to debug. What about making it so an `InstancePlatformFilter` is only created if the fields are valid to begin with?

tests/integration/sparse_manifest_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	Context("strict registry rejects sparse manifest", func() {
+		It("should fail when pushing sparse manifest to a standard registry without platforms:none", func() {
+			By("mirroring with amd64-only filter to a standard registry (no platforms:none config)")
+			result, err := runner.MirrorToMirror(ctx,
+				filepath.Join(iscDir, "sparse_manifest", "isc-additional-amd64.yaml"),
+				workDir, testRegistry.Endpoint(),
+				"--dest-tls-verify=false")
+			By("verifying oc-mirror exits with an error")
+			expectOcMirrorCommandFailure(result, err)
```
r4f4, line 247: Can we also check that we got the expected error and that this test is not passing just because of some unrelated error happening?

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
...
-// parseMultiArch
-func parseMultiArch(multiArch string) (copy.ImageListSelection, error) {
+// determinePlatformSelection determines the image list selection and platform filters
+// based on the provided copy options. It prioritizes InstancePlatforms (from ImageSetConfig)
+// over the MultiArch flag.
+func determinePlatformSelection(opts *CopyOptions) (copy.ImageListSelection, []copy.InstancePlatformFilter, error) {
+	// Priority 1: Use InstancePlatforms if provided (set by batch worker from ImageSetConfig)
+	if len(opts.InstancePlatforms) > 0 {
+		platformsStr := strings.Join(opts.InstancePlatforms, ",")
+		return parseMultiArch(platformsStr)
```
r4f4: This is a bit weird. You join the `InstancePlatforms` into a string just to call `parseMultiArch` which will split them again to create the `InstancePlatformFilter`.

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
...
 	case "index-only":
-		return copy.CopySpecificImages, nil
-	// We don't expose CopySpecificImages other than index-only above, because
-	// we currently don't provide an option to choose the images to copy. That
-	// could be added in the future.
+		return copy.CopySpecificImages, nil, nil
 	default:
-		return copy.CopySystemImage, fmt.Errorf("unknown multi-arch option %q. Choose one of the supported options: 'system', 'all', or 'index-only'", multiArch)
+		// Try to parse as comma-separated platform list (e.g., "linux/amd64,linux/arm64")
+		// Parse comma-separated platform list
```
r4f4: Nit: keep only one of these comment lines as they are saying the same thing.

internal/pkg/cli/executor.go
```diff
@@ -1396,6 +1401,27 @@ func mandatoryRegistries(opts *mirror.CopyOptions) map[string]struct{} {
...
+		combined = append(combined, v...)
+		seen := make(map[string]struct{}, len(combined))
+		deduped := make([]string, 0, len(combined))
+		for _, p := range combined {
+			if _, ok := seen[p]; !ok {
+				seen[p] = struct{}{}
+				deduped = append(deduped, p)
+			}
+		}
+		dst[k] = deduped
```
r4f4: This might not be performatic but it's simpler to read:
```suggestion
		for _, p := range v {
		    if !slices.Contains(dst[k], p) {
		        dst[k] = append(dst[k], p)
		    }
		}
```
Although the platform slices are not supposed to be big (just a handful of entries), so the impact is minimal.

internal/pkg/batch/concurrent_chan_worker.go
```diff
@@ -144,8 +145,24 @@ func (o *ChannelConcurrentBatch) Worker(ctx context.Context, collectorSchema v2a
...
+
 							err = o.Mirror.Run(timeoutCtx, img.Source, img.Destination, mirror.Mode(opts.Function), &options) //nolint:contextcheck
 
+							// "no instances found for platform" is only emitted by the copy library
+							// for manifest lists (multi-arch indexes) when none of the instances
+							// match the requested InstancePlatforms. The error itself confirms the
+							// source is a manifest list, so retrying without the filter would mirror
+							// all platforms and contradict the user's configuration. Return the
+							// original error so the user can correct their ImageSetConfiguration.
+							if err != nil && options.InstancePlatforms != nil && strings.Contains(err.Error(), "no instances found for platform") {
```
r4f4: I'm not sure we ever set `InstancePlatforms` to `nil` so the length check here checks if any `InstancePlatforms` were set.

```suggestion
							if err != nil && len(options.InstancePlatforms) > 0 && strings.Contains(err.Error(), "no instances found for platform") {
```

internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -188,3 +202,36 @@ func (o *ImageBlobGatherer) imageSignatureBlobs(ctx context.Context, in internal
...
+		} `json:"manifests"`
+	}
+	if err := json.Unmarshal(manifestBytes, &ml); err != nil {
+		return ""
+	}
+	for _, m := range ml.Manifests {
+		if m.Digest == digestStr {
+			return m.Platform.OS + "/" + m.Platform.Architecture
+		}
+	}
```
r4f4: Do we really need to do this adhoc parsing? Can't we do something like:
```suggestion
func platformForDigest(manifestList manifest.List, digest digest.Digest) string {
	if m, err := manifestList.Instance(digest); err != nil {
	    return m.ReadOnly.Platform.OS + "/" + m.ReadOnly.Platform.Architecture
	}

```

internal/pkg/release/cincinnati.go
```diff
@@ -285,6 +205,125 @@ func (o *CincinnatiSchema) GetReleaseReferenceImages(ctx context.Context) ([]v2a
...
+	}
+	// CLID-135: detect and log as early as possible
+	max, err := semver.Parse(ch.MaxVersion)
+	if err != nil {
+		return
+	}
+	min, err := semver.Parse(ch.MinVersion)
+	if err != nil {
+		return
+	}
```
r4f4: By changing from `MustParse` we are changing the behavior here from "fatal error if invalid channel version" to "ignore channel parsing failures without warning".

internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -188,3 +202,36 @@ func (o *ImageBlobGatherer) imageSignatureBlobs(ctx context.Context, in internal
 
 	return sigBlobs, nil
 }
+
+// platformForDigest returns the "os/arch" string for the manifest list entry
+// matching digestStr, or an empty string if not found.
+func platformForDigest(manifestBytes []byte, digestStr string) string {
```
r4f4: This should be unit-tested.

internal/pkg/mirror/mirror.go
```diff
@@ -323,22 +313,69 @@ func (o *Mirror) delete(ctx context.Context, image string, opts *CopyOptions) er
 	}, opts.RetryOpts)
 }
 
-// parseMultiArch
-func parseMultiArch(multiArch string) (copy.ImageListSelection, error) {
+// determinePlatformSelection determines the image list selection and platform filters
+// based on the provided copy options. It prioritizes InstancePlatforms (from ImageSetConfig)
+// over the MultiArch flag.
+func determinePlatformSelection(opts *CopyOptions) (copy.ImageListSelection, []copy.InstancePlatformFilter, error) {
```
r4f4, line 320: This should be unit-tested.

tests/integration/sparse_manifest_helpers_test.go
```diff
@@ -0,0 +1,102 @@
...
+		for _, exp := range expectedPlatforms {
+			if got == exp || strings.HasPrefix(got, exp+"/") {
+				matched = true
+				break
+			}
+		}
+		Expect(matched).To(BeTrue(),
+			"unexpected platform %q has a blob in the registry (should be filtered out)", got)
+	}
+	_ = missing // platforms in index but no blob — expected for sparse manifests
```
r4f4: Let's assert what's the expected value here instead of ignoring the variable. If no blobs are present I guess we want
```
Expect(missing).To(BeEmpty())
```

internal/pkg/additional/local_stored_collector.go
```diff
@@ -47,76 +47,81 @@ func (o LocalStorageCollector) destinationRegistry() string {
...
+			Type:        v2alpha1.TypeGeneric,
+		})
+		if platforms := v2alpha1.ConvertPlatformsToStringSlice(img.Platforms); len(platforms) > 0 {
+			platformFilters[origin] = platforms
 		}
+	}
+	cs := v2alpha1.CollectorSchema{AllImages: allImages}
+	if len(platformFilters) > 0 {
+		cs.PlatformFilters = platformFilters
+	}
```
r4f4: Nit:
```suggestion
	cs := v2alpha1.CollectorSchema{AllImages: allImages, PlatformFilters: platformFilters}
```

internal/pkg/archive/image-blob-gatherer.go
aguidirh: Implemented at commit [9ef1570a](https://github.com/aguidirh/oc-mirror/commit/9ef1570a)

internal/pkg/archive/image-blob-gatherer.go
aguidirh: Implemented at commit [9ef1570a](https://github.com/aguidirh/oc-mirror/commit/9ef1570a)

internal/pkg/release/cincinnati.go
aguidirh: Implemented at commit [63508dac](https://github.com/aguidirh/oc-mirror/commit/63508dac)

internal/pkg/release/cincinnati.go
aguidirh: Implemented at commit [63508dac](https://github.com/aguidirh/oc-mirror/commit/63508dac)

internal/pkg/release/cincinnati.go
aguidirh: Implemented at commit [13971025](https://github.com/aguidirh/oc-mirror/commit/13971025)

internal/pkg/release/cincinnati.go
aguidirh: Implemented at commit [9029e2c3](https://github.com/aguidirh/oc-mirror/commit/9029e2c3)

internal/pkg/mirror/mirror.go
aguidirh: Implemented at commit [92a7bfa8](https://github.com/aguidirh/oc-mirror/commit/92a7bfa8)

internal/pkg/mirror/mirror.go
aguidirh: Implemented at commit [048e6221](https://github.com/aguidirh/oc-mirror/commit/048e6221)

internal/pkg/api/v2alpha1/type_platform.go
aguidirh: Implemented at commit [ecefa2f0](https://github.com/aguidirh/oc-mirror/commit/ecefa2f0)

tests/integration/sparse_manifest_test.go
aguidirh: Implemented at commit [5b9b5de7](https://github.com/aguidirh/oc-mirror/commit/5b9b5de7)

internal/pkg/batch/concurrent_chan_worker.go
aguidirh: Implemented at commit [7853b179](https://github.com/aguidirh/oc-mirror/commit/7853b179)

internal/pkg/archive/image-blob-gatherer.go
aguidirh: Implemented at commits [d297df0b](https://github.com/aguidirh/oc-mirror/commit/d297df0b) and [79582fff](https://github.com/aguidirh/oc-mirror/commit/79582fff)

internal/pkg/additional/local_stored_collector.go
aguidirh: Implemented at commit [5da3f97c](https://github.com/aguidirh/oc-mirror/commit/5da3f97c)

internal/pkg/mirror/mirror.go
aguidirh: Already resolved — the duplicate comment block was removed when platform parsing was moved out of `parseMultiArch` into `determinePlatformSelection` (commit [048e6221](https://github.com/aguidirh/oc-mirror/commit/048e6221)).

internal/pkg/archive/image-blob-gatherer.go
aguidirh: Implemented at commit [79582fff](https://github.com/aguidirh/oc-mirror/commit/79582fff)

tests/integration/sparse_manifest_helpers_test.go
aguidirh: Implemented at commit [c5046987](https://github.com/aguidirh/oc-mirror/commit/c5046987) — asserted that every platform in `missing` is one that was not requested (`Expect(expectedPlatforms).NotTo(ContainElement(m))`). `BeEmpty()` cannot be used here since `missing` is legitimately non-empty for sparse manifests, and `NotTo(BeEmpty())` would fail for images that have exactly the requested set of platforms.

internal/pkg/mirror/mirror.go
aguidirh: Implemented at commit [d15b7f18](https://github.com/aguidirh/oc-mirror/commit/d15b7f18)

internal/pkg/cli/executor.go
aguidirh: Implemented at commit [adf8d591](https://github.com/aguidirh/oc-mirror/commit/adf8d591)

internal/pkg/mirror/mirror.go
aguidirh: This was implemented the same way as in `skopeo` #[2874](https://github.com/podman-container-tools/skopeo/pull/2874), but I extracted the piece of code related to specific platforms to `determinePlatformSelection` and kept legacy `parseMultiArch` as it was.

internal/pkg/cli/executor.go
aguidirh: No change taken — the PlatformFilters map approach was kept. The underlying concern (a maps.Copy that would silently overwrite platform filters on key collision) was fixed in [d5aaf141](https://github.com/aguidirh/oc-mirror/commit/d5aaf141) by replacing it with mergePlatformFilters, which appens platform slices per origin instead of overwriting.

internal/pkg/mirror/mirror.go
aguidirh: Implemented at [9769798f R328](https://github.com/aguidirh/oc-mirror/commit/9769798f#diff-68ec5ee426eb2daa792f69acfc2b190e76b2fd184fbb46f6edd3107cc193a017R328)

aguidirh:
> Currently, if I understand correctly, if a user writes a platform that doesn't exist in the ISC, we always continue the mirroring process.
> 
> In the case that the image is a manifest list but doesn't contain the architecture defined in the ISC, we retry ignoring the platform filter. In other cases (typo in the platform, or the image being single-arch) we just ignore the filter without warning.
> 
> IMHO, if the user defines a platform in the ISC, we can assume that he wants that specific architecture and nothing else, and if that platform doesn't exist for that image for whatever reason, we should just fail for that image, print a warning, and continue with the next image in the ISC, instead of retrying without filter or ignoring the filter altogether.
> 
> Wdyt @aguidirh ?

This was addressed in the latest code.

internal/pkg/helm/local_stored_collector.go
```diff
@@ -226,22 +248,34 @@ func GetDefaultChartDownloader() chartDownloader {
...
 		imgs, err := getImages(chart.Path, chart.ImagePaths...)
 		if err != nil {
 			errs = append(errs, err)
 		}
 
 		if len(imgs) > 0 {
 			allHelmImages = append(allHelmImages, imgs...)
+			if len(platforms) > 0 {
+				for _, img := range imgs {
+					platformFilters[img.Image] = append(platformFilters[img.Image], platforms...)
```
r4f4: It seems the `getImages` -> `findImages` call chain calls `parseRef` but [discards the result](https://github.com/openshift/oc-mirror/blob/main/internal/pkg/helm/local_stored_collector.go#L531). If we mix image references (raw vs parsed), we won't properly merge platform filters for the same image. Not sure if that can cause issues but it shows a disadvantage of using the shared map indexed by image ref approach.

internal/pkg/archive/image-blob-gatherer.go
```diff
@@ -93,13 +97,23 @@ func (o *ImageBlobGatherer) multiArchBlobs(ctx context.Context, in internalImage
...
 		if err != nil {
+			if strings.Contains(err.Error(), "manifest unknown") && len(in.allowedPlatforms) > 0 {
+				platform := platformForDigest(manifestList, digest)
+				if platform != "" && !slices.Contains(in.allowedPlatforms, platform) {
+					// Platform was intentionally not mirrored (sparse manifest filtering).
+					// Do not add its digest to blobs — the blob file does not exist locally.
+					o.log.Debug("Skipping absent platform %q for digest %s (not in allowed platforms)", platform, digest.String())
+					continue
+				}
+			}
```
r4f4: Let's say I just want the `amd64` arch and not the `arm64` one. In that case, errors other than `manifest unknown` (e.g, network errors) will still cause `oc-mirror` to fail on `arm64` which I do not want. Shouldn't this loop be:
```go
for _, digest := range digests {
    if platform := platformForDigest(manifestList, digest); platform != "" && !slices.Contains(in.allowedPlatforms, platform) {
        // Platform was intentionally not mirrored (sparse manifest filtering).
		// Do not add its digest to blobs — the blob file does not exist locally.
		o.log.Debug("Skipping absent platform %q for digest %s (not in allowed platforms)", platform, digest.String())
        continue
    }
    single := in
    singleIn.manifestBytes, singleIn.mimeType, err = o.ocmirrormanifest.ImageManifest(ctx, in.sourceCtx, in.imgRef, &digest)
    if err != nil {
        return nil, err
    }
// [...]
}
```

internal/pkg/operator/filtered_collector.go
```diff
@@ -70,13 +71,36 @@ func (o *FilterCollector) OperatorImageCollector(ctx context.Context) (v2alpha1.
...
+		// The same bundle appearing in two catalogs with different platforms gets both appended.
+		if platforms, err := v2alpha1.ConvertPlatformsToStringSlice(op.Platforms); err != nil {
+			allErrs = append(allErrs, fmt.Errorf("invalid platform for operator catalog %q: %w", op.Catalog, err))
+		} else if len(platforms) > 0 {
+			for _, imgs := range localRelatedImages {
+				for _, img := range imgs {
+					if ref, parseErr := image.ParseRef(img.Image); parseErr == nil {
+						origin := ref.ReferenceWithTransport
+						collectorSchema.PlatformFilters[origin] = append(collectorSchema.PlatformFilters[origin], platforms...)
+					}
```
r4f4, line 99: Nit: if `parseErr != nil` then we will silently not set platform filters for the image. I'm not sure if it would be better to add a debug log for this or just use the raw image ref:

```suggestion
					} else {
					    collectorSchema.PlatformFilters[img.Image] = append(collectorSchema.PlatformFilters[img.Image], platforms...)
					}
```

internal/pkg/helm/local_stored_collector.go
```diff
@@ -226,22 +248,34 @@ func GetDefaultChartDownloader() chartDownloader {
...
+			errs = append(errs, fmt.Errorf("invalid platform for chart %q: %w", chart.Name, err))
+			continue
+		}
+
 		imgs, err := getImages(chart.Path, chart.ImagePaths...)
 		if err != nil {
 			errs = append(errs, err)
 		}
 
 		if len(imgs) > 0 {
```
r4f4: Nit:
```suggestion
		if len(imgs) == 0 {
		    continue
		}
```

internal/pkg/helm/local_stored_collector.go
```diff
@@ -78,123 +79,144 @@ func WithV1Tags(o CollectorInterface) CollectorInterface {
...
+			if err != nil {
+				errs = append(errs, fmt.Errorf("invalid platform for chart %q: %w", chart.Name, err))
+				continue
+			}
+			chartImgs, err := getImages(path, chart.ImagePaths...)
+			if err != nil {
+				errs = append(errs, err)
+			}
+			allHelmImages = append(allHelmImages, chartImgs...)
+			for _, img := range chartImgs {
```
r4f4: Should we do:
```suggestion
            if len(platforms) > 0 {
			    for _, img := range chartImgs {
```

internal/pkg/helm/local_stored_collector.go
```diff
@@ -78,123 +79,144 @@ func WithV1Tags(o CollectorInterface) CollectorInterface {
...
```
r4f4: Should we do
```suggestion
            if len(platforms) > 0 {
			    for _, img := range chartImgs {
```

internal/pkg/archive/image-blob-gatherer_test.go
```diff
@@ -328,6 +330,32 @@ func TestImageBlobGatherer_ImageSourceError(t *testing.T) {
 	}
 
 	gatherer := NewImageBlobGatherer(&opts, clog.New("trace"))
-	_, err = gatherer.GatherBlobs(ctx, consts.DockerProtocol+u.Host+"/bad-test:latest")
+	_, err = gatherer.GatherBlobs(ctx, consts.DockerProtocol+u.Host+"/bad-test:latest", nil)
```
r4f4, line 379: We're missing unit tests for `GatherBlobs` where `allowedPlatforms` is not `nil`.

internal/pkg/release/local_stored_collector.go
```diff
@@ -68,24 +79,37 @@ func (o *LocalStorageCollector) ReleaseImageCollector(ctx context.Context) ([]v2
...
+	return cs, nil
+}
+
+// buildReleasePlatformFilters builds a map from image origin to platform filters,
+// excluding Cincinnati graph images which must always be copied whole.
+func buildReleasePlatformFilters(allImages []v2alpha1.CopyImageSchema, platforms []string) map[string][]string {
+	filters := make(map[string][]string, len(allImages))
+	for _, img := range allImages {
+		if img.Type != v2alpha1.TypeCincinnatiGraph {
```
r4f4, line 111: It would be good to have a unit test, not to verify this function, but more so to guarantee that this requirement is upheld.

internal/pkg/api/v2alpha1/type_internal.go
```diff
@@ -192,6 +192,9 @@ type CollectorSchema struct {
 	AllImages             []CopyImageSchema
 	CopyImageSchemaMap    CopyImageSchemaMap
 	CatalogToFBCMap       map[string]CatalogFilterResult // key is the mirror.operator.catalog
+	// PlatformFilters maps image origin to its platform filter list ("os/arch" strings).
+	// Populated by collectors; consumed by the batch worker to set InstancePlatforms per image.
+	PlatformFilters map[string][]string
```
r4f4:
```suggestion
	PlatformFilters map[string][]InstancePlatformFilter
```
Would make the intention more clear and avoid marlformed strings being added to the filters map.

internal/pkg/operator/filtered_collector.go
```diff
@@ -70,13 +71,36 @@ func (o *FilterCollector) OperatorImageCollector(ctx context.Context) (v2alpha1.
...
+		// Append this catalog's platform filters into PlatformFilters using localRelatedImages.
+		// The same bundle appearing in two catalogs with different platforms gets both appended.
+		if platforms, err := v2alpha1.ConvertPlatformsToStringSlice(op.Platforms); err != nil {
+			allErrs = append(allErrs, fmt.Errorf("invalid platform for operator catalog %q: %w", op.Catalog, err))
+		} else if len(platforms) > 0 {
+			for _, imgs := range localRelatedImages {
+				for _, img := range imgs {
+					if ref, parseErr := image.ParseRef(img.Image); parseErr == nil {
+						origin := ref.ReferenceWithTransport
+						collectorSchema.PlatformFilters[origin] = append(collectorSchema.PlatformFilters[origin], platforms...)
```
r4f4: Since we always append, we could be duplicating platforms for shared images. I think it should be fine because we might dedup them during merge later on. Maybe worth a comment to make it clear it's intentional.

internal/pkg/archive/image-blob-gatherer.go
aguidirh: Implemented at [8eb027df](https://github.com/aguidirh/oc-mirror/commit/8eb027df)

internal/pkg/operator/filtered_collector.go
aguidirh: Implemented at [3920bcf5](https://github.com/aguidirh/oc-mirror/commit/3920bcf5)

internal/pkg/helm/local_stored_collector.go
aguidirh: Implemented at [b7d42727](https://github.com/aguidirh/oc-mirror/commit/b7d42727)

internal/pkg/helm/local_stored_collector.go
aguidirh: Implemented at [82031d8e](https://github.com/aguidirh/oc-mirror/commit/82031d8e)

internal/pkg/archive/image-blob-gatherer_test.go
aguidirh: Implemented at [39cabbc0](https://github.com/aguidirh/oc-mirror/commit/39cabbc0)

internal/pkg/release/local_stored_collector.go
aguidirh: Implemented at [c2295789](https://github.com/aguidirh/oc-mirror/commit/c2295789)

internal/pkg/helm/local_stored_collector.go
aguidirh: Implemented at [b7d42727](https://github.com/aguidirh/oc-mirror/commit/b7d42727)

internal/pkg/operator/filtered_collector.go
aguidirh: Implemented at [974e9660](https://github.com/aguidirh/oc-mirror/commit/974e9660)

internal/pkg/api/v2alpha1/type_internal.go
aguidirh: Implemented at [8776ac20](https://github.com/aguidirh/oc-mirror/commit/8776ac20)

internal/pkg/helm/local_stored_collector.go
aguidirh: Implemented at [508cc96b](https://github.com/aguidirh/oc-mirror/commit/508cc96b)

r4f4 (commented):
/lgtm

PR #1444, opened 2026-06-16
CLID-657: Add warning when using different version for m2d and d2m

adolfo-ab:
# Description
When users do mirror-to-disk using a given version of oc-mirror, and then disk-to-mirror using a different version, they might find unexpected behaviors if major changes have occurred between versions. 

With this PR I propose saving the version metadata during m2d (in `working-dir/internal/oc-mirror-version.json`), and then compare it with the version of the binary during d2m. If there is a mismatch or the version metadata file cannot be found, we show a non-fatal warning. For now we compare the `GitVersion` and `GitCommit`, but when we are removed from the payload we can consider using the actual version of oc-mirror.

Github / Jira issue: 
[CLID-657](https://redhat.atlassian.net/browse/CLID-657)

## Type of change

# How Has This Been Tested?
- Running m2d from main, then d2m from this branch -> missing file warning
- Running m2d from this branch, change the version metadata file, then run d2m -> version mismatch warning
- Same thing with a malformed json -> malformed file warning

## Summary by CodeRabbit

* **New Features**
  * Added persistent version metadata to mirror-to-disk workflows to improve build provenance tracking.
  * When converting disk mirrors back to local registries, the tool now validates archived metadata against the current build (version/commit).
* **Bug Fixes**
  * Metadata write failures during mirror-to-disk now emit a warning instead of stopping the command.
* **Tests**
  * Added coverage for metadata write/read round-trips and validation behavior, including missing, mismatch, corrupt, and error scenarios.

### DIFF
--- a/internal/pkg/cli/executor.go
+++ b/internal/pkg/cli/executor.go
@@ -915,6 +915,12 @@ func (o *ExecutorSchema) RunMirrorToDisk(cmd *cobra.Command, args []string) erro
 
 	o.createConfigsWithPinnedCatalogs()
 
+	if err := version.WriteVersionMetadata(o.Opts.Global.WorkingDir, version.Get()); err != nil {
+		o.Log.Warn("unable to write oc-mirror version metadata: %v", err)
+	} else {
+		o.Log.Debug("wrote version metadata: %s", version.Get().GitVersion)
+	}
+
 	o.Log.Info(emoji.Package + " Preparing the tarball archive...")
 	return o.MirrorArchiver.BuildArchive(cmd.Context(), copiedSchema.AllImages)
 }
@@ -1008,6 +1014,12 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
 	if err := o.MirrorUnArchiver.Unarchive(); err != nil {
 		o.Log.Error(" %v ", err)
 		return err
+	}
+
+	if warning := version.CheckVersionMetadata(o.Opts.Global.WorkingDir, version.Get()); warning != "" {
+		o.Log.Warn(emoji.Warning+"  %s", warning)
+	} else {
+		o.Log.Debug("version metadata check passed: %s", version.Get().GitVersion)
 	}
 
 	// start the local storage registry
--- a/internal/pkg/version/metadata.go
+++ b/internal/pkg/version/metadata.go
@@ -1,0 +1,85 @@
+package version
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/oc-mirror/v2/internal/pkg/config"
+	"github.com/openshift/oc-mirror/v2/internal/pkg/parser"
+)
+
+const versionMismatchWarn = "It is recommended to use the same oc-mirror version for both mirror-to-disk and disk-to-mirror"
+
+// VersionMetadata records which oc-mirror build produced a mirror-to-disk archive.
+type VersionMetadata struct {
+	GitVersion string `json:"gitVersion"`
+	GitCommit  string `json:"gitCommit"`
+}
+
+// NewVersionMetadata builds metadata from the current binary version info.
+func NewVersionMetadata(info Info) VersionMetadata {
+	return VersionMetadata{
+		GitVersion: info.GitVersion,
+		GitCommit:  info.GitCommit,
+	}
+}
+
+// WriteVersionMetadata writes version metadata under workingDir/internal/.
+func WriteVersionMetadata(workingDir string, info Info) error {
+	internalDir := filepath.Join(workingDir, config.InternalDir)
+	if err := os.MkdirAll(internalDir, 0o755); err != nil {
+		return fmt.Errorf("unable to create internal directory: %w", err)
+	}
+
+	meta := NewVersionMetadata(info)
+	data, err := json.MarshalIndent(meta, "", "  ")
+	if err != nil {
+		return fmt.Errorf("unable to marshal version metadata: %w", err)
+	}
+
+	path := filepath.Join(workingDir, config.VersionMetadataPath)
+	if err := os.WriteFile(path, data, 0o644); err != nil { //nolint:gosec // version metadata is not sensitive info
+		return fmt.Errorf("unable to write version metadata: %w", err)
+	}
+
+	return nil
+}
+
+// ReadVersionMetadata reads version metadata from the given file path.
+func ReadVersionMetadata(path string) (VersionMetadata, error) {
+	meta, err := parser.ParseJsonFile[VersionMetadata](path)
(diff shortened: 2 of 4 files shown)


internal/pkg/version/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	meta := NewVersionMetadata(info)
+	data, err := json.MarshalIndent(meta, "", "  ")
+	if err != nil {
+		return fmt.Errorf("unable to marshal version metadata: %w", err)
+	}
+
+	path := filepath.Join(workingDir, config.VersionMetadataPath)
+	if err := os.WriteFile(path, data, 0o600); err != nil {
```
aguidirh: The users running `m2d` and `d2m` could be different, so `0o644` is needed which adds read permissions for group and others, otherwise `d2m` will fail with `permissions denied` when reading the file.

internal/pkg/version/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/openshift/oc-mirror/v2/internal/pkg/config"
+)
+
+// VersionMetadata records which oc-mirror build produced a mirror-to-disk archive.
+type VersionMetadata struct {
+	GitVersion string `json:"gitVersion"`
+	GitCommit  string `json:"gitCommit"`
+	CreatedAt  string `json:"createdAt"`
```
aguidirh: I did not find `CreatedAt` being used in the code. Is this only for troubleshooting reasons? 

The creation date is already in the file info so it could be used from there for troubleshooting as well.

aguidirh (changes requested):
Thanks for the PR Adolfo.

I did few suggestions, please let me know if you have questions about them.

internal/pkg/version/metadata.go
adolfo-ab: Fixed

internal/pkg/version/metadata.go
adolfo-ab: Added for troubleshooting but you're right it's not really needed, removed it.

aguidirh (approved):

internal/pkg/version/metadata.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	meta := NewVersionMetadata(info)
+	data, err := json.MarshalIndent(meta, "", "  ")
+	if err != nil {
+		return fmt.Errorf("unable to marshal version metadata: %w", err)
+	}
+
+	path := filepath.Join(workingDir, config.VersionMetadataPath)
+	if err := os.WriteFile(path, data, 0o644); err != nil { // nolint:gosec
```
r4f4: Nit: I think it's helpful to add a justification every time we silence the linter. No need to change this unless you have something else to change.

```suggestion
	if err := os.WriteFile(path, data, 0o644); err != nil { //nolint:gosec // version metadata is not sensitive info
```

internal/pkg/version/metadata.go
```diff
@@ -0,0 +1,87 @@
...
+
+// CheckVersionMetadata returns a warning message when the archive metadata is missing,
+// unreadable, or does not match the current binary. An empty string means no warning.
+func CheckVersionMetadata(workingDir string, current Info) string {
+	path := filepath.Join(workingDir, config.VersionMetadataPath)
+
+	meta, err := ReadVersionMetadata(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return "No oc-mirror version metadata found in this archive. Mirror-to-disk may have been run with an older oc-mirror release that did not record version information. We recommend using the same oc-mirror version for both mirror-to-disk and disk-to-mirror."
```
r4f4: The "We recommend using the same oc-mirror version for both mirror-to-disk and disk-to-mirror" part is being repeated below. Maybe we could extract and reuse it:
```
const versionMismatchWarn string = "Using different oc-mirror versions for mirror-to-disk and disk-to-mirror may cause compatibility issues"
```

internal/pkg/cli/executor.go
```diff
@@ -1010,6 +1014,10 @@ func (o *ExecutorSchema) RunDiskToMirror(cmd *cobra.Command, args []string) erro
 		return err
 	}
 
+	if warning := version.CheckVersionMetadata(o.Opts.Global.WorkingDir, version.Get()); warning != "" {
+		o.Log.Warn(emoji.Warning+"  %s", warning)
+	}
```
r4f4, line 1023:
```suggestion
	} else {
	    o.Log.Debug("version metadata check passed: %s", version.Get().GitVersion)
	}
```
So we can quickly check if the oc-mirror used is doing version checks in debug logs.

internal/pkg/cli/executor.go
```diff
@@ -915,6 +915,10 @@ func (o *ExecutorSchema) RunMirrorToDisk(cmd *cobra.Command, args []string) erro
 
 	o.createConfigsWithPinnedCatalogs()
 
+	if err := version.WriteVersionMetadata(o.Opts.Global.WorkingDir, version.Get()); err != nil {
+		o.Log.Warn("unable to write oc-mirror version metadata: %v", err)
+	}
```
r4f4, line 922:
```suggestion
	} else {
	    o.Log.Debug("wrote version metadata: %s", version.Get().GitVersion)
	}
```

internal/pkg/version/metadata.go
```diff
@@ -0,0 +1,87 @@
...
+func ReadVersionMetadata(path string) (VersionMetadata, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return VersionMetadata{}, fmt.Errorf("unable to read version metadata: %w", err)
+	}
+
+	var meta VersionMetadata
+	if err := json.Unmarshal(data, &meta); err != nil {
+		return VersionMetadata{}, fmt.Errorf("unable to parse version metadata: %w", err)
+	}
```
r4f4:
```suggestion
	meta, err := parser.ParseJsonFile[VersionMetadata](path)
	if  err != nil {
		return VersionMetadata{}, fmt.Errorf("version metadata: %w", err)
	}
```

adolfo-ab: Done

internal/pkg/cli/executor.go
adolfo-ab: Done

internal/pkg/cli/executor.go
adolfo-ab: Done

internal/pkg/version/metadata.go
adolfo-ab: Done

internal/pkg/version/metadata.go
adolfo-ab: Done

r4f4 (commented):
/lgtm

PR #1454, opened 2026-06-25
OCPBUGS-90135: retry blob copy on dropped connections

zaneb:
When downloading blobs from CDN, dropped connections result in io.ErrUnexpectedEOF being propagated up to the retry wrapper. Similarly, CDN connectivity issues can produce TLS alert errors. Neither was classified as retryable, making --retry-times a no-op for these common failure modes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling for mirror operations so some transient transfer and TLS errors are retried automatically instead of failing immediately.
  * Added support for retrying unexpected end-of-file errors and certain TLS alert failures, helping reduce occasional connection-related disruptions.

### DIFF
--- a/internal/pkg/mirror/mirror.go
+++ b/internal/pkg/mirror/mirror.go
@@ -2,6 +2,7 @@ package mirror
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


aguidirh:
Hi @zaneb,

Is this bug fix related to https://github.com/podman-container-tools/container-libs/pull/845?

zaneb:
No, that doesn't appear to be related.

PR #1468, opened 2026-07-07
CLID-628: Extend operator test with rebuild/cluster resource checks

adolfo-ab:
# Description

Automates the manual QE test case for OCPBUGS-56718 (index not being rebuilt on mirror). Extends the existing operator with pinned version test, which already mirrors a single package out of test-catalog-latest's 3 packages, to also verify that the rebuilt catalog excludes the non-selected packages, and that the generated IDMS and CatalogSource cluster resources correctly reference the mirrored catalog in the target registry.

Github / Jira issue: 
https://redhat.atlassian.net/browse/CLID-628

## Type of change
- [X] Code Improvements (Refactoring, Performance, CI upgrades, etc)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved operator catalog rebuild handling for pinned bundle versions, ensuring generated catalogs only include requested packages and reference rebuilt images in the target registry.
* **Tests**
  * Expanded the pinned-operator-version integration test to verify catalog rebuild filtering, generated cluster resource correctness (including IDMS content), and that generated `CatalogSource` manifests reference target-registry rebuilt catalog images.


r4f4 (commented):
/lgtm

tests/integration/operators_test.go
```diff
@@ -100,3 +118,84 @@ var _ = Describe("operators", func() {
 		})
 	})
 })
+
+// expectCatalogContainsOnlyExpectedPackages verifies that the rebuilt catalog in the registry
+// contains configs for exactly the packages selected in the ISC, and no others. This guards
+// against a regression where oc-mirror copies the source catalog unfiltered instead of rebuilding
+// a filtered index (OCPBUGS-56718).
+func expectCatalogContainsOnlyExpectedPackages(ctx context.Context, reg registry.Registry, iscPath string) {
```
aguidirh, line 126: I see usually we put `func expect*` on the helper, but for `expectCatalogContainsOnlyExpectedPackages` and `expectCorrectCatalogSourceYAMLs` we kept it here, only curious to know why.

aguidirh (commented):
/lgtm

tests/integration/operators_test.go
adolfo-ab: The helpers file is getting too crowded, so when a function is only used in a single test file, I'm trying to put it there, to avoid cluttering the helpers file unnecessarily.

aguidirh: Got it, thanks for the explanation, it makes a lot of sense.

nidangavali (approved):

PR #1496, opened 2026-08-12
CLID-703: integration test to check d2m works without network access

adolfo-ab:
# Description
Adds integration test to check d2m works without network access, replaces manual test OCP-75219

Github / Jira issue: 

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

## Expected Outcome
Please describe the outcome expected from the tests.

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage confirming local archive mirroring succeeds without outbound network access.
  * Verified mirroring remains successful when external HTTP and HTTPS connections are unavailable.

### DIFF
--- a/tests/integration/network_isolation_test.go
+++ b/tests/integration/network_isolation_test.go
@@ -1,0 +1,56 @@
+package integration_test
+
+import (
+	"os"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/oc-mirror/tests/integration/pkg/ocmirror"
+)
+
+// OCP-75219 - diskToMirror should not require network access to source registries.
+//
+// The d2m workflow by design reads all image sources from the local cache (localhost); it never
+// contacts source registries. As an extra safety net, HTTP_PROXY/HTTPS_PROXY are pointed at an
+// unreachable address so any accidental outbound call fails immediately. This works because Go's
+// net/http stack (used by both oc-mirror and containers/image) universally honours proxy env vars
+// via http.ProxyFromEnvironment. Retries are capped at 1 (--retry-times=1) to avoid exponential
+// backoff delays in case a call does hit the blocked proxy.
+var _ = Describe("diskToMirror without network access", func() {
+	var workDir string
+
+	BeforeEach(func() {
+		workDir = setupWorkDir()
+	})
+
+	AfterEach(func() {
+		cleanupWorkDir(workDir)
+	})
+
+	It("should mirror successfully from a local archive when outbound network access is blocked", func() {
+		iscPath := filepath.Join(iscDir, "secure_policy", "isc-operators-only.yaml")
+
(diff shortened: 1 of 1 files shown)


tests/integration/network_isolation_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+		By("removing the local cache to simulate a clean disk-to-mirror environment")
+		Expect(os.RemoveAll(cacheDir)).To(Succeed())
+
+		By("running diskToMirror with outbound network access blocked")
+		offlineRunner := ocmirror.NewRunner(os.Getenv("OC_MIRROR_BINARY")).
+			WithEnv([]string{
+				"HTTP_PROXY=http://127.0.0.1:1",
+				"HTTPS_PROXY=http://127.0.0.1:1",
+				"NO_PROXY=localhost,127.0.0.1",
```
aguidirh, line 47: If there are client calls that are not using proxy env vars (inside of oc-mirror or containers/image), would them be able to reach the internet ?

tests/integration/network_isolation_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/openshift/oc-mirror/tests/integration/pkg/ocmirror"
+)
+
+// OCP-75219 - diskToMirror should not require network access to source registries.
+//
+// Outbound network access is simulated as blocked by pointing HTTP_PROXY/HTTPS_PROXY at an
+// unreachable local address, while excluding localhost (used by the destination registry and
+// oc-mirror's own local cache) via NO_PROXY. Any accidental external call would fail fast with a
+// connection error, so a successful run proves diskToMirror only relied on the local archive and
+// the destination registry.
```
aguidirh: Let's say containers/image retries with backoff, would it cause slowness on our suite test by the repeated attempts of the container/image against the refused proxy? If this is true, can we pass to the containers/image retry-times with only 1 attempt?

aguidirh (commented):
I added few comments, please let me know what you think about them.

tests/integration/network_isolation_test.go
adolfo-ab: I verified that all v2 HTTP clients either explicitly set Proxy: http.ProxyFromEnvironment or use Go's http.DefaultTransport, which does the same. containers/image also goes through Go's net/http stack, so proxy env vars are respected.

tests/integration/network_isolation_test.go
adolfo-ab: Good catch. The default is 5 retries with exponential backoff, and ECONNREFUSED is in the retryable error list, so in the worst case a single failed call could stall for ~62 seconds. I've added `--retry-times=1` to the d2m invocation to eliminate that risk. Since d2m shouldn't be making any outbound calls in the first place, this is purely defensive.

aguidirh (commented):
/lgtm

PR #1499, opened 2026-08-12
OCPBUGS-78497: reject uppercase registry hostnames in docker destinations

shishind:
## Summary
- Reject `docker://` destinations whose registry hostname contains uppercase characters during Validate (mirror and delete), before any push work starts.
- This prevents mirroring content that cluster nodes cannot pull, because CRI-O rejects uppercase registry hostnames.
- Adds unit coverage for the helper and CLI validate paths; IPv4/IPv6 literals remain allowed.

Fixes https://issues.redhat.com/browse/OCPBUGS-78497

## Test plan
- [ ] `go test ./internal/pkg/image/ -run 'TestValidateDockerDestinationRegistry|TestRegistryHostname'`
- [ ] `CGO_ENABLED=0 go test -tags "json1 exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp" ./internal/pkg/cli/ -run 'TestExecutorValidate$|TestExecutorValidateDelete'`
- [ ] Manually confirm `oc-mirror ... docker://REGISTRY.EXAMPLE.COM:5000` fails with a clear lowercase suggestion
- [ ] Manually confirm `docker://registry.example.com:5000` still validates successfully

## Summary by CodeRabbit

* **Bug Fixes**
  * Docker destinations with uppercase or mixed-case registry hostnames are now rejected with a clear compatibility error.
  * Valid lowercase hostnames, IPv4/IPv6 addresses, namespaced destinations, and non-Docker destinations continue to be accepted.
  * Validation now rejects malformed, missing, or invalid bracketed registry hostnames, including destinations with leading whitespace.

* **Tests**
  * Expanded coverage for valid and invalid Docker destination formats, including ports, bracketed IPv6 addresses, leading whitespace, and malformed hostnames.

### DIFF
--- a/internal/pkg/image/registry_hostname.go
+++ b/internal/pkg/image/registry_hostname.go
@@ -1,0 +1,68 @@
+package image
+
+import (
+	"fmt"
+	"net"
+	"strings"
+	"unicode"
+
+	"github.com/openshift/oc-mirror/v2/internal/pkg/consts"
+)
+
+// ValidateDockerDestinationRegistry validates a docker:// destination argument.
+// CRI-O rejects uppercase letters in registry hostnames, so oc-mirror fails early
+// rather than producing mirrored content that cluster nodes cannot pull (OCPBUGS-78497).
+func ValidateDockerDestinationRegistry(dest string) error {
+	if !strings.HasPrefix(dest, consts.DockerProtocol) {
+		return nil
+	}
+
+	ref := strings.TrimPrefix(dest, consts.DockerProtocol)
+	hostname := registryHostname(ref)
+	if hostname == "" {
+		return fmt.Errorf("destination registry hostname is empty")
+	}
+
+	// IP literals (IPv4 / IPv6) are not subject to DNS hostname case rules.
+	// IPv6 hex digits may be uppercase and remain valid.
+	if hostForIP := hostname; strings.HasPrefix(hostForIP, "[") && strings.HasSuffix(hostForIP, "]") {
+		hostForIP = hostForIP[1 : len(hostForIP)-1]
+		if net.ParseIP(hostForIP) != nil {
+			return nil
+		}
+	} else if net.ParseIP(hostname) != nil {
+		return nil
+	}
+
+	for _, r := range hostname {
+		if unicode.IsUpper(r) {
+			return fmt.Errorf("destination registry hostname %q contains uppercase characters; use lowercase (e.g. %q) because CRI-O rejects uppercase registry hostnames", hostname, strings.ToLower(hostname))
+		}
+	}
+	return nil
+}
+
+// registryHostname returns the host (without path) from a docker destination
+// reference such as "registry.example.com:5000/ns" or "[::1]:5000/ns".
+func registryHostname(ref string) string {
+	hostPort := ref
+	if i := strings.Index(ref, "/"); i >= 0 {
+		hostPort = ref[:i]
+	}
+	if hostPort == "" {
+		return ""
+	}
+
+	if strings.HasPrefix(hostPort, "[") {
+		end := strings.Index(hostPort, "]")
+		if end < 0 {
+			return hostPort
+		}
+		return hostPort[:end+1]
+	}
+
+	if host, _, err := net.SplitHostPort(hostPort); err == nil {
+		return host
+	}
+	return hostPort
+}
--- a/internal/pkg/image/registry_hostname.go
+++ b/internal/pkg/image/registry_hostname.go
@@ -13,11 +13,13 @@ import (
 // CRI-O rejects uppercase letters in registry hostnames, so oc-mirror fails early
 // rather than producing mirrored content that cluster nodes cannot pull (OCPBUGS-78497).
 func ValidateDockerDestinationRegistry(dest string) error {
-	if !strings.HasPrefix(dest, consts.DockerProtocol) {
+	// Cut (not HasPrefix) so this matches call sites that detect docker:// with Contains,
+	// including a leading space before the protocol.
+	_, ref, found := strings.Cut(dest, consts.DockerProtocol)
+	if !found {
 		return nil
 	}
 
-	ref := strings.TrimPrefix(dest, consts.DockerProtocol)
 	hostname := registryHostname(ref)
 	if hostname == "" {
 		return fmt.Errorf("destination registry hostname is empty")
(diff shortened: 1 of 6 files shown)


internal/pkg/image/registry_hostname.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/openshift/oc-mirror/v2/internal/pkg/consts"
+)
+
+// ValidateDockerDestinationRegistry validates a docker:// destination argument.
+// CRI-O rejects uppercase letters in registry hostnames, so oc-mirror fails early
+// rather than producing mirrored content that cluster nodes cannot pull (OCPBUGS-78497).
+func ValidateDockerDestinationRegistry(dest string) error {
+	if !strings.HasPrefix(dest, consts.DockerProtocol) {
+		return nil
+	}
```
aguidirh: Nice catch on the bug — failing early here is the right call, and the host/port
parsing handles the realistic cases correctly (`registry.example.com:5000/ns`,
bracketed IPv6, bare `:5000/ns`). CI is green. A few non-blocking observations:

**1. `Contains` at the call sites vs. `HasPrefix` in the helper**

The call sites gate on `strings.Contains(dest[0], consts.DockerProtocol)`, but
`ValidateDockerDestinationRegistry` early-returns `nil` unless the string
`HasPrefix` the protocol. So an argument like `" docker://REGISTRY.EXAMPLE.COM:5000"`
(leading space) skips the new validation entirely, while the surrounding logic —
the `--from`/`--workspace` check just above, and the `return nil` a few lines
below — still treats it as a docker destination via `Contains`. Net result: it
proceeds and fails late at push, which is the case this PR is trying to prevent.

The `HasPrefix` guard and the `TrimPrefix` on the next line are a matched pair:
`TrimPrefix` is a no-op when the protocol isn't at position 0, so it would leave
the leading space in `ref`, and the guard exists to stop that. `strings.Cut`
replaces both with one operation that agrees with the caller:

```go
_, ref, found := strings.Cut(dest, consts.DockerProtocol)
if !found {
    return nil
}

File: internal/pkg/image/registry_hostname.go [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+	if hostForIP := hostname; strings.HasPrefix(hostForIP, "[") && strings.HasSuffix(hostForIP, "]") {
+		hostForIP = hostForIP[1 : len(hostForIP)-1]
+		if net.ParseIP(hostForIP) != nil {
+			return nil
+		}
+	} else if net.ParseIP(hostname) != nil {
+		return nil
+	}
+
+	for _, r := range hostname {
```
aguidirh, line 44: Only the hostname is checked

docker://registry.example.com:5000/MyNamespace still passes validation and
fails at push with "repository name must be lowercase". Possibly intentional
scope-limiting to the reported bug — just flagging in case catching the path too
is cheap here.

Are the namespace and repository name also needs to be lowercase on CRI-O?

internal/pkg/image/registry_hostname.go
```diff
@@ -0,0 +1,68 @@
...
+		if unicode.IsUpper(r) {
+			return fmt.Errorf("destination registry hostname %q contains uppercase characters; use lowercase (e.g. %q) because CRI-O rejects uppercase registry hostnames", hostname, strings.ToLower(hostname))
+		}
+	}
+	return nil
+}
+
+// registryHostname returns the host (without path) from a docker destination
+// reference such as "registry.example.com:5000/ns" or "[::1]:5000/ns".
+func registryHostname(ref string) string {
```
aguidirh, line 54: Malformed bracketed hosts

docker://[]:5000/ns and docker://[not-an-ip]:5000/ns are accepted, and
docker://[MyHost]:5000 produces an error suggesting "[myhost]", which is
still invalid. Minor, but the suggestion being unusable is a little confusing.

aguidirh (commented):
Hi @shishind,

Thanks for your PR, None of these suggestions I made are regressions — they're all cases where an invalid destination slips through to the same late failure that happens today without this patch.

internal/pkg/image/registry_hostname.go
shishind: Addressed in 17cb113c — `ValidateDockerDestinationRegistry` now uses `strings.Cut`, so a leading space before `docker://` is validated the same way as the `Contains` call sites. Added unit coverage for that case on the helper and the CLI validate paths.

internal/pkg/image/registry_hostname.go
shishind: Keeping this scoped to the hostname, which is the OCPBUGS-78497 hole: DNS/registry clients often accept mixed-case hosts, so the mirror can succeed and CRI-O fails later on pull.

Namespace and repository path must also be lowercase in docker/CRI-O references, but those already fail at push with `repository name must be lowercase`. Happy to follow up separately if we want that check too.

internal/pkg/image/registry_hostname.go
shishind: Addressed in 17cb113c — bracketed hosts that are not valid IP literals are now rejected instead of falling through to the hostname case check, so `[MyHost]` no longer suggests `[myhost]`.

aguidirh:
Hi @shishind,

Can you squash the commits in only one and sign it? Without signature it won't allow us to merge the PR.

shishind:
@aguidirh Squashed to a single commit and added the DCO sign-off: `8850a7b4` (`Signed-off-by: Shivam shinde <shishind@redhat.com>`). Also rebased onto current `main`.

aguidirh:
> @aguidirh Squashed to a single commit and added the DCO sign-off: `8850a7b4` (`Signed-off-by: Shivam shinde <shishind@redhat.com>`). Also rebased onto current `main`.

Hi @shishind the commit is not assigned, you need to use the flag `-s` when running the `git commit` command.

shishind:
@aguidirh Amended with `git commit -s`. The commit is now `341c8de8`, authored by me (`shishind@redhat.com`) with:

```
Signed-off-by: Shivam shinde <shishind@redhat.com>
```

aguidirh:
@shishind for some reason github is not identifying your commit as signed as you can see in the picture below:
image

shishind:
It should go ahead now, it was issue with ssh key.

aguidirh:
> It should go ahead now, it was issue with ssh key.

Yeap, it worked now! Thanks :)

aguidirh (approved):

aguidirh:
/verified by unit-tests implemented in the PR

aguidirh:
@shishind I will wait for someone else from the team to add `lgtm`

PR #1505, opened 2026-08-17
OCPBUGS-111612: retry HTTP 429 responses from the destination registry

AlexSmithGH:
## Summary

This is a **partial classification gap**, not a claim that 429s are never retried.

`internal/pkg/mirror.IsErrorRetryable` (formerly the unexported `isErrorRetryable`) is
oc-mirror's own retry classifier, installed into `go.podman.io/common/pkg/retry.Options`.
A rate limit can reach it in three distinct error shapes, and it only recognised one:

| Error shape | Where it comes from | Retried before this PR? |
|---|---|---|
| `errcode.Error{Code: TOOMANYREQUESTS}` (renders as `toomanyrequests: Rate exceeded`) | `registryHTTPResponseToError`, when the registry returns a distribution-spec JSON error body | **Yes.** Falls through to `default:` → `retry.IsErrorRetryable`, which returns true for every errcode except `unauthorized`/`denied`/`nameunknown`/`manifestunknown`. |
| `docker.UnexpectedHTTPStatusError{StatusCode: 429}` | `newUnexpectedHTTPStatusError`, when the 429 body is not a parseable distribution error | **No.** The `errors.As` case intercepts it, 429 misses the 500–504 window, and it returns `false`. Upstream is no help either: it only covers 502–504. |
| `docker.ErrTooManyRequests` — a bare `errors.New("too many requests to registry")` sentinel | `httpResponseToError`, which special-cases `http.StatusTooManyRequests` before it ever builds a typed error | **No.** Matches no case in our switch, and upstream's type switch has no case for it either, so it is retryable by neither classifier. |

The ticket's attached logs contain 40,816 `Failed, retrying in Ns ... (n/5)` lines that are
100% rate limits — that is the first shape working exactly as designed. The bug is that the
other two shapes silently abandon the image on the first 429, on a registry that is telling
us to slow down and try again.

## Changes

**1. Classify all three 429 shapes as retryable** (`internal/pkg/mirror/mirror.go`)

Two new cases in the switch. The `errors.Is(err, docker.ErrTooManyRequests)` case must
precede the `errors.As` case, otherwise the sentinel would be shadowed by whatever the
`errors.As` branch decides.

The classifier is also exported (`isErrorRetryable` → `IsErrorRetryable`) so the `batch`
package can reuse it. The name deliberately mirrors the upstream
`retry.IsErrorRetryable` that the `default` branch delegates to; the package qualifier keeps
the two distinct, so there is no shadowing.

**2. Apply the classifier on all three retry paths** (`internal/pkg/mirror/mirror.go`)

Only `copy` installed the custom classifier. `Check` and `delete` passed `opts.RetryOpts`
straight through to `retry.IfNecessary`, so they silently fell back to upstream's default
heuristic — 502–504 only, no 429 of any shape, and not even `context.DeadlineExceeded`.

A single `retryOptionsFrom(opts)` helper now serves all three sites. It returns a **copy**,
because `opts.RetryOpts` is shared across every image in a run and must not be mutated.

Secondary benefit: this removes a latent panic. `retry.IfNecessary` dereferences
`options.MaxRetry` with no nil guard, so a nil `RetryOpts` reaching `Check` or `delete`
would have crashed inside the upstream call. `retryOptionsFrom` always returns a non-nil
`*retry.Options`.

## Explicitly not changed, because it is already correct

- **`--retry-times` / `--retry-delay` already reach this path.** `mirror.RetryFlags()`
  (`internal/pkg/mirror/options.go:274-280`) binds them directly into an upstream
  `retry.Options` (defaults 5 and 0), wired through
  `internal/pkg/cli/executor.go:175,182,260` into `CopyOptions.RetryOpts`. No new flag is
  needed, and no default is changed.
- **Backoff is already exponential with jitter.** Upstream `retry.IfNecessary` uses
  `2^attempt` seconds (1/2/4/8/16) plus 10% jitter whenever `Delay == 0`, which is the
  default. There is nothing to add here.

## `Retry-After` cannot be honoured

Registries commonly send `Retry-After` with a 429, and honouring it would be strictly better
than blind exponential backoff. It is not implementable from this repo:

- `docker.UnexpectedHTTPStatusError` retains only `StatusCode` and an unexported `status`
  string. The `*http.Response` and its headers are discarded before the error reaches us, so
  the value is simply not available.
- `docker.ErrTooManyRequests` is a bare sentinel and carries nothing at all.
- Even with the value in hand, `retry.IfNecessary` exposes no hook for a per-attempt delay —
  `Options.Delay` is a single constant applied to every attempt.

Honouring `Retry-After` would require an upstream change in `go.podman.io/image` (to preserve
the header on the error) and `go.podman.io/common` (to accept a per-attempt delay). Out of
scope here.

## Out of scope: the c/image format-renegotiation claim

The ticket also suggests a 429 causes c/image to fall back to manifest format conversion,
producing `Error determining manifest MIME type` / conversion failures. **This mechanism is
unproven and is not addressed by this PR.** Specifically, in the attached logs:

- No log line shows a 429 becoming a conversion error.
- The 429 string and the conversion-error string never co-occur on the same image push.
- Every conversion error shares a single end-of-run timestamp, so temporal correlation
  between a 429 and a conversion error was not possible to establish even in principle.

If the behaviour is real, it lives in `go.podman.io/image`'s copy logic, not in oc-mirror, and
belongs in a separate report against that project with a reproducer.

Relatedly: setting `PreserveDigests=false` for related images is **not** a viable workaround.
It permits the manifest to be rewritten, which invalidates any signatures over the original
digest.

## Expected objection: precedent

The likely objection is **OCPBUGS-24348**, "oc-mirror exceeds public AWS ECR rate limit
without retry or backoff", closed *Won't Do*. That ticket asked oc-mirror to work around a
registry's published rate limit — to throttle itself, or to treat a documented quota as an
oc-mirror problem. This report is different in kind: retry-on-429 is already the intended
behaviour here (that is what the custom classifier exists for, and one of the three shapes
already works), and two of the three shapes miss it because of a code-level
misclassification. `docker.ErrTooManyRequests` matching no case in a switch that is
specifically meant to decide whether a rate limit is retryable is a defect regardless of what
any registry's quota happens to be.

## Test coverage

`TestIsErrorRetryable` (`internal/pkg/mirror/mirror_test.go`) is a table-driven test covering:

- `nil` → false
- `docker.UnexpectedHTTPStatusError` with 429 → true; bare and wrapped with `%w`
- 500, 502, 504 → true (existing behaviour preserved)
- 400, 401, 404, 505 → false (no over-broad retrying)
- `context.DeadlineExceeded` → true, `context.Canceled` → false (existing behaviour preserved)
- `docker.ErrTooManyRequests` → true; bare and wrapped with `%w`
- `errcode.Error{Code: ErrorCodeTooManyRequests}` → true, exercising the `default:`
  delegation to `go.podman.io/common`
- `errcode.Error{Code: ErrorCodeManifestUnknown}` → false, confirming the delegation still
  declines the non-retryable errcodes

`TestRetryOptionsFrom` covers the helper: a nil `RetryOpts` produces usable options instead of
a nil dereference, the caller's `retry.Options` is copied rather than mutated, and the
installed classifier does retry a 429.

### Note on the errcode test dependency

The two `errcode` cases required importing `github.com/docker/distribution/registry/api/errcode`
and `.../registry/api/v2`. This is deliberate and load-bearing: it is the exact errcode package
`go.podman.io/common/pkg/retry` itself uses, so it is the only one whose `errcode.Error` the
upstream type switch recognises. Constructing the equivalent value from
`github.com/distribution/distribution/v3/registry/api/errcode` (already a direct dependency)
produces an identical error string but a *different Go type*, which upstream does not match —
verified empirically, that variant classifies as **not** retryable and the test would have
asserted the opposite of real behaviour.

The only dependency effect is that `go mod tidy` promotes
`github.com/docker/distribution v2.8.3+incompatible` from indirect to direct in `go.mod`. No
module is added and the version is unchanged; `go.sum` is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling for image copying, existence checks, and deletion.
  * Added retries for HTTP 429 responses, including unparsed errors.
  * Prevented retry settings from being unexpectedly modified during operations.
  * Preserved caller-provided retry configuration while applying default retry behavior.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 
 require (
 	github.com/docker/cli v29.6.1+incompatible
+	github.com/docker/distribution v2.8.3+incompatible
 	k8s.io/klog v1.0.0
 )
 
@@ -72,7 +73,6 @@ require (
 	github.com/cyphar/filepath-securejoin v0.7.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
-	github.com/docker/distribution v2.8.3+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.9.7 // indirect
 	github.com/docker/go-connections v0.7.0 // indirect
 	github.com/docker/go-events v0.0.0-20250808211157-605354379745 // indirect
--- a/internal/pkg/mirror/mirror.go
+++ b/internal/pkg/mirror/mirror.go
@@ -192,12 +192,6 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions) 
 		co.ReportWriter = opts.Stdout
 	}
 
-	var retryOpts retry.Options
-	if opts.RetryOpts != nil {
-		retryOpts = *opts.RetryOpts
-	}
-	retryOpts.IsErrorRetryable = isErrorRetryable
-
 	//nolint:wrapcheck // context will be added by the calling function
 	return retry.IfNecessary(ctx, func() error {
 		manifestBytes, err := o.mc.CopyImage(ctx, policyContext, destRef, srcRef, co)
@@ -214,11 +208,26 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions) 
 			}
 		}
 		return nil
-	}, &retryOpts)
+	}, retryOptionsFrom(opts))
 }
 
-// Custom implementation to extend `containers/common/pkg/retry.retry`
-func isErrorRetryable(err error) bool {
+// retryOptionsFrom returns a copy of the caller's retry options with oc-mirror's
+// error classification installed. The copy matters: opts.RetryOpts is shared
+// across all images and must not be mutated.
+func retryOptionsFrom(opts *CopyOptions) *retry.Options {
+	var ro retry.Options
+	if opts.RetryOpts != nil {
+		ro = *opts.RetryOpts
+	}
+	ro.IsErrorRetryable = IsErrorRetryable
+	return &ro
+}
+
+// IsErrorRetryable is a custom implementation that extends
+// `go.podman.io/common/pkg/retry.IsErrorRetryable`. The name intentionally mirrors
+// the imported upstream function that the `default` branch delegates to; the
+// package qualifier keeps the two distinct, so there is no conflict.
+func IsErrorRetryable(err error) bool {
 	var httpError docker.UnexpectedHTTPStatusError
 	switch {
 	case err == nil:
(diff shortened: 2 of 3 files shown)


aguidirh:
@AlexSmithGH please squash your commits and sign them. 

I will approve as soon as we have them signed and squashed.

Thanks.

aguidirh (approved):

dorzel (approved):

PR #1526, opened 2026-09-07
OCPBUGS-62723: Skip attestation manifests during multi-arch image copy

aguidirh:
# Description

Multi-arch images from ISVs (e.g. Couchbase, VAST Data) may include attestation manifest entries in their manifest list. These entries carry the platform `unknown/unknown` and the annotation `vnd.docker.reference.type: attestation-manifest` (buildx SLSA provenance).

Proxy registries such as `registry.connect.redhat.com` do not serve those attestation manifests. When oc-mirror copies with `--all` / `--multi-arch all`, every entry in the list is pulled, so the mirror fails with `Image not found`.

This change inspects the source manifest list before copying. When attestation entries are present, only the real platforms (e.g. `linux/amd64`, `linux/arm64`) are copied, and the unpullable attestation manifests are skipped transparently. Every other path is untouched, and if the inspection fails for any reason the original copy behaviour is preserved. The mirrored index keeps its digest.

This is a workaround. A `TODO` marks it for removal once OCI 1.1 is supported by containers-libs ([OCPNODE-2018](https://issues.redhat.com/browse/OCPNODE-2018)), where attestations are referrers rather than manifest list entries.

Github / Jira issue: [OCPBUGS-62723](https://issues.redhat.com/browse/OCPBUGS-62723)

## Type of change

# How Has This Been Tested?

Unit tests were added covering an in-memory registry with three fixtures: a manifest list containing an attestation entry alongside `linux/amd64` and `linux/arm64`, a clean manifest list with no attestation entry, and a single-arch image. A nonexistent reference is also covered.

Run with:

```
make build
make test-unit
```

## Expected Outcome

- The list with an attestation entry copies only the two real platforms and skips the attestation manifest.
- The clean list and the single-arch image keep the existing copy behaviour, unchanged.
- A failed inspection falls back to the existing copy behaviour rather than erroring.

`make build` succeeds and `make test-unit` passes across all packages.

Mirroring an affected ISV image with `--all` should now complete instead of failing with `Image not found`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mirroring for multi-platform images containing attestation manifests.
  * Platformless entries associated with selected image instances are now preserved.
  * Attestation inspection failures no longer prevent the underlying copy operation from continuing.
  * Errors encountered while inspecting manifest-list instances now include additional context.
  * Images without attestations, single-platform images, and standard manifest lists retain their existing mirroring behavior.


internal/pkg/mirror/mirror.go
```diff
@@ -141,6 +142,21 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions)
...
 	}
 
+	// TODO: remove filterAttestationPlatforms when OCI1.1 is supported by containers-libs [OCPNODE-2018](https://redhat.atlassian.net/browse/OCPNODE-2018)
+	// OCPBUGS-62723: When copying all images from a manifest list, check for
+	// attestation entries (unknown/unknown platform) that cannot be pulled from
+	// proxy registries. If found, switch to CopySpecificImages with only the
+	// real platforms so the copy skips unpullable attestation manifests.
+	var instanceDigests []digest.Digest
+	if imageListSelection == copy.CopyAllImages {
+		filter, attestationErr := filterAttestationInstances(ctx, srcRef, sourceCtx)
```
adolfo-ab, line 152: Should we do something with `attestationErr` when it's not nil?

internal/pkg/mirror/mirror_attestation_test.go
```diff
@@ -0,0 +1,293 @@
...
+		Digest:    digest,
+		Size:      size,
+		Platform: &v1.Platform{
+			OS:           os,
+			Architecture: arch,
+		},
+	}, nil
+}
+
+func TestFilterAttestationPlatforms(t *testing.T) {
```
adolfo-ab, line 214: It should be `TestFilterAttestationInstances`, right?
```suggestion
func TestFilterAttestationInstances(t *testing.T) {
```

internal/pkg/mirror/mirror_attestation.go
```diff
@@ -0,0 +1,111 @@
...
+
+// isAttestationManifest returns true if a manifest list entry is an attestation
+// manifest — identified by the "vnd.docker.reference.type" annotation or by
+// having an unknown/unknown platform (common for buildx SLSA provenance).
+func isAttestationManifest(info manifest.ListUpdate) bool {
+	if info.ReadOnly.Annotations["vnd.docker.reference.type"] == "attestation-manifest" {
+		return true
+	}
+	p := info.ReadOnly.Platform
+	return p != nil && p.OS == "unknown" && p.Architecture == "unknown"
```
adolfo-ab, line 110: Does `unknown/unknown` means that we're dealing with an attestation 100% of the time? If there's any doubt maybe we should be conservative and only trust the annotation

internal/pkg/mirror/mirror.go
aguidirh: Implemented at commit [83a766ff](https://github.com/openshift/oc-mirror/pull/1526/commits/83a766ff68b7399a3e05321ac750a55c8723ee58).

internal/pkg/mirror/mirror_attestation_test.go
aguidirh: Implemented at commit [ee58fb9c](https://github.com/openshift/oc-mirror/pull/1526/commits/ee58fb9c).

internal/pkg/mirror/mirror_attestation.go
aguidirh: Implemented at commit [810336cc](https://github.com/openshift/oc-mirror/pull/1526/commits/810336cc).

internal/pkg/mirror/mirror.go
```diff
@@ -141,6 +143,26 @@ func (o *Mirror) copy(ctx context.Context, src, dest string, opts *CopyOptions)
...
+	// unpullable attestation manifests.
+	var instanceDigests []digest.Digest
+	if imageListSelection == copy.CopyAllImages {
+		filter, attestationErr := filterAttestationInstances(ctx, srcRef, sourceCtx)
+		switch {
+		case attestationErr != nil:
+			// Best effort: a failed inspection must not break the copy. If the source
+			// is genuinely unreachable, copy.Image below reports it with better context.
+			clog.New(opts.Global.LogLevel).Debug("attestation check skipped for %s: %v", src, attestationErr)
+		case filter != nil && !filter.empty():
```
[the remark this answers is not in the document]
aguidirh: @coderabbitai

Not applying this one.
 
 `filter.empty()` is only true when the manifest list contains nothing but attestations. Selecting `CopySpecificImages` with an empty selection means "copy no images" — but the index still gets written to the destination. The result is an index pointing at manifests that were never copied, and oc-mirror exits 0, so the user thinks the mirror succeeded.

 Today that case fails with `Image not found`. A visible failure is better than a broken mirror that looks fine.

adolfo-ab (approved):
/lgtm

aguidirh:
@r4f4 could you please have a look as well and let me know what you think about this PR?

